Issue #15, opened 2015-10-07
TimerTriggerAttribute.UseMonitor can cause timer to run at unexpected intervals


brendankowitz:
The following line of code uses a `now` variable that is used to determine the TimeSpan until the next invocation. If the UseMonitor option is enabled then this can be incorrect as it is the time before the function was run.

https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs#L100

mathewc:
Right - not really an issue for functions that don't take long to run, but you're pointing out that if the function is  long running, the skew will be more important. Will fix by recalculating now after the monitor "catch up" invocation.

brendankowitz:
Yeah, I believe this is the behavior I was seeing back in August where I mentioned that jobs didn't exactly start on the the hour when specified. We have various jobs that can take between 1-30 mins to run.

mathewc:
Fixed: https://github.com/Azure/azure-webjobs-sdk-extensions/commit/08ca63f321bbe81e709e5430762d1ebd5b24f120

PR #19, opened 2015-10-20
using same scheduling logic as above

brendankowitz:
One more place that could have the same behavior in (https://github.com/Azure/azure-webjobs-sdk-extensions/issues/15)


mathewc:
Merged, thanks.

PR #20, opened 2015-10-20
tests fail in some cultures that have different date formats (en-AU, etc...)

brendankowitz:
One more place that could have the same behavior in (https://github.com/Azure/azure-webjobs-sdk-extensions/issues/15)

### DIFF
--- a/test/WebJobs.Extensions.Tests/Timers/TimerInfoTests.cs
+++ b/test/WebJobs.Extensions.Tests/Timers/TimerInfoTests.cs
@@ -2,6 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


mathewc:
We should probably apply the same approach that was taken in the SDK core: https://github.com/Azure/azure-webjobs-sdk/commit/379ec7275df938e794ddb9199d3ec587101607db

mathewc:
We should probably apply the same approach that was taken in the SDK core: https://github.com/Azure/azure-webjobs-sdk/commit/379ec7275df938e794ddb9199d3ec587101607db

mathewc:
Resolved via https://github.com/Azure/azure-webjobs-sdk-extensions/commit/74afc31a323963564bd77bd82dbd447dd6d86b81

PR #47, opened 2016-03-10
adding DocumentDB output binding

brettsam:


src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+                    "The DocumentDB Authentication Key must be set either via a '{0}' app setting or directly in code via DocumentDBConfiguration.AuthKey.",
+                    DocumentDBConfiguration.AzureWebJobsDocumentDBKeyName));
+            }
+
+            DocumentDBContext documentDBContext = CreateContext(_docDBConfig, attribute, _jobHostConfig.NameResolver);
+
+            await CreateIfNotExists(attribute, documentDBContext);
+
+            Type coreType = DocumentDBUtility.GetCoreType(parameter.ParameterType);
+
```
brettsam: I shouldn't have to do this here as well (EasyTables finds its core type too). I've filed an issue in the SDK repo to clean this up: https://github.com/Azure/azure-webjobs-sdk/issues/651

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+            IBinding binding = null;
+            try
+            {
+                binding = BindingFactory.BindGenericCollector(parameter, typeof(DocumentDBAsyncCollector<>), coreType,
+                _jobHostConfig.GetService<IConverterManager>(), (s) => documentDBContext);
+            }
+            catch (Exception ex)
+            {
+                // BindingFactory uses a couple levels of reflection so we get TargetInvocationExceptions here.
```
brettsam: Exceptions from BindingFactory should be easier to handle. Filed issue in sdk repo to track: https://github.com/Azure/azure-webjobs-sdk/issues/652

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            ParameterInfo parameter = context.Parameter;
+            DocumentDBAttribute attribute = parameter.GetCustomAttribute<DocumentDBAttribute>(inherit: false);
+            if (attribute == null)
+            {
+                return null;
+            }
+
+            if (_docDBConfig.EndpointUri == null)
```
paulbatum: Should you be checking for string empty here?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            return new DocumentDBContext
+            {
+                Service = new DocumentDBService(config.EndpointUri, config.AuthKey),
+                ResolvedDatabaseName = Resolve(attribute.DatabaseName, resolver),
+                ResolvedCollectionName = Resolve(attribute.CollectionName, resolver)
+            };
+        }
+
+        internal static async Task CreateIfNotExists(DocumentDBAttribute attribute, DocumentDBContext context)
```
paulbatum: This doesnt feel quite right, you're breaking the naming convention by having an async method that does not have an async prefix. Not sure exactly what you should change it to.. CreateIfNotExistsAndSettingIsEnabledAsync is a bit too wordy...

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                throw new InvalidOperationException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    "The DocumentDB Authentication Key must be set either via a '{0}' app setting or directly in code via DocumentDBConfiguration.AuthKey.",
+                    DocumentDBConfiguration.AzureWebJobsDocumentDBKeyName));
+            }
+
+            DocumentDBContext documentDBContext = CreateContext(_docDBConfig, attribute, _jobHostConfig.NameResolver);
+
+            await CreateIfNotExists(attribute, documentDBContext);
```
paulbatum: If I understand the binding lifecycle correctly, by doing this here we would create a collection just by loading up the host with a function configured with CreateIfNotExists = true. i.e. before the function is even invoked. Is that what we want?

src/WebJobs.Extensions.DocumentDB/DocumentDBUtility.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /// <summary>
+        /// Execute the function with retries on throttle.
+        /// </summary>
+        /// <typeparam name="T">The type of return value from the execution.</typeparam>
+        /// <param name="function">The function to execute.</param>
+        /// <returns>The response from the execution.</returns>
+        // Taken from: https://github.com/ryancrawcour/azure-documentdb-dotnet/blob/e7dd2f685554b0a9def63c8925f8e3ef2ad3bff8/samples/code-samples/Shared/Util/DocumentClientHelper.cs
+        public static async Task<T> ExecuteWithRetriesAsync<T>(Func<Task<T>> function)
```
paulbatum: Is there some type of verbose logging that would be appropriate to do here each time an exception is caught and retried?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
brettsam: That was my intention, yes. DocDB team recommended doing this during a setup phase. The other options could result in a slower initial invocation. I wasn't sure if that was a good plan either.

I originally had a check for 404 built into the AsyncCollector that would then create the db and collection if they weren't there but that seemed messy and I didn't like the idea of doing that in every other (future) binding as well.

Other bindings (Queue, Table) do this in the call to BindAsync which I believe happens for every invocation. That seems heavy too. 

I vote we keep it like this and come back to it if there's a better pattern... thoughts?

---
In reply to: [56258300](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56258300) [](ancestors = 56258300)

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
brettsam: Agree on the naming. 

This was kind of a silly way for me to be able to test the logic better. I didn't really like it when I did it.

I did something different in EasyTables where the Configuration takes an optional factory for creating the DocDBService. That allows for fuller testing of stuff like this. I think I'll plug that in here and remove this completely.

---
In reply to: [56257877](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56257877) [](ancestors = 56257877)

src/WebJobs.Extensions.DocumentDB/DocumentDBUtility.cs
brettsam: Good question. I'm not sure I've seen that anywhere else. I'll ask.

---
In reply to: [56258775](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56258775) [](ancestors = 56258775)

src/ExtensionsSample/App.config
```diff
@@ -1,9 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <connectionStrings>
```
mathewc: You're probably removing this because you don't want to forget not to put your stuff there? I have this here just because this sample is intended to be a regular WebJobs SDK project, which usually allows for these settings. I'd just leave these.

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -0,0 +1,37 @@
...
+using ExtensionsSample.Models;
+using Microsoft.Azure.WebJobs;
+
+namespace ExtensionsSample.Samples
+{
+    // To use the DocumentDB samples:
+    // 1. Create a new DocumentDB account
+    // 2. Add the DocumentDB Endpoint Uri to the 'AzureWebJobsDocumentDBUri' App Setting in app.config
+    // 3. Add the DocumentDB Primary Key to the 'AzureWebJobsDocumentDBKey' App Setting in app.config
+    // 4. Add typeof(DocumentDBSamples) to the SamplesTypeLocator in Program.cs
```
mathewc: This new pattern works well I think :)

src/ExtensionsSample/App.config
brettsam: Yeah I was backing my own stuff out. I got a little carried away :-)

---
In reply to: [56353289](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56353289) [](ancestors = 56353289)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -0,0 +1,37 @@
...
+    public static class DocumentDBSamples
+    {
+        // POCO Output binding
+        //   The binding inserts the newly created item into the specified Collection when the
+        //   method successfully exits.
+        //   If the database or the collection do not exist, they are automatically created.
```
mathewc: What is the default for CreateIfNotExists? Many/most of our bindings default to creating the output queue/table/container if it doesn't exist (see http://go.microsoft.com/fwlink/?LinkID=524028&clcid=0x409). I think yours should default to true, then you don't have to specify explicitly below.

brettsam: False, by design. Document Collections are *billable* entities so the DocDB team recommends this is opt-in.

---
In reply to: [56354011](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56354011) [](ancestors = 56354011)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -0,0 +1,37 @@
...
+        //
+        // Other supported types:
+        //   out T[]
+        //   IAsyncCollector<T>
+        //   ICollector<T>
+        public static void InsertDocument(
+            [TimerTrigger("00:01")] TimerInfo timer,
+            [DocumentDB(DatabaseName = "ItemDb", CollectionName = "ItemCollection", CreateIfNotExists = true)] out Item newItem)
```
mathewc: I'm thinking you should have a simplified constructor pattern assuming the 90% case will be that you need to specify the DB name and collection. E.g. `[DocumentDB("ItemDb", "ItemCollection")] out Item newItem`. Makes the code easier to read I think, w/o precluding people from using the longer syntax.

src/ExtensionsSample/Samples/DocumentDBSamples.cs
brettsam: Yes, much nicer. Still needs a little work here -- all of the host.Call() calls fail if those types aren't actually loaded in the TypeLocator.

But even with that it's a lot easier to deal with. I figure we can incrementally improve the experience as we go forward.

---
In reply to: [56353574](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56353574) [](ancestors = 56353574)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
brettsam: Good point.

---
In reply to: [56354632](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56354632) [](ancestors = 56354632)

mathewc: Also, do you have any DbName defaulting in place, like you have for EasyTables? I wonder if the Document DB SDK has a declarative mapping attribute or anything people could apply to their poco types that we could read/use. That's just a future thought.

src/ExtensionsSample/packages.config
```diff
@@ -6,6 +6,7 @@
   <package id="Microsoft.AspNet.WebHooks.Receivers" version="1.2.0-rc1" targetFramework="net45" />
   <package id="Microsoft.AspNet.WebHooks.Receivers.GitHub" version="1.2.0-rc1" targetFramework="net45" />
   <package id="Microsoft.Azure.Mobile.Client" version="2.0.1" targetFramework="net45" />
+  <package id="Microsoft.Azure.DocumentDB" version="1.5.3" targetFramework="net451" />
```
mathewc: Why is the target framework different on this one?

src/WebJobs.Extensions.DocumentDB.Nuget/WebJobs.Extensions.DocumentDB.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <copyright>&#169; Microsoft Corporation. All rights reserved.</copyright>
+    <summary>Microsoft.Azure.WebJobs.Extensions.DocumentDB adds DocumentDB binders to the WebJobs SDK.</summary>
+    <description>This package contains the runtime assemblies for Microsoft.Azure.WebJobs.Extensions.DocumentDB. For more information, please visit https://azure.microsoft.com/en-us/documentation/articles/websites-webjobs-resources</description>
+    <language>en-US</language>
+    <projectUrl>https://azure.microsoft.com/en-us/documentation/articles/websites-webjobs-resources</projectUrl>
+    <licenseUrl>$PackageEULA$</licenseUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <tags>Microsoft Azure WebJobs Jobs Extensions DocumentDB document windowsazureofficial Web</tags>
+    <dependencies>
+      <dependency id="Microsoft.Azure.WebJobs.Extensions" version="[$WebJobsPackageVersion$]" />
```
mathewc: I guess you'll do the same thing here w.r.t. package versions that you did in EasyTables?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
mathewc: Cool. Might link to that bug in a source comment

src/WebJobs.Extensions.DocumentDB.Nuget/WebJobs.Extensions.DocumentDB.nuspec
brettsam: Yes. I missed the brackets here. Good catch. Everything else is done in this PR already.

---
In reply to: [56355693](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56355693) [](ancestors = 56355693)

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBAttributeBindingProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+            // If we queried for the Database or Collection before creation, we may hit a race condition
+            // if multiple instances are running the same code. So let's just create and ignore a Conflict.
+            await DocumentDBUtility.ExecuteAndIgnoreStatusCodeAsync(HttpStatusCode.Conflict,
+                () => service.CreateDatabaseAsync(database));
+
+            await DocumentDBUtility.ExecuteAndIgnoreStatusCodeAsync(HttpStatusCode.Conflict,
+                () => service.CreateDocumentCollectionAsync(databaseUri, documentCollection));
+        }
+
+        private static string Resolve(string value, INameResolver resolver)
```
mathewc: I see every single extension has to duplicate this. Probably something I should expose as a public helper/extension method. Not for now, just bringing it up. Maybe log a bug for it.

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Attribute used to binds a parameter to an DocumentDB type.
```
mathewc: nit: bind

Perhaps "Attribute used to bind to an Azure DocumentDB Table"?

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The method parameter type can be one of the following:
+    /// <list type="bullet">
+    /// <item><description><see cref="ICollector{T}"/></description></item>
+    /// <item><description><see cref="IAsyncCollector{T}"/></description></item>
```
mathewc: The binding helpers you're using don't support binding to string and other simple types? We'll eventually want to add that support - all the other bindings do. Perhaps log a bug.

src/WebJobs.Extensions.DocumentDB/DocumentDBUtility.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <list type="bullet">
+        /// <item><description><see cref="ICollector{T}"/></description></item>
+        /// <item><description>T[]</description></item>
+        /// <item><description>out T</description></item>
+        /// <item><description>out T[]</description></item>
+        /// </list>
+        /// </remarks>
+        /// <param name="type">The Type to evaluate.</param>
+        /// <returns>The core Type</returns>
+        public static Type GetCoreType(Type type)
```
mathewc: Correct me if I'm wrong, but I think I saw the same set of generics helper methods also defined in EasyTables? If so, rather than duplicate, I'd be fine with creating a Common folder in the core Extensions project, and source linking some of these types of things, until we decide what to promote to public helpers.

src/WebJobs.Extensions.DocumentDB/DocumentDBUtility.cs
mathewc: Currently the various bindings handle their internal retries like this silently, only emitting logs and failing an invocation if retries didn't succeed. So I don't think there is any logging that needs to be done here.

src/ExtensionsSample/packages.config
brettsam: I was testing a private nuget package from docdb that required net451. This must have stuck around when I backed it out. I've fixed it.

---
In reply to: [56355454](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56355454) [](ancestors = 56355454)

src/WebJobs.Extensions.DocumentDB/DocumentDBUtility.cs
brettsam: Yes it's identical. I'll fix that up as you suggest.

---
In reply to: [56358369](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56358369) [](ancestors = 56358369)

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
brettsam: They could -- I have to supply the proper Converter to convert from the type to string, etc. It's not high-pri for this one as it accepts an object directly so that's the easiest thing to use. But I will file an issue to come back to it.

---
In reply to: [56357361](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56357361) [](ancestors = 56357361)

mathewc: Wondering how you support non C# languages in Functions w/o a string binding? E.g. all the other languages have to bind to string/JSON

brettsam: You bring up a good point. This worked for EasyTables b/c JObject was inherently supported. Let me start writing the script code and make sure it works E2E in node before I move forward.

---
In reply to: [56370824](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56370824) [](ancestors = 56370824)

brettsam: I just tested it. DocumentClient inherently supports JObject as well. So nothing fancy has to happen for node scripts to work.

---
In reply to: [56380664](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56380664) [](ancestors = 56380664,56370824)

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
brettsam: Done.

---
In reply to: [56357111](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56357111) [](ancestors = 56357111)

src/WebJobs.Extensions.DocumentDB/DocumentDBUtility.cs
brettsam: I've done this but it required an extern alias in the test project. I'll file an issue to undo that once this common code can be removed.

It's a byproduct of us having a single test project for simplicity. Since two projects that the tests reference share the same file, there's a conflict.

---
In reply to: [56360306](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56360306) [](ancestors = 56360306,56358369)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
brettsam: At the moment, DocumentClient doesn't do anything like that. You need to explicitly pass it the URI to the collection. If they support it, we can definitely do that.

---
In reply to: [56355268](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/47#discussion_r56355268) [](ancestors = 56355268)

paulbatum:
:shipit:

PR #53, opened 2016-03-18
EasyTables: TableName setting wins;adding anonymous type support

brettsam:
Previously, even if you did something like `[EasyTable(TableName="SomeTable")] TodoItem item`, we'd still send to the `TodoItem` table. Now we send to the `SomeTable` table.

Also allowing you to use a parameter of type `object` to use anonymous types. We convert them directly to `JObject` if you do this. See the new method in `EasyTableSamples.cs` for an example.


src/WebJobs.Extensions.EasyTables/Extensions/MobileServiceClientExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+            PropertyInfo settingsProperty = serializer.GetType().GetProperty("SerializerSettings", BindingFlags.Public | BindingFlags.Instance);
+            MobileServiceJsonSerializerSettings settings = settingsProperty.GetValue(serializer) as MobileServiceJsonSerializerSettings;
+
+            // Get cache
+            FieldInfo tableNameCacheField = settings.ContractResolver.GetType().GetField("tableNameCache", BindingFlags.NonPublic | BindingFlags.Instance);
+            Dictionary<Type, string> tableNameCache = tableNameCacheField.GetValue(settings.ContractResolver) as Dictionary<Type, string>;
+
+            // Update cache
+            tableNameCache[type] = tableName;
+        }
```
paulbatum: This is just awful, who's idea was it to do private reflection?? :trollface: 

Seriously though, lets be a bit more defensive here. If any of this reflection fails, it will do so with a hugely nasty error (an impossible to understand null reference or some such). Can you add some basic null checks and just in general try to make sure the method throws a useful exception if something goes wrong?

paulbatum:
Looks good! Lets make the reflection code a bit more defensive and then :shipit:

src/WebJobs.Extensions.EasyTables/Extensions/IMobileServiceClientExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    internal static class IMobileServiceClientExtensions
+    {
+        private const string SerializerPropertyName = "Serializer";
+        private const string SerializerSettingsPropertyName = "SerializerSettings";
+        private const string TableNameCacheFieldName = "tableNameCache";
+
+        // Specifying TableName on the attribute overrides any table name inferred by the object's type. So if a 
+        // TableName has been set, we have to do some private reflection to update the internal tableNameCache.
+        // We will address this in the Mobile Services code to make this scenario supported and then remove this
```
mathewc: Do you guys have a bug tracking that?

brettsam: We do now! https://github.com/Azure/azure-mobile-apps-net-client/issues/105

PR #57, opened 2016-03-21
adding attribute-level overrides to EasyTables, DocumentDB, and NotificationHubs

brettsam:


src/WebJobs.Extensions.EasyTables/EasyTableAttribute.cs
```diff
@@ -37,5 +38,21 @@ public sealed class EasyTableAttribute : Attribute
...
+
+        /// <summary>
+        /// Optional. A string value indicating the app setting to use as the Azure Mobile App Api Key, if different
+        /// than the one specified in the <see cref="EasyTablesConfiguration"/>.
+        /// This value affects this parameter as follows:
+        /// - If it is null, the ApiKey value from <see cref="EasyTablesConfiguration"/> is used.
+        /// - If it is equal to string.Empty, no Api Key is used.
+        /// - If it is any other string, the value of that string is used as the Api Key.
+        /// </summary>
+        public string ApiKey { get; set; }
```
brettsam: Thoughts on this? I needed a way to let the parameter itself say "I don't want any api key" -- but also need a default way (null) to say "just use the global one"

src/Common/ConfigurationUtility.cs
```diff
@@ -0,0 +1,24 @@
...
+            if (!string.IsNullOrEmpty(configValue))
+            {
+                // config values take precedence over environment values
+                return configValue;
+            }
+
+            return Environment.GetEnvironmentVariable(settingName) ?? configValue;
+        }
+    }
+}
```
brettsam: I'm going to undo this -- all these extensions already have some internal Resolve() method. I'll use that for now and we can expose one from core in the future.

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -56,6 +55,11 @@ internal static string Resolve(string key)
...
                 if (string.IsNullOrEmpty(value))
                 {
+                    value = ConfigurationManager.AppSettings[key];
+                }
+
+                if (string.IsNullOrEmpty(value))
+                {
                     value = Environment.GetEnvironmentVariable(key);
                 }
             }
```
MikeStall: We repeat this code a lot. Can we call NameResolver?

brettsam: @mathewc and I talked about this in another review and punted it for now. The issue was that the default name resolver throws so it wasn't useful and we had to be defensive against that. We decided to open this to track: https://github.com/Azure/azure-webjobs-sdk/issues/653

MikeStall: What's supposed to happen if the key doesn't exist here?  Is that still a success case?

---
In reply to: [57023038](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/57#discussion_r57023038) [](ancestors = 57023038)

brettsam: It returns null. This isn't really a 'Resolve' method (which is why I renamed it). This is treating the string strictly as a connectionstring/appsetting/environmentvar. If it's not there, it returns null.

src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
```diff
@@ -27,5 +28,17 @@ public sealed class NotificationHubAttribute : Attribute
         /// A tag expression is any boolean expression constructed using the logical operator
```
pragnagopa: Can you add Optional in comments for TagExpression as well. If not set, it sends broadcast

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -55,8 +43,34 @@ public void Initialize(ExtensionConfigContext context)
             }
             var converterManager = context.Config.GetService<IConverterManager>();
             converterManager = converterManager.AddNotificationHubConverters();
-            var provider = new NotificationHubAttributeBindingProvider(context.Config.NameResolver, converterManager, this);
+            var provider = new NotificationHubAttributeBindingProvider(converterManager, this);
             context.Config.RegisterBindingExtension(provider);
         }
+
+        internal static string GetSettingFromConfigOrEnvironment(string key)
```
pragnagopa: GetSettingFromConfigOrEnvironment looks like a common helper. Can we this shared across extensions?

brettsam: Ideally yeah. But we want to actually do it the 'right' way at the core level, so we're keeping all of this internal for beta. Tracking issue: https://github.com/Azure/azure-webjobs-sdk/issues/653

pragnagopa:
:+1:

PR #63, opened 2016-03-30
docdb partition support

brettsam:
Adds support for `PartitionKey` and `CollectionThroughput` on `DocumentDbAttribute`. Also changes `CreateIfNotExists` to only apply to output bindings. For background, see: https://azure.microsoft.com/en-us/documentation/articles/documentdb-partition-data/

The behavior of the two new properties is as such:

- `PartitionKey` on an **output** paramter: If  `CreateIfNotExists` is `true`, the collection will be created with this partition key path. If the value is invalid, DocumentClient will throw an exception.
- `PartitionKey` on an **input** parameter: The value will be used in conjunction with `Id` to lookup the record. If there is no partition for the collection, this value is not required.
- `CollectionThroughput` on an **output** parameter: If `CreateIfNotExists` is `true` and the value is non-zero, the value will be used to specify throughput for the collection. If the value is invalid, DocumentClient will throw an exception.

Note: Also updated ApiHub to target the latest Core SDK. I was getting build warnings that there were version mismatches.

### DIFF
--- a/src/ExtensionsSample/Models/ItemDoc.cs
+++ b/src/ExtensionsSample/Models/ItemDoc.cs
@@ -10,6 +10,9 @@ namespace ExtensionsSample.Models
         [JsonProperty("id")]
         public string Id { get; set; }
 
+        [JsonProperty("partition")]
+        public string Partition { get; set; }
+
         [JsonProperty("text")]
         public string Text { get; set; }
     }
--- a/src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
@@ -5,6 +5,7 @@ using System;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.Documents;
 using Microsoft.Azure.Documents.Client;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Newtonsoft.Json;
@@ -17,11 +18,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
         private ParameterInfo _parameter;
         private DocumentDBContext _context;
         private JObject _originalItem;
-        private string _id;
+        private DocumentDBItemBindingContext _itemBindingContext;
 
-        public DocumentDBItemValueBinder(ParameterInfo parameter, DocumentDBContext context, string id)
+        public DocumentDBItemValueBinder(ParameterInfo parameter, DocumentDBContext context, DocumentDBItemBindingContext itemBindingContext)
         {
-            _id = id;
+            _itemBindingContext = itemBindingContext;
             _parameter = parameter;
             _context = context;
         }
@@ -43,9 +44,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 
         public object GetValue()
         {
-            Uri documentUri = UriFactory.CreateDocumentUri(_context.ResolvedDatabaseName, _context.ResolvedCollectionName, _id);
+            Uri documentUri = UriFactory.CreateDocumentUri(_context.ResolvedDatabaseName, _context.ResolvedCollectionName, _itemBindingContext.Id);
+            RequestOptions options = null;
 
-            T document = DocumentDBUtility.ExecuteWithRetriesAsync(() => _context.Service.ReadDocumentAsync<T>(documentUri),
+            if (!string.IsNullOrEmpty(_itemBindingContext.PartitionKey))
+            {
+                options = new RequestOptions
+                {
+                    PartitionKey = new PartitionKey(_itemBindingContext.PartitionKey)
+                };
+            }
+
+            T document = DocumentDBUtility.ExecuteWithRetriesAsync(() => _context.Service.ReadDocumentAsync<T>(documentUri, options),
                 _context.MaxThrottleRetries, ignoreNotFound: true).Result;
 
             if (document != null)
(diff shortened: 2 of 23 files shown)


src/ExtensionsSample/Models/ItemDoc.cs
```diff
@@ -10,6 +10,9 @@ public class ItemDoc
         [JsonProperty("id")]
         public string Id { get; set; }
 
+        [JsonProperty("partition")]
+        public string Partition { get; set; }
```
pragnagopa: I don't see any samples using ItemDoc. DocumentDBSamples are using Item (same as easytables). Update docdb samples to use itemdoc

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
```diff
@@ -17,11 +18,11 @@ internal class DocumentDBItemValueBinder<T> : IValueBinder where T : class
...
         private DocumentDBContext _context;
         private JObject _originalItem;
-        private string _id;
+        private DocumentDBItemBindingContext _bindingContext;
 
-        public DocumentDBItemValueBinder(ParameterInfo parameter, DocumentDBContext context, string id)
+        public DocumentDBItemValueBinder(ParameterInfo parameter, DocumentDBContext context, DocumentDBItemBindingContext bindingContext)
         {
-            _id = id;
+            _bindingContext = bindingContext;
```
pragnagopa: Nit: rename _bindingContext  to _itemBindingContext

src/WebJobs.Extensions.DocumentDB/WebJobs.Extensions.DocumentDB.csproj
```diff
@@ -137,7 +138,9 @@
       <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
     </PropertyGroup>
     <Error Condition="!Exists('..\..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets'))" />
+    <Error Condition="!Exists('..\..\packages\Microsoft.Azure.DocumentDB.1.6.3\build\Microsoft.Azure.DocumentDB.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\packages\Microsoft.Azure.DocumentDB.1.6.3\build\Microsoft.Azure.DocumentDB.targets'))" />
```
pragnagopa: According to https://docs.nuget.org/consume/package-restore/migrating-to-automatic-package-restore Target EnsureNuGetPackageBuildImports is not needed any more for automatic nuget restore. Delete if not needed.

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
```diff
@@ -64,8 +66,24 @@ public DocumentDBAttribute(string databaseName, string collectionName)
...
+        /// May include binding parameters.
+        /// </summary>
+        public string PartitionKey { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// When specified on an output binding and <see cref="CreateIfNotExists"/> is true, defines the throughput of the created
+        /// collection.
+        /// </summary>
+        public int CollectionThroughput { get; set; }
```
pragnagopa: Rename to OfferThroughput to match property exposed by documentdb?

src/ExtensionsSample/Models/ItemDoc.cs
brettsam: I must have backed that out. Fixed.

---
In reply to: [59794188](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/63#discussion_r59794188) [](ancestors = 59794188)

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
brettsam: Fixed.

---
In reply to: [59795439](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/63#discussion_r59795439) [](ancestors = 59795439)

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
brettsam: I've pinged the DocDB PM to see if he has input. 'Offer' sounded too generic to me and I wasn't sure if that property was used elsewhere. I'll do whatever he suggests :-)

---
In reply to: [59797287](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/63#discussion_r59797287) [](ancestors = 59797287)

brettsam: Aravind liked keeping it 'CollectionThroughput' since it simplifies the explanation for Functions.

---
In reply to: [59906136](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/63#discussion_r59906136) [](ancestors = 59906136,59797287)

pragnagopa:
:+1:  Do get the latest changes and squash commits

src/WebJobs.Extensions.DocumentDB/WebJobs.Extensions.DocumentDB.csproj
brettsam: If we want to fix this, let's tackle this across all projects. I'll leave it as-is for now.

---
In reply to: [59796473](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/63#discussion_r59796473) [](ancestors = 59796473)

PR #71, opened 2016-04-07
fixing docdb retry bug

brettsam:
I accidentally removed the Task.Delay in my last checkin. That effectively broke docdb under any kind of stress where clients were throttled.

### DIFF
--- a/src/WebJobs.Extensions.DocumentDB/DocumentDBContext.cs
+++ b/src/WebJobs.Extensions.DocumentDB/DocumentDBContext.cs
@@ -1,13 +1,24 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Host;
+
 namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 {
     internal class DocumentDBContext
     {
+        private const int DefaultMaxThrottleRetries = 10;
+
+        public DocumentDBContext()
+        {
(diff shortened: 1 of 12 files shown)


fabiocav:
Is this bug in the package that is currently consumed by the script host we're planning to deploy? Do we need to create a new package and update the host to use it?

brettsam:
Yes -- this bug has been there since just before //build. Once signed off, this needs to be merged and a new package pulled into script.

fabiocav:
This looks good.

src/WebJobs.Extensions.DocumentDB/DocumentDBContext.cs
```diff
@@ -1,13 +1,24 @@
...
+        public DocumentDBContext()
+        {
+            MaxThrottleRetries = DefaultMaxThrottleRetries;
+        }
+
         public IDocumentDBService Service { get; set; }
         public string ResolvedDatabaseName { get; set; }
         public string ResolvedCollectionName { get; set; }
         public string ResolvedId { get; set; }
+        public int MaxThrottleRetries { get; set; }
```
brettsam: I'm putting this on the DocDBContext so that in the future I can more easily support a per-binding override.

src/WebJobs.Extensions.DocumentDB/DocumentDBContext.cs
```diff
@@ -1,13 +1,24 @@
...
+        public TraceWriter Trace { get; set; }
```
brettsam: Even though I'm doing no tracing anywhere right now, I decided to leave this property here. It should make debugging easier in the future.

mathewc:
:ship:

Issue #73, opened 2016-04-12
[NotificationHub]Add support to send notifications per platform


pragnagopa:

pragnagopa:
https://github.com/Azure/azure-webjobs-sdk-extensions/pull/74 added support.

PR #74, opened 2016-04-12
[NotificationHub]Add support to send native notifications

pragnagopa:


pragnagopa:
Fixes #73

src/WebJobs.Extensions.NotificationHubs/Config/Converter.cs
```diff
@@ -9,19 +10,63 @@ namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
...
+                JObject jobj = JObject.Parse(notificationAsString);
+                Dictionary<string, string> templateProperties = jobj.ToObject<Dictionary<string, string>>();
+                notification = new TemplateNotification(templateProperties);
+            }
+            else
+            {
+                NotificationPlatform platform;
+                try
+                {
+                    platform = (NotificationPlatform)Enum.Parse(typeof(NotificationPlatform), NotificationPlatform);
```
brettsam: Can you use the enum as the attribute property?

pragnagopa: Thought of exposing NotificationPlatform type. But that would require users to add reference to NotificationHubs nuget.

src/WebJobs.Extensions.NotificationHubs/Config/Converter.cs
```diff
@@ -9,19 +10,63 @@ namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
 {
     internal static class Converter
     {
+        public static string NotificationPlatform { get; set; }
+
```
brettsam: Can you just pass this into AddNotificationHubConverters as a parameter? That'd be nicer.

src/WebJobs.Extensions.NotificationHubs/Config/Converter.cs
brettsam: This would be set via bindings.json -- so this shouldn't affect user scripts. The only issue is whether the bindings.json is deserialized properly. If not, we can make it work. It'd be nicer to just use the enum here and remove all of your parsing code.

src/ExtensionsSample/Samples/NotificationHubSamples.cs
```diff
@@ -36,7 +47,17 @@ public static class NotificationHubSamples
             messageProperties = "{\"message\":\"Hello\",\"location\":\"Redmond\"}";
         }
 
-        // The binding sends multiple push notification to any clients registered with the template
+        // NotificationHub binding out Notification type
+        // This binding broadcasts native apple push notification to all the registered iOS clients
+        // method successfully exits.
```
mathewc: I think here and throughout you're missing a "when" before "successfully exits". Recommend fixing up the wording on all these.

src/ExtensionsSample/Samples/NotificationHubSamples.cs
```diff
@@ -59,8 +80,8 @@ public static class NotificationHubSamples
             await notifications.AddAsync(GetTemplateNotification("World"));
         }
 
-        //   The binding creates a strongly-typed AsyncCollector, which is used to send push notifications. 
-        //   The binding does not do anything with the results when the function exits.  
+        // This binding creates a strongly-typed AsyncCollector, which is used to send push notifications.
```
mathewc: This method doesn't use AsyncCollector, right? It uses IDictionary. If you're talking about internal implementation details, I don't think that belongs in the sample description.

pragnagopa: Typo in the comments. Fixed.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAttributeBindingProvider.cs
```diff
@@ -59,9 +59,12 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
 
             Func<string, INotificationHubClientService> invokeStringBinder = (invokeString) => service;
 
+            var converterManager = _jobHostConfig.GetService<IConverterManager>();
+            converterManager = converterManager.AddNotificationHubConverters(attribute.Platform);
```
MikeStall: converterManager is global to the host. So if you have 5 function definitions with 5 different attributes, each with a different Platform setting, they'll all overwrite each other.
We need to find some more generalized way to do this with the converter.

MikeStall:
:-1:

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAttributeBindingProvider.cs
pragnagopa: Will hold off on this change until we have support converter manager at attribute level

pragnagopa: Opened issue# https://github.com/Azure/azure-webjobs-sdk/issues/697

mathewc:
What is the status of this review - should it be closed?

pragnagopa:
Closing this PR. This will be addressed as part of refactoring code to use  latest binding helpers

PR #76, opened 2016-04-12
renaming EasyTables to MobileTables

brettsam:
Now with less Easy!

image


src/ExtensionsSample/Program.cs
```diff
@@ -35,7 +35,7 @@ public static void Main(string[] args)
             config.UseFiles(filesConfig);
             config.UseTimers();
             config.UseSample();
-            config.UseEasyTables();
+            config.UseMobileTables();
```
mathewc: Do you guys think there will be any other Mobile artifacts that you want to bind to other than Tables? I ask because if so, it would make sense to have an extension named WebJobs.Extensions.Mobile, and config.UseMobile(). Then your extension is a package for one or more mobile bindings.

mathewc:
Renames look good. Just see my comment on package name.

src/ExtensionsSample/Program.cs
brettsam: Renaming seems like a decent idea to me. Maybe .MobileApps? @paulbatum  -- any preference? This would affect the project name, project, and namespaces.

fabiocav: I like that idea. .MobileApps sounds like the right name to use.

fabiocav: Or AppServiceMobileAppsApps :trollface:

mathewc: AzureAppServiceEasyMobileAppsApps

PR #77, opened 2016-04-13
Adding tests to work with local file system and updating to latest SDK version.

safihamid:
Updating extensions to use the latest version of ApiHub SDK

Conflicts:
	src/ExtensionsSample/Program.cs
	test/WebJobs.Extensions.Tests/Extensions/ApiHub/ApiHubBindingEndToEndTests.cs

Updates to support PollInterval


azurecla:
Hi __@safihamid__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (hamids). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

src/ExtensionsSample/Program.cs
```diff
@@ -66,7 +69,8 @@ public static void Main(string[] args)
                 typeof(SendGridSamples),
                 typeof(TableSamples),
                 typeof(TimerSamples),
-                typeof(WebHookSamples));
+                typeof(WebHookSamples),
+                typeof(ApiHubSamples));
```
mathewc: I'd rather you left this the way it was - when I merged your stuff last week, I moved the ApiHub samples to the same model that the EasyTables and other samples use. See both the sample files - they have doc on how to enable them. The idea is that if/when people want to run them, they can just add this line. This makes the samples much easier to run for people - they don't have to configure the world to get them running.

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
```diff
@@ -9,13 +9,15 @@ namespace Microsoft.Azure.WebJobs
     public sealed class ApiHubFileTriggerAttribute : ApiHubFileAttribute
     {
         /// <summary>
-        ///  Create an instance of an attribute that points to the file on SAAS file trigger
+        /// Create an instance of an attribute that points to the file on SAAS file trigger
```
mathewc: This attribute description is hard to understand. How about "Attribute used to bind a parameter to a SAAS file"?

src/ExtensionsSample/Program.cs
safihamid: There is no need to configure anything anymore to have ApiHub stuff work, the samples are now much easier and work with local file system. they can be run without ANY extra configuration than other regular stuff.

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
```diff
@@ -9,13 +9,15 @@ namespace Microsoft.Azure.WebJobs
...
         /// </summary>
         /// <param name="key">App settings key name that have the connections string</param>
-        /// <param name="path">Relative path to the file <example>/folder/subfolder/file.txt</example> </param>
-        public ApiHubFileTriggerAttribute(string key, string path)
+        /// <param name="path">Relative path to the file <example>/folder/subfolder/file.txt</example></param>
+        /// <param name="pollIntervalInSeconds">The poll interval in seconds.</param>
+        public ApiHubFileTriggerAttribute(string key, string path, int pollIntervalInSeconds = 0)
```
mathewc: What does the default value of zero mean? Shouldn't this be a higher value (e.g. 15 seconds?)

src/ExtensionsSample/Program.cs
safihamid: Take a look at CheckForApiHub function below:

 else
 +            {
 +                apiHubConnectionString = "UseLocalFileSystem=true;Path=" + Path.GetTempPath() + "ApiHubDropBox";
 +            }

src/ExtensionsSample/Program.cs
```diff
@@ -112,5 +116,39 @@ private static void EnsureSampleDirectoriesExist(string rootFilesPath)
 
             File.WriteAllText(Path.Combine(rootFilesPath, "input.txt"), "WebJobs SDK Extensions!");
         }
+
+        private static bool CheckForApiHub(JobHostConfiguration config)
```
mathewc: Recommend removing this - when people enable the sample, they will have configured a valid connection string, so this shouldn't be needed, right?

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
safihamid: 0 means use the default value which is 90 seconds now. look at this code:
        private const int DefaultPollIntervalInSeconds = 90;

            _poll = _folderSource.CreateNewFileWatcher(OnNewFile, _pollIntervalInSeconds > 0 ? _pollIntervalInSeconds : DefaultPollIntervalInSeconds);

src/ExtensionsSample/Program.cs
safihamid: No this means if there is no valid connection string just use the local file system. and then everything just works fine.

src/ExtensionsSample/Program.cs
mathewc: I'd still prefer if you left these opt in as they are now - that's not a problem, right? That's the pattern we're using for all the other extensions we've been onboarding lately: EasyTables, DocumentDB, NotificaitonHubs, etc.

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
mathewc: I see, thanks.

mathewc:
:ship:

PR #82, opened 2016-04-20
Fixing the bug for non-C# functions not being triggered as well as ad…

safihamid:
Fixing the bug for non-C# functions not being triggered as well as adding support for FileWatcherType
And making the ApiHubListener as singleton.

### DIFF
--- a/src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
+++ b/src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Microsoft.Azure.ApiHub;
 
 namespace Microsoft.Azure.WebJobs
 {
@@ -13,11 +14,13 @@ namespace Microsoft.Azure.WebJobs
         /// </summary>
         /// <param name="key">App settings key name that have the connections string</param>
         /// <param name="path">Relative path to the file <example>/folder/subfolder/file.txt</example></param>
+        /// <param name="fileWatcherType">Type of the file watcher.</param>
         /// <param name="pollIntervalInSeconds">The poll interval in seconds.</param>
-        public ApiHubFileTriggerAttribute(string key, string path, int pollIntervalInSeconds = 0)
+        public ApiHubFileTriggerAttribute(string key, string path, FileWatcherType fileWatcherType = FileWatcherType.Created, int pollIntervalInSeconds = 0)
             : base(key, path)
         {
             this.PollIntervalInSeconds = pollIntervalInSeconds;
+            this.FileWatcherType = fileWatcherType;
         }
 
         /// <summary>
@@ -27,5 +30,13 @@ namespace Microsoft.Azure.WebJobs
         /// The poll interval in seconds.
         /// </value>
         public int PollIntervalInSeconds { get; set; }
+
+        /// <summary>
+        /// Gets or sets the type of the file watcher. Default is 'Created'.
+        /// </summary>
+        /// <value>
+        /// The type of the file watcher.
+        /// </value>
+        public FileWatcherType FileWatcherType { get; set; }
     }
 }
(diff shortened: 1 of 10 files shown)


azurecla:
Hi __@safihamid__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (hamids). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
```diff
@@ -13,11 +14,13 @@ public sealed class ApiHubFileTriggerAttribute : ApiHubFileAttribute
         /// </summary>
         /// <param name="key">App settings key name that have the connections string</param>
         /// <param name="path">Relative path to the file <example>/folder/subfolder/file.txt</example></param>
+        /// <param name="fileWatcherType">Type of the file watcher.</param>
         /// <param name="pollIntervalInSeconds">The poll interval in seconds.</param>
-        public ApiHubFileTriggerAttribute(string key, string path, int pollIntervalInSeconds = 0)
+        public ApiHubFileTriggerAttribute(string key, string path, FileWatcherType fileWatcherType = FileWatcherType.Created, int pollIntervalInSeconds = 0)
```
mathewc: Where is FileWatcherType defined? ApiHub defined its own enum? In FileTriggerAttribute (source [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions/Extensions/Files/FileTriggerAttribute.cs#L37)) I used the core WatcherChangeTypes flags enum.

mathewc:
It wasn't clear from the source changes or your comments what the issue was preventing non-C# scenarios from working?

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
safihamid: Yes this is defined in ApiHub SDK as SDK needs this info.It only supports Created or Updated as that's the limitation of the underlying SAAS connectors.

safihamid:
The issue was in SetValueAsync call. I needed to call OnComplete even if the input value was null.

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
mathewc: Is this enun a flags enum, allowing you to have a single function that monitors for both Creation / Update? That will be a very common scenario and I think we should make that easy for people to accomplish. E.g. Both FileTrigger and BlobTrigger allow this.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -46,11 +51,13 @@ public void Dispose()
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
-            _poll = _folderSource.CreateNewFileWatcher(OnNewFile, _pollIntervalInSeconds > 0 ? _pollIntervalInSeconds : DefaultPollIntervalInSeconds);
+            _poll = _folderSource.CreateFileWatcher(_fileWatcherType, OnFileWatcher, nextItem: null, pollIntervalInSeconds: _pollIntervalInSeconds > 0 ? _pollIntervalInSeconds : DefaultPollIntervalInSeconds);
+
+            // TODO: need to decide what to do when _poll is null i.e. trigger folder does not exist.
```
mathewc: See how I handle this in FileTrigger https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions/Extensions/Files/Listener/FileListener.cs#L67 - I throw an error in this case. I think that's what you should do as well, rather than silently ignoring the issue.

src/WebJobs.Extensions.ApiHub/ApiHubFileTriggerAttribute.cs
galiniliev: Currently those are two separate triggers. Let's take this as a separate work item

PR #97, opened 2016-05-17
Support for newest BindingFactory

brettsam:


src/ExtensionsSample/Program.cs
```diff
@@ -30,13 +30,13 @@ public static void Main(string[] args)
                 filesConfig.RootPath = @"c:\temp\files";
             }
 
-            config.UseFiles(filesConfig);
+            //config.UseFiles(filesConfig);
```
mathewc: Why are these commented out?

brettsam: That was for testing -- I'll back out.

---
In reply to: [63577537](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63577537) [](ancestors = 63577537)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
```diff
@@ -53,8 +62,194 @@ public void Initialize(ExtensionConfigContext context)
...
+
+            BindingFactory factory = new BindingFactory(nameResolver, converterManager);
+
+            IBindingProvider outputProvider = factory.BindToGenericAsyncCollector<MobileTableAttribute, MobileTableContext>(typeof(MobileTableAsyncCollector<>), BindForOutput);
+
+            IBindingProvider clientProvider = factory.BindToExactType<MobileTableAttribute, IMobileServiceClient>(BindForClient);
+
+            IBindingProvider queryProvider = factory.BindToGenericItem<MobileTableAttribute>(BindForQuery);
+            queryProvider = factory.AddTypeFilter(IsQueryType, queryProvider);
+            queryProvider = factory.AddValidator<MobileTableAttribute>(ThrowIfGenericArgumentIsInvalid, queryProvider);
```
brettsam: Suggestions for Validators/TypeFilters --

1. Can we have a single 'Filter' concept here that passes me all the context I'll need to make a decision. For now, that context is the (%% resolved) Attribute and the parameter Type.

The sig could be Func<TAttribute, Type, bool> and the contract:
- Return true if this matches. Call next filter in the line.
- Return false if this attribute does not match and do not continue. Move onto the next rule.
- Throw if you want to abort indexing of this parameter completely. For example: The type is correct but you haven't set an 'Id' property on the attribute. You don't want any other rule to claim this parameter so you throw.

If all return true, then this rule has been matched.
If any return false, move onto the next rule.

2. I'd expect the Filters to be called before the binding is created -- I think right now the TypeFilter is called before creation and Validators are called after?

3. Since these are wrapping existing BindingProviders, I believe the Validators are called in reverse-order (last one added is the first one called). It would feel more natural if we could build up a list of Filters/Validators and have them called in the order they were added.

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -13,14 +19,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
...
 
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
         public DocumentDBConfiguration()
         {
-            this.ConnectionString = GetSettingFromConfigOrEnvironment(AzureWebJobsDocumentDBConnectionStringName);
-            this.DocumentDBServiceFactory = new DefaultDocumentDBServiceFactory();
+            ConnectionString = GetSettingFromConfigOrEnvironment(AzureWebJobsDocumentDBConnectionStringName);
+            DocumentDBServiceFactory = new DefaultDocumentDBServiceFactory();
```
brettsam: We talked about this elsewhere -- but I wanted to call out that (I think) this is the only places that we still check Environment vars and Configuration directly anymore. It'd be nice to send these calls through a NameResolver (or similar).

MikeStall:
>         /// - If it is any other string, the value of that string is used as the Api Key.

This sounds inconsisntent with line 50 (that it comes from appsettings). Can you clarify?


---
Refers to: src/WebJobs.Extensions.MobileApps/MobileTableAttribute.cs:55 in 5f998a0. [](commit_id = 5f998a053f5acd491694f52ceb479746b08d8c9d, deletion_comment = False)

src/WebJobs.Extensions.MobileApps/Bindings/MobileTableQueryValueProvider.cs
```diff
@@ -2,26 +2,25 @@
...
-using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.WindowsAzure.MobileServices;
 
 namespace Microsoft.Azure.WebJobs.Extensions.MobileApps
 {
-    internal class MobileTableQueryValueProvider<T> : IValueProvider
+    internal class MobileTableQueryValueProvider<T> : IValueBinder
```
MikeStall: Why do you still have ValueBinder if SetValue() is a nop? Can we get rid of this class?

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -41,10 +44,57 @@ public void Initialize(ExtensionConfigContext context)
             {
                 throw new ArgumentNullException("context");
             }
+
+            INameResolver nameResolver = context.Config.NameResolver;
+            IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
             var converterManager = context.Config.GetService<IConverterManager>();
             converterManager = converterManager.AddNotificationHubConverters();
```
MikeStall: This pattern seems wrong.

brettsam:
@pragnagopa is added to the review. #Closed

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
```diff
@@ -53,8 +62,194 @@ public void Initialize(ExtensionConfigContext context)
...
+            HttpMessageHandler[] handlers = null;
+            if (!string.IsNullOrEmpty(apiKey))
+            {
+                handlers = new[] { new MobileServiceApiKeyHandler(apiKey) };
+            }
+
+            return factory.CreateClient(mobileAppUri, handlers);
         }
 
         internal static string GetSettingFromConfigOrEnvironment(string key)
```
MikeStall: Can we avoid the direct dependency on ConfigurationManager / Environment and plumb thing through INameResolver?. 

Or we could extend Attribute Cloner with another form or [AutoResolve]   to handle this. But we really don't want extensions doing this adhoc. NotificationHub does it too; and does it slightly differently.

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
MikeStall: #1 - I can add this helper. 
#2 - Correct. 
#3 - if we do #1, is this still an issue?

---
In reply to: [63583395](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63583395) [](ancestors = 63583395)

src/WebJobs.Extensions.MobileApps/Bindings/MobileTableQueryValueProvider.cs
brettsam: There's no helpers for dealing with IValueProviders -- I'm not sure which is better: having both IValueProvider and IValueBinder or just having IValueBInder but making SetAsync a no-op. I'd like us to narrow down the interfaces we tell people to deal with :-)

I can't remove this class (yet) b/c BindToExactType won't work due to the open generics. Functions can have IMobileServiceTable<T> as a parameter and I want all of those to run the GetValue() code below.

---
In reply to: [63598844](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63598844) [](ancestors = 63598844)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
brettsam: Yes, I called this out elsewhere that we want to remove this. I don't think we have a NameResolver that does this currently. Would we get it from the JobHostConfig?

---
In reply to: [63600341](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63600341) [](ancestors = 63600341)

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
brettsam: The extension method to add converters? @pragnagopa can probably discuss that :-)

---
In reply to: [63599917](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63599917) [](ancestors = 63599917)

pragnagopa: @MikeStall src/WebJobs.Extensions.NotificationHubs/Config/Converter.cs handles conversions.  Having it in a separate class makes it easier for unit testing

MikeStall: A helper method is great. But having it assign to itself is wrong, particularly when we're getting converterManager from the config. 

---
In reply to: [63605331](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63605331) [](ancestors = 63605331)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
MikeStall: You could have the ctor require a INameResolver;  or we could leverage AttributeCloner. 
I believe in ScriptHost (Functions), the configuration is supplied with a NameResolver that does this.  
[https://github.com/Azure/azure-webjobs-sdk-script/blob/21f1cd77957ec9107e68e938ece8af1886becc92/src/WebJobs.Script/Config/NameResolver.cs](https://github.com/Azure/azure-webjobs-sdk-script/blob/21f1cd77957ec9107e68e938ece8af1886becc92/src/WebJobs.Script/Config/NameResolver.cs) 

---
In reply to: [63604093](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63604093) [](ancestors = 63604093,63600341)

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
pragnagopa: Fixed

---
In reply to: [63606377](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63606377) [](ancestors = 63606377,63605331)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -27,13 +27,19 @@ public static class DocumentDBSamples
         //   IAsyncCollector<T>
         //   ICollector<T>
         public static void InsertDocument(
-            [TimerTrigger("00:01")] TimerInfo timer,
-            [DocumentDB("ItemDb", "ItemCollection")] out ItemDoc newItem)
+            [TimerTrigger("00:01", RunOnStartup = true)] TimerInfo timer,
+            [DocumentDB("ItemDb", "ItemCollection")] out ItemDoc newItem,
```
mathewc: If someone wants to output two items to the same collection, it would be better for them to us the collector binding, right? Why are you showing this pattern?

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -45,7 +51,7 @@ public static class DocumentDBSamples
         //
         // This example uses the binding template "{QueueTrigger}" to specify that the Id should come from
         // the string value of the queued item.
-
+        [Disable]
```
mathewc: are these disable attributes to be removed?

src/ExtensionsSample/Samples/NotificationHubSamples.cs
```diff
@@ -16,28 +16,59 @@ namespace ExtensionsSample
...
+            [NotificationHub] out Notification notification)
+        {
+            string toastPayload = "<toast><visual><binding template=\"ToastText01\"><text id=\"1\">Test message</text></binding></visual></toast>";
+            notification = new WindowsNotification(toastPayload);
+        }
+
         // NotificationHub binding out String
-        // The binding sends push notification to any clients registered with the template
-        // method successfully exits.
+        // This binding builds tempalte notification from the string payload and
```
mathewc: spelling errors

src/ExtensionsSample/Samples/NotificationHubSamples.cs
```diff
@@ -49,8 +80,8 @@ public static class NotificationHubSamples
                 };
         }
 
-        //   The binding creates a strongly-typed AsyncCollector, which is used to send push notifications. 
-        //   The binding does not do anything with the results when the function exits.  
+        // This binding creates a strongly-typed AsyncCollector, which is used to send push notifications. 
+        // The binding does not do anything with the results when the function exits.
```
mathewc: this "the binding does not do anything with the results" comment is confusing - I don't understand it, and I doubt users will either :) remcommend removing it here and elsewhere

src/Sample.Extension/packages.config
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.Azure.KeyVault.Core" version="1.0.0" targetFramework="net45" />
-  <package id="Microsoft.Azure.WebJobs" version="1.2.0-alpha-10306" targetFramework="net45" />
-  <package id="Microsoft.Azure.WebJobs.Core" version="1.2.0-alpha-10306" targetFramework="net45" />
+  <package id="Microsoft.Azure.WebJobs" version="2.0.0-alpha-10324" targetFramework="net45" />
```
mathewc: Did you update the nuspecs for the packages to declare that we require this new version?

src/WebJobs.Extensions.DocumentDB/Bindings/CompositeBindingProvider.cs
```diff
@@ -1,33 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
```
mathewc: Yay!

src/ExtensionsSample/Samples/DocumentDBSamples.cs
brettsam: This was actually to verify my new caching code that I had to add. I wanted to make sure it would use the same underlying DocumentClient. I'll back all these changes out.

---
In reply to: [63893134](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63893134) [](ancestors = 63893134)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
brettsam: Yup -- in general, ignore this file. I meant to exclude it from the commit but obviously missed it.

---
In reply to: [63893331](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63893331) [](ancestors = 63893331)

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
```diff
@@ -16,21 +15,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
...
+        public DocumentDBItemValueBinder(DocumentDBContext context)
         {
-            _itemBindingContext = itemBindingContext;
-            _parameter = parameter;
             _context = context;
         }
 
         public Type Type
         {
-            get { return _parameter.ParameterType; }
```
mathewc: Will the parameter type always be the same as T?

src/Sample.Extension/packages.config
brettsam: Wow, no. Good catch. 

Should we file an issue to handle this automatically? We have a script that updates this automatically during the build in our Zumo projects -- we may want to try something similar here.

---
In reply to: [63894366](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63894366) [](ancestors = 63894366)

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -38,7 +45,86 @@ public void Initialize(ExtensionConfigContext context)
                 throw new ArgumentNullException("context");
             }
 
-            context.Config.RegisterBindingExtension(new DocumentDBAttributeBindingProvider(context.Config, this, context.Trace));
+            INameResolver nameResolver = context.Config.GetService<INameResolver>();
+            IConverterManager converterManager = context.Config.GetService<IConverterManager>();
+
+            BindingFactory factory = new BindingFactory(nameResolver, converterManager);
+
+            IBindingProvider outputProvider = factory.BindToGenericAsyncCollector<DocumentDBAttribute, DocumentDBContext>(typeof(DocumentDBAsyncCollector<>),
```
mathewc: Wondering why this generic method has to take the typeof expression? Why can't it also be a generic param?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
brettsam: Do we have documentation about what this property is used for? I was following other patterns but I'm not sure where this gets called.

To answer your question -- yes, if you specify 'MyPoco' as the type you want, we'll construct a DocumentDBItemValueBinder<MyPoco>. I tried to think of a scenario where this wouldn't be the case, but couldn't. 

---
In reply to: [63894947](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63894947) [](ancestors = 63894947)

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
brettsam: B/c we don't know the type of T at compile time. You can't specify DocumentDBAsyncCollector<T> (or use <>) as a generic parameter. So the method will use reflection to create this type for us at runtime.

But this is going away anyway -- it'll be my job to construct the collector myself via reflection. That gets rid of the need for us to care about your constructor parameters. That's in Mike's latest PR.

---
In reply to: [63895687](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63895687) [](ancestors = 63895687)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
```diff
@@ -53,8 +62,194 @@ public void Initialize(ExtensionConfigContext context)
...
+            tableProvider = factory.AddValidator<MobileTableAttribute>(ThrowIfGenericArgumentIsInvalid, tableProvider);
+
+            IBindingProvider itemProvider = factory.BindToGenericItem<MobileTableAttribute>(BindForItem);
+            itemProvider = factory.AddValidator<MobileTableAttribute>(IsItemType, itemProvider);
+
+            IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
+            extensions.RegisterBindingRules<MobileTableAttribute>(ValidateMobileAppUri, nameResolver, outputProvider, clientProvider, jObjectTableProvider, queryProvider, tableProvider, itemProvider);
+        }
+
+        internal bool IsQueryType(Type paramType)
```
mathewc: I'm thinking it would be good to move all this binding stuff out into its own MobileAppsBindingProvider class to seperate concerns. For all of these bindings. Thoughts? That way this config class stays focused on plugging in the extension.

mathewc:
Thanks for taking this on for all the bindings :) I made some minor comments. After those are resolved and you work out the remaining issues with Mike, I'm good with this. We can iterate on it further in subsequent PRs.

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
brettsam: I actually debating going further -- having a separate class per Binding (so one to setup ItemBinding, one to setup QueryBinding,one to set up OutputBinding etc). So I think I'd end up with 6 small classes from this, which is still a massive improvement over the classes I had before. 

To me, that makes it clearer where my functionality lies. And when I want to support something new, it's clear that I'd need to add a new class with the logic, and then wire it up with a line of code in the Config.

I've discussed that with Mike and I'll see what it looks like -- but not in this PR.

---
In reply to: [63897796](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63897796) [](ancestors = 63897796)

mathewc: Yeah, I think we should promote a clean class separation pattern. People will end up doing that themselves anyways to keep things straight, rather than a mega class with a bunch of little possibly unrelated funcs.

src/Sample.Extension/packages.config
mathewc: I'd be all for something that at a minimum caught this issue automatically. I'm the only person who catches it it seems - *every* review where someone updates the refs, I have to remind them of this :)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
MikeStall: I think the new rules in the SDK addresses these.

---
In reply to: [63600657](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63600657) [](ancestors = 63600657,63583395)

src/WebJobs.Extensions.MobileApps/Bindings/MobileTableQueryValueProvider.cs
MikeStall: The latest rules should help here and let you skip IValueBinder. 

---
In reply to: [63603726](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63603726) [](ancestors = 63603726,63598844)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
MikeStall: This is a fine thing to test. 


---
In reply to: [63894581](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63894581) [](ancestors = 63894581,63893134)

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
MikeStall: New rule is in the SDK, so this can be updated. 

---
In reply to: [63896758](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63896758) [](ancestors = 63896758,63895687)

src/ExtensionsSample/Samples/DocumentDBSamples.cs
mathewc: Right, these are samples, not tests :) The samples show people the patterns to follow, hence my comment

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
brettsam: We decided to do this in a follow-up PR.

---
In reply to: [63607453](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/97#discussion_r63607453) [](ancestors = 63607453,63604093,63600341)

Issue #67, opened 2016-03-30
For all the new extensions, rename ConnectionString attribute prop (and others)


mathewc:
New extensions like ApiHub (Key), DocumentDB (ConnectionString), MobileTables (ApiKey), NotificationHub (ConnectionString) need to rename their attribute properties to make it clearer that these properties contain app setting names that point to the actual value.

The recommendation is to change "ConnectionString" to just "Connection" as in the script runtime.

mathewc:
@brettsam @safihamid

brettsam:
I'm not sure 'Connection' works for all 'key name' attributes. For example, MobileTables has both ApiKey and MobileTableUri that are lookup values. In reality, MobileTableUri doesn't really need to be stored in an app setting (since it's not a secret), but we eventually want a picker for it. ApiKey likely will not have a picker unless Mobile Apps makes it an official feature.

What if we suffixed all of these settings with 'Setting'?
ApiKey -> ApiKeySetting
MobileTableUri -> MobileTableUriSetting
ConnectionString -> ConnectionStringSetting

Thoughts?

Issue #52, opened 2016-03-17
Reduce code duplication across extensions for connection/env access


mathewc:
Virtually all the extensions are duplicating this logic, e.g. https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.EasyTables/Config/EasyTablesConfiguration.cs#L66

We need to add a single shared helper, either in this repo, or in core.

brettsam:
Dupe of this: https://github.com/Azure/azure-webjobs-sdk/issues/653

PR #98, opened 2016-05-17
ApiHub extension support for tabular connectors (no triggers yet).

emilcicos:


emilcicos:
I think I also have to increment the ApiHub SDK version, make it available on the feed and update the solution to use it. Will update the pull request.

src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -54,7 +54,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.Azure.ApiHub.Sdk, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.Azure.ApiHub.Sdk.0.6.5-alpha\lib\net45\Microsoft.Azure.ApiHub.Sdk.dll</HintPath>
+      <HintPath>..\..\packages\Microsoft.Azure.ApiHub.Sdk.0.6.6-alpha\lib\net45\Microsoft.Azure.ApiHub.Sdk.dll</HintPath>
```
mathewc: Any time you rev your required version, you have to update the nuspec accordingly: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.ApiHub.NuGet/WebJobs.Extensions.ApiHub.nuspec#L18

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
```diff
@@ -0,0 +1,80 @@
...
+    //      CREATE TABLE table1
+    //      (
+    //          c1 int NOT NULL,
+    //          c2 nvarchar(10) NULL
+    //          CONSTRAINT PK_table1 PRIMARY KEY(c1)
+    //      )
+    public static class ApiHubTableSamples
+    {
+        public static void BindToTableClient(
+            [ApiHubTable("AzureWebJobsSql")]
```
mathewc: I think rather than having the connection name repeated on each binding, you should instead follow the pattern of the other bindings (e.g. DocumentDB https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs#L65) which allows the default connection to be specified on the global config, and allows per attribute overrides via the attribute property.

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
```diff
@@ -0,0 +1,80 @@
...
+            ITableClient tableClient)
+        {
+            // Use the table client.
+        }
+
+        public static void BindToTableOfJObject(
+            [ApiHubTable("AzureWebJobsSql", TableName = "[dbo].[table1]")]
```
mathewc: The table name doesn't have to be in this full form does it? Can it be just "table1"?

Also a nit for these samples: perhaps you can name the test table "Samples" to match your SampleEntity?

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <summary>
+    /// Attribute that specifies an ApiHub table or entity.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class ApiHubTableAttribute : Attribute
+    {
+        /// <summary>
+        /// Gets the key of the configuration setting that specifies the connection string.
+        /// </summary>
+        public string Key { get; }
```
mathewc: Please rename this to "Connection". Several of the other attributes are using "ConnectionString" but our plan is to rename those as per https://github.com/Azure/azure-webjobs-sdk-extensions/issues/67

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// Gets the entity identifier.
+        /// </summary>
+        public string EntityId { get; set; }
+
+        /// <summary>
+        /// Creates a new instance of this class.
+        /// </summary>
+        /// <param name="key">The key of the configuration setting that 
+        /// specifies the connection string.</param>
+        public ApiHubTableAttribute(string key)
```
mathewc: I don't think this should be a required param - instead, if not specified on attribute, default it to the connection specified on the config object.

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
```diff
@@ -0,0 +1,80 @@
...
+    //      (
+    //          c1 int NOT NULL,
+    //          c2 nvarchar(10) NULL
+    //          CONSTRAINT PK_table1 PRIMARY KEY(c1)
+    //      )
+    public static class ApiHubTableSamples
+    {
+        public static void BindToTableClient(
+            [ApiHubTable("AzureWebJobsSql")]
+            ITableClient tableClient)
```
mathewc: The table client doesn't require a table name to bind to? Does that mean it isn't bound to a particular table?

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public static void BindToTableOfJObject(
+            [ApiHubTable("AzureWebJobsSql", TableName = "[dbo].[table1]")]
+            ITable<JObject> table)
+        {
+            // Use the table.
+        }
+
+        public static void BindToTableOfSampleEntity(
+            [ApiHubTable(key: "AzureWebJobsSql", TableName = "[dbo].[table1]")]
```
mathewc: See attribute comments. I think you should make the connection name a property, not a ctor param.

src/WebJobs.Extensions.ApiHub/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ApiHub.Extensions
+{
+    internal static class TypeExtensions
```
mathewc: Add unit tests for any new helper classes like this

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Attribute that specifies an ApiHub table or entity.
```
mathewc: "Attribute used to bind a parameter to an ApiHub table or entity."

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class ApiHubTableAttribute : Attribute
+    {
+        /// <summary>
+        /// Gets the key of the configuration setting that specifies the connection string.
+        /// </summary>
+        public string Key { get; }
+
+        /// <summary>
+        /// Gets the data set name.
```
mathewc: "Gets or sets" here and for others below

src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
```diff
@@ -23,6 +24,17 @@ public class ApiHubConfiguration : IExtensionConfigProvider, IFileTriggerStrateg
 
         ApiHubLogger _logger;
 
+        private ConnectionFactory ConnectionFactory { get; }
+
+        /// <summary>
+        /// Creates a new instance of this class.
+        /// </summary>
+        /// <param name="connectionFactory">The factory used to create connections</param>
+        public ApiHubConfiguration(ConnectionFactory connectionFactory = null)
```
mathewc: Let's keep anything like ConnectionFactory internal. We've got some open tracking items for improving the connection/setting access across all the extensions, and having a shared mechanism for doing this. See https://github.com/Azure/azure-webjobs-sdk-extensions/issues/52

So it would be good not to introduce a new public type that we'll have to remove. However, the fact that you went through the effort to create this underscores the need we have for a central solution :) See how the other extensions are accessing settings and follow suit for now.

src/WebJobs.Extensions.ApiHub/Table/TableBinding.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            return BindAsync(null, context.ValueContext);
+        }
+
+        public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+        {
+            if (context == null)
```
mathewc: You're not using value anywhere here, which means you're not yet supporting string invoke scenarios which are how our portal/Dashboard "Run" button functionality works.

src/WebJobs.Extensions.ApiHub/Table/TableClientValueProvider.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Reflection;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ApiHub.Table
+{
+    internal class TableClientValueProvider : IValueProvider
```
mathewc: nit: might consider making this a private class of the binding. we do that for the other bindings to reduce the number of files.

src/WebJobs.Extensions.ApiHub/Table/TableBinding.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public ParameterDescriptor ToParameterDescriptor()
+        {
+            return new ParameterDescriptor
+            {
+                Name = Parameter.Name
+            };
+        }
+
+        private static void ValidateParameter(ParameterInfo parameter)
```
mathewc: Please add unit tests for all of these Validate methods. They're doing non trivial string formatting, so it would be great to have coverage in place.

src/WebJobs.Extensions.ApiHub/Table/TableEntityValueBinder.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
```
mathewc: All of your new files need file headers. Aren't you getting build warnings/errors for this? You should be.

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
emilcicos: Yes, it is not bound to a table. That is the topmost object in the ApiHub SDK for tabular connectors. From there we can navigate to data sets (sets of tables), tables, entities.

src/WebJobs.Extensions.ApiHub/Table/TableClientValueProvider.cs
emilcicos: I like the one class per file approach. I won't change unless you feel strongly about it.

src/WebJobs.Extensions.ApiHub/Table/TableBinding.cs
emilcicos: There are unit tests that cover the validate methods, see ApiHubTableBindingTests: ProviderThrowsIfInvalidTableClientBinding, ProviderThrowsIfInvalidTableBinding, ProviderThrowsIfInvalidEntityBinding. They don't validate the exception message but they ensure that the validations are performed as expected. I could add validation of the exception message, however my understanding is that Mike has a new way to do bindings which does some validation so I prefer to to invest time in something that will change.

src/WebJobs.Extensions.ApiHub/Table/TableEntityValueBinder.cs
emilcicos: Nope, no build errors. But sure, I can add headers where missing.

src/ExtensionsSample/ExtensionsSample.csproj
emilcicos: Yup, I missed that.

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
emilcicos: Yes, it can be "table1". Will update.

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
emilcicos: I wanted to do that but then opted for keeping it consistent with the ApiHubFileAttribute. Will change to Connection.

src/WebJobs.Extensions.ApiHub/Extensions/TypeExtensions.cs
emilcicos: Ok

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
emilcicos: Ok.

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
emilcicos: Yes, I missed that

src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
emilcicos: It seemed that both the extensions and script engine lacks testability features. A bunch of objects (services) are new-ed inside without the ability to replace with mocks/fakes etc. This was the minimum I could do to improve that and it is great because nobody has to worry about setting up environment variables and SqlAzure tables to run the ApiHub extensions tests for tabular connectors. The environment variable variable approach is obscure and counter-productive. Each team has to know about the other teams' specific details to be able to run tests. I would prefer to keep this until a better general solution exists. Removing stuff is usually easy.

src/WebJobs.Extensions.ApiHub/Table/TableBinding.cs
emilcicos: Yes, I left that out on purpose until I figure out the details around it.

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
emilcicos: The problem is that there could be a bunch of ApiHub tabular connector types (Sql, GoogleSheets, Excel, etc) and a few connections for each connector type, all used in the same function app. Maybe we could invent some notion of default but I am not sure how useful that will be.

src/WebJobs.Extensions.ApiHub/Table/ApiHubTableAttribute.cs
emilcicos: A default connection in the ApiHub case may not make sense, so required is probably ok. But sure, if we find a reason to introduce some sort of default then I can change.

src/ExtensionsSample/Samples/ApiHubTableSamples.cs
MikeStall: I think this is ok for now and we can see how it feels. Connections are so broad that the common case will be working with multiple connections. 
Cost of failure here is small: we can always add a default to the global config later. 


---
In reply to: [64059728](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/98#discussion_r64059728) [](ancestors = 64059728)

src/WebJobs.Extensions.ApiHub/Table/TableClientValueProvider.cs
MikeStall: even for nested classes? It works out pretty nicely in that case. 

---
In reply to: [64083666](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/98#discussion_r64083666) [](ancestors = 64083666)

src/WebJobs.Extensions.ApiHub/Table/TableEntityValueBinder.cs
MikeStall: Can you enable StyleCop (see the example from the other projects in the repo)?  

---
In reply to: [64084638](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/98#discussion_r64084638) [](ancestors = 64084638)

mathewc: I'm enabling it now

src/WebJobs.Extensions.ApiHub/Table/TableClientValueProvider.cs
mathewc: Yeah, for bindings people usually hide the IValueProvider and other simple classes as nested classes to avoid having too many files. If you *really* want the separate classes, so be it, but all the other bindings are using nested classes (hint hint :))

emilcicos: Alright, will nest.

src/WebJobs.Extensions.ApiHub/Table/TableBinding.cs
mathewc: You should log a bug to track the remaining work. Without this, these functions won't work with the portal UI

mathewc:
:ship:

mathewc:
Merged

PR #101, opened 2016-05-23
MobileTable: allowing objects as output parameter

brettsam:


test/WebJobs.Extensions.Tests/Extensions/MobileApps/MobileTableEndToEndTests.cs
```diff
@@ -242,6 +244,10 @@ private class MobileTableEndToEndFunctions
                 [MobileTable] out TodoItem[] arrayPoco,
                 [MobileTable] IAsyncCollector<TodoItem> asyncCollectorPoco,
                 [MobileTable] ICollector<TodoItem> collectorPoco,
+                [MobileTable(TableName = tableName)] out object newObject, // we explicitly allow object
+                [MobileTable(TableName = tableName)] out object[] arrayObject,
```
pragnagopa: Looks good. Do we need negative tests to verify invalid outputs: [MobileTable] out object newObject ?

mathewc:
Supporting object opens the door for people to try using instances of IAsyncCollector<IAsyncCollector<object>>, etc. I know your intent is for people to use object for sane things like poco/dynamic, but how do you handle things that aren't supported.?

test/WebJobs.Extensions.Tests/Extensions/MobileApps/MobileTableEndToEndTests.cs
brettsam: Good catch. I wanted to fail if they did that, but I wasn't. I've added that check and a new test.

---
In reply to: [64283568](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/101#discussion_r64283568) [](ancestors = 64283568)

brettsam:
My PR description is not very good -- MobileTable has supported 'object' for a while (for output bindings only), as has DocDB. My latest changes to move to the new BindingFactory broke this support b/c I added a bunch of 'fail at index-time' checks and was a little too strict. This relaxes it so it continues to work. Without support for 'object', any usage of [MobileTable] in Functions requires references to JSON.NET or creating your own POCO.

We already support POCOs and JObject, this just makes it easier to turn any object into a JObject (which is what we do behind the scenes if you use 'object' directly as the type). So if you try to use something goofy, it'll likely fail for you during JObject.FromObject(). And if it passes that, then you're sending the server valid JSON, which is a supported scenario.

DocDB supports it 'natively' in their SDK -- the 'CreateDocument()' call actually takes an object, so they do validation on it either there or on the server to make sure it's valid JSON. If it is (and has an 'id', I believe), it gets inserted.

---
In reply to: [221308594](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/101#issuecomment-221308594) [](ancestors = 221308594)

mathewc:
:ship:

PR #102, opened 2016-05-23
apihub triggers should not stop working if the trigger path does not exist

safihamid:


safihamid:
ping

mathewc:
:ship:

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -57,13 +57,11 @@ public Task StartAsync(CancellationToken cancellationToken)
 
             if (_poll == null)
             {
-                var errorText = string.Format("Path '{0}' is invalid. IFolderItem.RootPath must be set to a valid directory location.", _folderSource.Path);
-                _trace.Error(errorText);
-
-                throw new InvalidOperationException(errorText);
+                var errorText = string.Format("Path '{0}' is invalid. Path must be set to a valid directory location.", _folderSource.Path);
+                var ex = new InvalidOperationException(errorText);
+                _trace.Error(errorText, ex);
```
mathewc: In the context of a standard SDK extension, if the listener for a function cannot start, that shouldn't be ignored - it is an exceptional situation. This really should be an index time error. Is there a way for you to validate this path earlier?

safihamid: The idea was the user will get an error to fix the path but the host will start in case there are other functions/triggers. currently nothing gets started and all other functions/triggers in the function app will get blocked. no I can't validate the path any earlier than this. this is called when the json file gets saved or read after the function app starts.

mathewc: Would any of the APIs like GetFolderReferenceAsync or ListAsync allow you to validate the path early w/o creating the file watcher up front?

safihamid: GetFolderreferenceAsync does not hit the backend and just returns an IFolder item same design as storage SDK. ListAsync returns list of the files which not sure why it can be used here.

The only way to check if the path is valid before creating the FileWatcher(i.e. saving the json file) is to have the portal call FolderExists when a path is entered in the 'Path' textbox. but then it means the portal will need to take a dependency on ApiHub SDK and all the complexities which come with it.

mathewc: There is no portal or anything in the context of this repo :) This extension should be designed to stand alone in a regular SDK WebJob, outside of Functions. The right behavior in that context is to bring down the host if there are any errors in starting up the job methods.

I think the right thing for us to do is make a change to the core SDK to allow listener exceptions to be "handled". I did something similar for indexing exceptions (in this [commit](https://github.com/Azure/azure-webjobs-sdk/commit/4842e35d5b6b780904ec484b46b8aea063fd566b)), to allow the Functions runtime to mark indexing exceptions as "handled" - they normally bring down the host. This listener case is similar I think.

safihamid: Ok making the changes in the core SDK to allow listener exceptions to be "handled" is very reasonable. Please let me know if you want me to close this PR then.

mathewc: Let's leave this open for now.

PR #105, opened 2016-05-25
Fix for stylecop errors.

emilcicos:
Moved test class at the end of file as suggested by stylecop.


mathewc:
Checked in the fix.

PR #108, opened 2016-05-31
Renaming all key lookup attribute properties to end with 'Setting'

brettsam:
Addresses #67.

@safihamid @emilcicos FYI -- I've renamed one of the ApiHubAttribute properties (Key -> ConnectionStringSetting) to match the convention that I'm introducing here. I'm not 100% sure that's the right name. I'll wait for your 'OK' before merging.

### DIFF
--- a/src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
+++ b/src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Azure.WebJobs
         /// Optional. A string value indicating the app setting to use as the DocumentDB connection string, if different
         /// than the one specified in the <see cref="DocumentDBConfiguration"/>.
         /// </summary>
-        public string ConnectionString { get; set; }
+        public string ConnectionStringSetting { get; set; }
 
         /// <summary>
         /// Optional. The Id of the document to retrieve from the collection.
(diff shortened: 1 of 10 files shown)


safihamid:
looks good to me.

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
```diff
@@ -64,7 +64,7 @@ public DocumentDBAttribute(string databaseName, string collectionName)
         /// Optional. A string value indicating the app setting to use as the DocumentDB connection string, if different
         /// than the one specified in the <see cref="DocumentDBConfiguration"/>.
         /// </summary>
-        public string ConnectionString { get; set; }
+        public string ConnectionStringSetting { get; set; }
```
mathewc: The value on DocuemntDBConfiguration is an actual connection string, or is it a setting as well? Same question for notification hubs

mathewc: Also, I wonder if we can't simplify this to just "ConnectionSetting"? Thoughts? Since these attributes are used inline in source, the more concise we can make them, the easier to read.

brettsam: Those are actual ConnectionStrings. I sent a mail with the breakdown of the pattern we all use, but to simplify -- the Configuration classes take all real values, the attributes all take Settings.

---
In reply to: [65440424](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/108#discussion_r65440424) [](ancestors = 65440424)

mathewc:
:ship:

Issue #100, opened 2016-05-20
Need Automated way of keeping Nuspecs up to date


mathewc:
Too often we're moving to a new required version of the SDK or other packages but forgetting to reflect that in the nuspec.

One simple way to handle this is to write a static analysis test that reads the nuspecs and package files and does some comparisons.

We should employ this same strategy across all our WebJobs repos.

brettsam:
Any objections if I take this one?

PR #109, opened 2016-05-31
Ensuring that nuspecs and packages.config contain matching versions

brettsam:
Addresses #100. The build will now fail if a dependency in a nuspec has a different version than one in a packages.config. This actually found two mismatches that I've fixed.


mathewc:
Awesome, thanks. We'll try this out in Extensions and if it looks like it is working, we can apply to Script as well.

:ship:

PR #113, opened 2016-06-14
Replat SendGrid binding on the new Helpers, and onboarding to Functions

mathewc:
Addresses issue https://github.com/Azure/azure-webjobs-sdk-script/issues/132

I'm doing this work now to further vet the new Functions extensibility model. In this case, with these changes, I'm able to get the SendGrid binding working in Functions with **no changes** to the Functions runtime itself.

In addition to adding the `ScriptBindingProvider` to make it work, the other key changes was to add a `JObject` binding, which allows `SendGridMessage` to be exposed to Node.js functions.


src/ExtensionsSample/Samples/SendGridSamples.cs
```diff
@@ -20,32 +19,63 @@ public static class SendGridSamples
                 To = "{CustomerEmail}",
                 Subject = "Thanks for your order (#{OrderId})!",
                 Text = "{CustomerName}, we've received your order ({OrderId}) and have begun processing it!")]
-            SendGridMessage message)
+            out SendGridMessage message)
```
mathewc: This is a breaking change over the first release of this extension. However our next release we're revving our major version, so this is OK - all the functionality remains if you change to out.

src/ExtensionsSample/Samples/SendGridSamples.cs
```diff
@@ -20,32 +19,63 @@ public static class SendGridSamples
...
         /// <summary>
-        /// Demonstrates imperatively defining all email properties in the function body.
-        /// Also demonstrates use of a 'ref' parameter, allowing you to decide in your
-        /// method whether to send the message.
+        /// Demonstrates imperatively setting email properties inline in the function.
         /// </summary>
+        [Disable]
         public static void ProcessOrder_Imperative(
             [QueueTrigger(@"samples-orders")] Order order,
-            [SendGrid] ref SendGridMessage message)
```
mathewc: Same here - we no longer support `ref`. To conditionally send messages, you use `IAsyncCollector` which is a new addition with this replat. See the example below.

src/WebJobs.Extensions.SendGrid/Config/SendGridJobHostConfigurationExtensions.cs
```diff
@@ -48,7 +54,26 @@ public void Initialize(ExtensionConfigContext context)
...
+                if (string.IsNullOrEmpty(_sendGridConfig.ApiKey))
+                {
+                    throw new InvalidOperationException(
+                        string.Format("The SendGrid ApiKey must be set either via a '{0}' app setting, via a '{0}' environment variable, or directly in code via SendGridConfiguration.ApiKey.",
+                        SendGridConfiguration.AzureWebJobsSendGridApiKeyName));
+                }
+                _sendGrid = new Web(_sendGridConfig.ApiKey);
+
+                IConverterManager converterManager = context.Config.GetService<IConverterManager>();
+                converterManager.AddConverter<JObject, SendGridMessage>(SendGridHelpers.CreateMessage);
```
mathewc: This JObject->SendGridMessage converter is the key that allows this binding to work for Node and other languages.

PR #114, opened 2016-06-21
Support case insensitive bindings in ApiHub

safihamid:


mathewc:
:ship:

src/WebJobs.Extensions.ApiHub/Table/TableBinding.cs
```diff
@@ -30,12 +30,12 @@ internal class TableBinding : IBinding
             ApiHubTableAttribute attribute = Parameter.GetTableAttribute();
             if (!string.IsNullOrEmpty(attribute.DataSetName))
             {
-                _dataSetNameBindingTemplate = BindingTemplate.FromString(attribute.DataSetName);
+                _dataSetNameBindingTemplate = BindingTemplate.FromString(attribute.DataSetName, ignoreCase: true);
```
safihamid: Yes, we discussed this and looks like they are all case insensitive.

PR #115, opened 2016-06-28
Update to have Apihub file triggers save latest trigger status in case of down times

safihamid:
This is needed for Antares scale controller as well.

### DIFF
--- a/src/ExtensionsSample/Samples/ApiHubSamples.cs
+++ b/src/ExtensionsSample/Samples/ApiHubSamples.cs
@@ -96,7 +96,7 @@ namespace ExtensionsSample.Samples
             if (!string.IsNullOrEmpty(apiHubConnectionString))
             {
                 var apiHubConfig = new ApiHubConfiguration();
-                apiHubConfig.AddKeyPath("dropbox", apiHubConnectionString);
+                apiHubConfig.AddConnection("dropbox", apiHubConnectionString);
                 config.UseApiHub(apiHubConfig);
 
                 // Create some initialization files.
--- a/src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
+++ b/src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Azure.WebJobs
             extensions.RegisterExtension<IBindingProvider>(bindingProvider);
 
             var triggerBindingProvider = new GenericFileTriggerBindingProvider<ApiHubFileTriggerAttribute, ApiHubFile>(
-                BuildListener, bindingProvider, this, context.Trace);
+                BuildListener, config, bindingProvider, this, context.Trace);
             extensions.RegisterExtension<ITriggerBindingProvider>(triggerBindingProvider);
 
             extensions.RegisterExtension<IBindingProvider>(
@@ -93,7 +93,7 @@ namespace Microsoft.Azure.WebJobs
                         nameResolver)));
         }
 
-        private async Task<IListener> BuildListener(ApiHubFileTriggerAttribute attribute, ITriggeredFunctionExecutor executor, TraceWriter trace)
+        private async Task<IListener> BuildListener(JobHostConfiguration config, ApiHubFileTriggerAttribute attribute, string functionName, ITriggeredFunctionExecutor executor, TraceWriter trace)
         {
             var root = GetFileSource(attribute.ConnectionStringSetting);
 
@@ -117,7 +117,7 @@ namespace Microsoft.Azure.WebJobs
 
             var folder = await root.GetFolderReferenceAsync(folderName);
 
-            var listener = new ApiHubListener(folder, executor, trace, attribute.PollIntervalInSeconds, attribute.FileWatcherType);
+            var listener = new ApiHubListener(config, folder, functionName, executor, trace, attribute);
 
             return listener;
         }
@@ -136,14 +136,14 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <summary>
-        /// Add path to the configuration
+        /// Adds a connection to the configuration
         /// </summary>
-        /// <param name="key">App settings key name that have the connections string</param>
+        /// <param name="settingName">App settings settingName name that have the connections string</param>
         /// <param name="connectionString">Connection string to access SAAS via ApiHub. <seealso cref="ApiHubJobHostConfigurationExtensions.GetApiHubProviderConnectionStringAsync" /></param>
-        public void AddKeyPath(string key, string connectionString)
+        public void AddConnection(string settingName, string connectionString)
         {
             var root = ItemFactory.Parse(connectionString, _logger);
-            _map[key] = root;
+            _map[settingName] = root;
         }
 
         string IFileTriggerStrategy<ApiHubFile>.GetPath(ApiHubFile file)
--- a/src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
+++ b/src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.ApiHub;
@@ -9,6 +10,9 @@ using Microsoft.Azure.WebJobs.Extensions.ApiHub.Common;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.WebJobs.Extensions.ApiHub
 {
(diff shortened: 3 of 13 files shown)


safihamid:
Ping

src/ExtensionsSample/Samples/ApiHubSamples.cs
```diff
@@ -96,7 +96,7 @@ public static void UseApiHub(JobHostConfiguration config)
             if (!string.IsNullOrEmpty(apiHubConnectionString))
             {
                 var apiHubConfig = new ApiHubConfiguration();
-                apiHubConfig.AddKeyPath("dropbox", apiHubConnectionString);
+                apiHubConfig.AddConnection("dropbox", apiHubConnectionString);
```
mathewc: Thanks :)

src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
```diff
@@ -136,11 +136,11 @@ private IFolderItem GetFileSource(string key)
         }
 
         /// <summary>
-        /// Add path to the configuration
+        /// Adds a connection to the configuration
         /// </summary>
         /// <param name="key">App settings key name that have the connections string</param>
```
mathewc: Might consider renaming "key" to "settingName"

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -9,35 +9,53 @@
...
 
 namespace Microsoft.Azure.WebJobs.Extensions.ApiHub
 {
     // TODO: leveraging singleton listener for now until the framework supports a generalized listener/queue mechanism.
     [Singleton(Mode = SingletonMode.Listener)]
     internal class ApiHubListener : IListener
     {
-        private IFolderItem _folderSource;
-        private ITriggeredFunctionExecutor _executor;
+        // Need to be kept in sync with the Antares Scale Controller
```
mathewc: Recommend removing this comment. Let's keep the extensions code independent of outside particulars.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -51,31 +69,67 @@ public void Dispose()
             // nop
         }
 
-        public Task StartAsync(CancellationToken cancellationToken)
+        public async Task StartAsync(CancellationToken cancellationToken)
         {
-            _poll = _folderSource.CreateFileWatcher(_fileWatcherType, OnFileWatcher, nextItem: null, pollIntervalInSeconds: _pollIntervalInSeconds > 0 ? _pollIntervalInSeconds : DefaultPollIntervalInSeconds);
+            CloudStorageAccount account;
+            if (!CloudStorageAccount.TryParse(this._config.StorageConnectionString, out account))
```
mathewc: You don't need this validation. If this connection was invalid, the core SDK startup would have errored out much earlier

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -51,31 +69,67 @@ public void Dispose()
...
+            CloudBlobClient blobClient = account.CreateCloudBlobClient();
+
+            string apiHubBlobDirectoryPath = string.Format(ApiHubBlobDirectoryPathTemplate, this._siteName);
+            this._apiHubBlobDirectory = blobClient.GetContainerReference(HostContainerName).GetDirectoryReference(apiHubBlobDirectoryPath);
+
+            // Need to start polling from the point where it was left off. if this is the first time then lastPoll will be null.
+            var lastPoll = await GetLastPollStatusAsync();
+
+            Uri nextUri = null;        
+            Uri.TryCreate(lastPoll, UriKind.Absolute, out nextUri);
```
mathewc: You're using TryCreate presumably because this might fail? If so, then recommend handling if this returns false

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -51,31 +69,67 @@ public void Dispose()
...
+                if (uri != null)
+                {
+                    pollStatus = uri.AbsoluteUri;
+                }
+                else
+                {
+                    pollStatus = obj.ToString();
+                }
+
+                // If function successfully completes then the next poll Uri will be logged in an Azure Blob.
```
mathewc: Same here - please remove all particulars about Scale Controllers, etc. This is a general extension that is used in contexts outside of Functions.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -51,31 +69,67 @@ public void Dispose()
...
                 TriggerValue = apiHubFile
             };
-            return _executor.TryExecuteAsync(input, CancellationToken.None);
+
+            var functionResult = await _executor.TryExecuteAsync(input, CancellationToken.None);
+
+            if (functionResult.Succeeded)
+            {
+                string pollStatus = null;
+                var uri = obj as Uri;
```
mathewc: What is this obj? Where does it come from and what is its format?

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -86,5 +140,68 @@ private Task OnFileWatcher(IFileItem file, object obj)
                 _poll = null;
             }
         }
+
+        private async Task<string> GetLastPollStatusAsync()
```
mathewc: You need to add some tests for this code. Recommend adding some tests that exercise Get/Set to verify that the blobs are created correctly, etc.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -86,5 +140,68 @@ private Task OnFileWatcher(IFileItem file, object obj)
...
+            // Path to the apiHub blob is:
+            // apihubs/{siteName}/{functionaName}/{triggerPath}/{connectionStringSetting}
+            // This needs to be kept in sync with Antares Scale Controller
+            string blobName = string.Format("{0}/{1}/{2}", this._functionName, this._folderSource.Path, this._connectionStringSetting);
+            return this._apiHubBlobDirectory.GetBlockBlobReference(blobName);
+        }
+
+        private static string GetSiteName(JobHostConfiguration config)
+        {
+            var siteName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
```
mathewc: I think you should just be using config.HostId, not getting the site name. Outside of functions you want HostId, and in Functions, config.HostId will be set appropriately already, right?

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -86,5 +140,68 @@ private Task OnFileWatcher(IFileItem file, object obj)
...
+            {
+                var errorText = string.Format("Encountered an exception while setting the next poll status to the blob storage. Path: {0}, Message: {1} ", this._folderSource.Path, e.Message);
+                _trace.Error(errorText);
+            }
+        }
+
+        private CloudBlockBlob GetBlobReference()
+        {
+            // Path to the apiHub blob is:
+            // apihubs/{siteName}/{functionaName}/{triggerPath}/{connectionStringSetting}
```
mathewc: nit: spelling mistakes

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
safihamid: If this returns false, then I just want to pass null as the value of nextUri. which is already set.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
safihamid: This obj comes from the callback being returned from CreateFileWatcher. if it is a real connector, it will be a Uri object and if it is against local files it will just be a DateTime object.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
safihamid: So here is my question, inside of Functions, HostId will always be set to sitename?

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
mathewc: Got it

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
mathewc: Here is the logic: https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/src/WebJobs.Script.WebHost/App_Start/AutofacBootstrap.cs#L24

safihamid: Great!

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
safihamid: @mathewc  This is now done.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -86,5 +132,53 @@ private Task OnFileWatcher(IFileItem file, object obj)
...
+            {
+                var errorText = string.Format("Encountered an exception while setting the next poll status to the blob storage. Path: {0}, Message: {1} ", this._folderSource.Path, e.Message);
+                _trace.Error(errorText);
+            }
+        }
+
+        private CloudBlockBlob GetBlobReference()
+        {
+            // Path to apiHub blob is:
+            // apihubs/{siteName}/{functionName}/{triggerPath}/{connectionStringSetting}
```
mathewc: It seems odd to me that triggerPath/connectionString would be part of the path. Why isn't the function name sufficient? A function maps to a single path/connection. Same as timer trigger - I don't include the timer schedule as part of the path.

I'd expect the path to be `apihubs/{siteName}/{functionName}/status`, using the name "status" for a blob file that contains the status info. Also recommend that you read/write JSON to the file so this is extensible in the future if you need to add more properties. For now it would just have your single property.

safihamid: Hmm isn't it possible for a function to have multiple triggers? I think I saw the support for it in the portal. I was thinking a function might have multiple input triggers and it would be possible for triggers to use the same path or connection and that's why made it like above to guarantee uniqueness. good point on JSON...

safihamid:
@mathewc plz let me know if there are any other feedbacks or if this can be merged now. thanks

test/WebJobs.Extensions.Tests/Extensions/ApiHub/ApiHubBindingEndToEndTests.cs
```diff
@@ -243,6 +323,16 @@ private void DeleteTestFiles(string path)
             }
         }
 
+        private async Task DeleteApiHubBlobsAsync()
+        {
+            await GetBlobReference("ImportTestJob").DeleteIfExistsAsync();
```
mathewc: You should just do a blob directory listing on apihubs/{siteName} and iterate over all blobs found and delete them. That way you don't have to remember to add string literals for each new function (they can get out of sync).

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -86,5 +133,83 @@ private Task OnFileWatcher(IFileItem file, object obj)
...
+                {
+                    _serializer.Serialize(stringWriter, status);
+                    statusLine = stringWriter.ToString();
+                }
+
+                await apiHubBlob.UploadTextAsync(statusLine);
+            }
+            catch (Exception e)
+            {
+                var errorText = string.Format("Encountered an exception while setting the next poll status to the blob storage. Path: {0}, Message: {1} ", this._folderSource.Path, e.Message);
```
mathewc: would be good not to repeat this same literal 3+ times. recommend moving to resource file

mathewc:
I made a couple small comments. Other than that looks ok.

PR #117, opened 2016-07-07
moving MobileApp and DocDB 'Setting' properties to use [AutoResolve];…

brettsam:
… improving test coverage


mathewc:
:ship:

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
```diff
@@ -64,6 +64,7 @@ public DocumentDBAttribute(string databaseName, string collectionName)
         /// Optional. A string value indicating the app setting to use as the DocumentDB connection string, if different
         /// than the one specified in the <see cref="DocumentDBConfiguration"/>.
         /// </summary>
+        [AutoResolve(AllowTokens = false)]
         public string ConnectionStringSetting { get; set; }
```
mathewc: It is somewhat strange for such AutoResolve properties that the property name is ConnectionStringSetting, but the value after resolution is a connection string. I.e. you're usint resolvedAttr.ConnectionStringSetting as a connection string. Oh well.

brettsam: I agree with you completely. Mike and I had discussed a few options, like (a) requiring a sister 'ConnectionString' property that we'd auto-populate, or (b) giving you a wrapper that contained your original attribute plus a dictionary with all the replacements so you could look them up.

The good news is that this gets rid of all the calls to AppSetting/Environment from the extensions. The bad news is that it still doesn't feel quite right. That's part of the bigger refactor I was trying to do before I reeled it back a bit. I think that'll be a bigger work item that we can tackle later if we see that it's worthwhile.

---
In reply to: [69997735](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/117#discussion_r69997735) [](ancestors = 69997735)

PR #118, opened 2016-07-08
ApiHub to support poison queues 

safihamid:
ApiHub to support poison queues when there are multiple unsuccessful function execution attempts.


src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -20,13 +21,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.ApiHub
     [Singleton(Mode = SingletonMode.Listener)]
     internal class ApiHubListener : IListener
     {
+        private const int MaxFunctionExecutionRetryCount = 5;
         private const string HostContainerName = "azure-webjobs-hosts";
+        private const string PoisonQueueName = "azure-webjobs-apihub-poison";
```
mathewc: Thinking you should follow the naming scheme for blob trigger https://azure.microsoft.com/en-us/documentation/articles/websites-dotnet-webjobs-sdk-storage-blobs-how-to/

E.g. **webjobs-apihubtrigger-poison**

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -20,13 +21,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.ApiHub
     [Singleton(Mode = SingletonMode.Listener)]
     internal class ApiHubListener : IListener
     {
+        private const int MaxFunctionExecutionRetryCount = 5;
```
mathewc: I think you should make this a configuration knob on your api hub config object. For queues/blobs this is configurable at the host level

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -105,24 +114,56 @@ public void Dispose()
...
-                    pollStatus = uri.AbsoluteUri;
+                    status = new ApiHubStatus();
                 }
-                else
+
+                if (status.RetryCount < MaxFunctionExecutionRetryCount)
                 {
-                    pollStatus = obj.ToString();
+                    status.RetryCount++;
+                    _trace.Error(string.Format("Function {0} failed to successfully proces file {1}. Number of retries: {2}. ", _functionName, apiHubFile.Path, status.RetryCount));
```
mathewc: spelling errors

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -105,24 +114,56 @@ public void Dispose()
...
+                {
+                    // The maximum retries for the function execution has reached. The file info will be added to the poison queue and the next poll status will be updated to skip the file.
+                    status.PollUrl = pollStatus;
+                    status.RetryCount = 0;
+                    _trace.Error(string.Format("Function {0} failed to successfully proces file {1} after max allowed retries of {2}. The file info will be moved to queue {3}.", _functionName, apiHubFile.Path, MaxFunctionExecutionRetryCount, PoisonQueueName));
 
-                // If function successfully completes then the next poll Uri will be logged in an Azure Blob.
-                await SetNextPollStatusAsync(pollStatus);
+                    await MoveToPoisonQueueAsync(apiHubFile);
+                    await SetNextPollStatusAsync(status);
```
mathewc: If the status update fails after you've moved to the poison queue, you could get duplicates in the queue right? I realize these two operations can't be done as an atomic transaction.

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -134,7 +175,34 @@ public void Dispose()
             }
         }
 
-        private async Task<string> GetLastPollStatusAsync()
+        private async Task MoveToPoisonQueueAsync(ApiHubFile apiHubFile)
+        {
+            CloudQueueClient queueClient = CloudStorageAccount.Parse(this._config.StorageConnectionString).CreateCloudQueueClient();
```
mathewc: You should do this once on construction, or otherwise find a way to cache the queue client

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -134,7 +175,34 @@ public void Dispose()
...
+
+            var queue = queueClient.GetQueueReference(PoisonQueueName);
+
+            await queue.CreateIfNotExistsAsync();
+
+            var fileStatus = new ApiHubFileInfo
+            {
+                FilePath = apiHubFile.Path,
+                FunctionName = this._functionName,
+                Connection = this._connectionStringSetting
```
mathewc: I think the connection string is resolved at this point? You want to avoid writing the actual connection string to the queue

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
```diff
@@ -134,7 +175,34 @@ public void Dispose()
...
-        private async Task<string> GetLastPollStatusAsync()
+        private async Task MoveToPoisonQueueAsync(ApiHubFile apiHubFile)
+        {
+            CloudQueueClient queueClient = CloudStorageAccount.Parse(this._config.StorageConnectionString).CreateCloudQueueClient();
+
+            var queue = queueClient.GetQueueReference(PoisonQueueName);
+
+            await queue.CreateIfNotExistsAsync();
+
+            var fileStatus = new ApiHubFileInfo
```
mathewc: Has this stuff been released yet? This would be a breaking change if so.

mathewc:
Need to add tests

src/WebJobs.Extensions.ApiHub/Listener/ApiHubListener.cs
safihamid: It's not the actual connection string. It's the connectionStringSetting aka the key.

mathewc: Can you check its runtime value? Most of the binding attributes now use the new [AutoResolve] attribute which causes these values to be resolved automatically. Perhaps you're not using that and resolving manually. @brettsam might have comments as well.

safihamid: Not sure, why is this needed? Am I missing something? This is just to let the user know which trigger failed. We are giving them the file path, function name and the connection string name which is something like DROPBOX_dropbox for them to know which connector was the file coming from which failed.

brettsam: It's only resolved if he's using [AutoResolve(AllowTokens=false)] on this property. I think ApiHub handles it differently and resolves these on their own.

---
In reply to: [70373036](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/118#discussion_r70373036) [](ancestors = 70373036)

src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
```diff
@@ -65,6 +68,25 @@ public TraceWriter Logger
             }
         }
 
+        /// <summary>
+        /// Gets or sets the number of times to try processing a file before moving it to the poison queue (where
+        /// possible).
+        /// </summary>
+        public int MaxFunctionExecutionRetryCount
```
mathewc: nit: recommend simplifying this to MaxRetryCount

mathewc:
:ship:

PR #120, opened 2016-07-12
Moving NotificationHub setting properties to use AutoResolve

pragnagopa:


src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
```diff
@@ -39,12 +39,14 @@ public sealed class NotificationHubAttribute : Attribute
...
         /// string, if different than the one specified in the <see cref="NotificationHubsConfiguration"/>.
         /// </summary>
+        [AutoResolve(AllowTokens = false)]
         public string ConnectionStringSetting { get; set; }
 
         /// <summary>
         /// Optional. The Notification Hub Name to use, if different than the one specified in the
         /// <see cref="NotificationHubsConfiguration"/>.
         /// </summary>
+        [AutoResolve(AllowTokens = false)]
```
mathewc: You don't want AllowTokens false here. @brettsam  please confirm

brettsam: Depends -- is this filled in by the picker? If so, it should be AllowTokens = false. This means it will always be an app setting and no direct strings are allowed.
@pragnagopa -- Is that the intent?


---
In reply to: [70489560](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/120#discussion_r70489560) [](ancestors = 70489560)

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -69,58 +65,48 @@ private IAsyncCollector<Notification> BuildFromAttribute(NotificationHubAttribut
...
             }
-            return configHubName;
-        }
-
-        internal static string GetSettingFromConfigOrEnvironment(string key)
-        {
-            string value = null;
 
-            if (string.IsNullOrEmpty(value))
+            // Second, try the config's ConnectionString
```
brettsam: >ConnectionString [](start = 40, length = 16)

nit -- this is actually the hubname

src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
mathewc: HubName is kind of like QueueName, and for queues we support strings containing literal portions as well as %%

pragnagopa: Currently, UI does not support picking NotificationHub. But eventually we need to improve this. NotificationHubName is different from QueueName as it needs to be pre-configured before using it.

brettsam: So using AllowTokens=false seems wrong -- that would require it to be a lookup value. If you want to support both literal strings as well as replacements, we should just have this be [AutoResolve] and when the picker works is done, it needs to write %appsetting%.

---
In reply to: [70492852](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/120#discussion_r70492852) [](ancestors = 70492852)

brettsam: Pragna and I were IMing -- she actually doesn't want [AutoResolve] at all on this property. She'll pull the default fallback from app settings using the name resolver, but everything else will use a literal.

---
In reply to: [70493194](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/120#discussion_r70493194) [](ancestors = 70493194,70492852)

mathewc:
:ship:

brettsam:
:shipit:

mathewc:
FYI - I'm merging this now

mathewc:
Merged

PR #121, opened 2016-07-13
Fixing Config setting defaulting for DocDB/NotificationHub/MobileApps

mathewc:
When trying to pull in all the recent config setting changes into the script repo, lots of tests broke. The reason was because the were expecting (as was the case before) that when a default config object is constructed, it's various values are set to their environmental defaults. E.g., existing test code in script:

```csharp
// Get the URI by creating a config.
var config = new MobileAppsConfiguration();
var client = new MobileServiceClient(config.MobileAppUri);
```

I think we want to retain the previous behavior, where the default is populated (and can be overridden by setting). With these changes, all the tests pass in both extensions and script repos.


src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -19,14 +19,17 @@ public class DocumentDBConfiguration : IExtensionConfigProvider
...
 
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
         public DocumentDBConfiguration()
         {
             DocumentDBServiceFactory = new DefaultDocumentDBServiceFactory();
+
+            // default the connection string
+            var nameResolver = new DefaultNameResolver();
```
mathewc: This guy should probably have a static instance method so people can use DefaultNameResolver.Instance, rather than newing this up.

brettsam: I'd rather not do this -- it breaks the DI/mockability that my change purposefully introduced. 

It's nice to be able to register your own NameResolver and know that we'd use it everywhere we did resolution. If we hardcode the DefaultNameResolver here, the story is "we'll use your registered INameResolver for all resolution except for when we fill in default values, which you cannot change."

I'd rather fix the tests in Script (which I can take care of). They were taking advantage of an implementation detail that has since changed. The behavior of real functions shouldn't be affected.

---
In reply to: [70554004](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/121#discussion_r70554004) [](ancestors = 70554004)

mathewc: Please send a PR for the test fixes

PR #122, opened 2016-07-15
Added support to bind NotificationHubClient

### LINKED ISSUE #86: [NotificationHub]Add support to bind NotificationHubClient

pragnagopa:
Fixes #86

### DIFF
--- a/src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
+++ b/src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Globalization;
 using Microsoft.Azure.NotificationHubs;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Bindings;
@@ -86,7 +87,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
 
         internal INotificationHubClientService GetService(string connectionString, string hubName)
         {
-            return ClientCache.GetOrAdd(new Tuple<string, string>(connectionString, hubName), (c) => NotificationHubClientServiceFactory.CreateService(c.Item1, c.Item2));
+            return ClientCache.GetOrAdd(new Tuple<string, string>(connectionString, hubName.ToLowerInvariant()), (c) => NotificationHubClientServiceFactory.CreateService(c.Item1, c.Item2));
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


pragnagopa:
@brettsam - Please review

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -47,19 +59,36 @@ public void Initialize(ExtensionConfigContext context)
...
+            string resolvedConnectionString = ResolveConnectionString(attribute.ConnectionStringSetting);
+            string resolvedHubName = ResolveHubName(attribute.HubName);
+            INotificationHubClientService service = GetService(resolvedConnectionString, resolvedHubName);
+
+            return service.GetNotificationHubClient();
+        }
+
+        internal INotificationHubClientService GetService(string connectionString, string hubName)
+        {
+            return ClientCache.GetOrAdd(new Tuple<string, string>(connectionString, hubName), (c) => NotificationHubClientServiceFactory.CreateService(c.Item1, c.Item2));
```
brettsam: >new Tuple<string, string>(connectionString, hubName) [](start = 40, length = 52)

Should the hubname comparison be case sensitive or case insensitive? Not sure how it works. But you should add a test using a hubname of "TEST" and "test" and see if it behaves the way you want.

test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubEndToEndTests.cs
```diff
@@ -0,0 +1,238 @@
...
+                .Returns(serviceMock.Object);
+
+            var testTrace = new TestTraceWriter(TraceLevel.Warning);
+
+            //Act
+            RunTest("Outputs", factoryMock.Object, testTrace);
+
+            // Assert
+            factoryMock.Verify(f => f.CreateService(ConfigConnStr, DefaultHubName), Times.Once());
+            serviceMock.Verify(m => m.SendNotificationAsync(It.IsAny<Notification>(), It.IsAny<string>()), Times.Exactly(9));
```
brettsam: >Times.Exactly(9) [](start = 107, length = 16)

For the sake of completion, I'd have your Collectors each add at least one too. Then just bump this up.

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
pragnagopa: Good point. Verified hubname is case insensitive. Will add tests and update key to be case insensitive as well.

pragnagopa:
@brettsam - SO?

brettsam:
:shipit:

PR #124, opened 2016-07-16
[WIP] Initial Twilio SMS binding implementation

fabiocav:
Opening this for initial feedback


src/WebJobs.Extensions.Twilio/Bindings/TwilioScriptBindingProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+            : base(config, hostMetadata, traceWriter)
+        {
+            _twilioAssemblyName = typeof(SMSMessage).Assembly.GetName().Name;
+        }
+
+        internal TwilioSMSConfiguration CreateConfiguration(JObject metadata)
+        {
+            var twilioConfig = new TwilioSMSConfiguration();
+
+            JObject configSection = (JObject)metadata.GetValue("twilioSMS", StringComparison.OrdinalIgnoreCase);
```
mathewc: I think it would be cleaner if the metadata section property was just "twilio" w/o SMS on it?

src/WebJobs.Extensions.Twilio/Bindings/TwilioSmsMessageAsyncCollector.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                throw new InvalidOperationException("A 'From' number must be specified for the message.");
+            }
+
+            if (string.IsNullOrEmpty(item.From))
+            {
+                throw new InvalidOperationException("A 'Body' must be specified for the message.");
+            }
+
+            var response = await Task.Run(() => _client.SendMessage(item.From, item.To, item.Body)).ConfigureAwait(false);
```
mathewc: Why Task.Run?

mathewc: Also, AsyncCollectors usually collect all the items and only send them all in the Flush method (see SendGrid and other bindings).

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.Azure.WebJobs.Host.Config;
+using Newtonsoft.Json.Linq;
+using Twilio;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Extension methods for Twilio integration
+    /// </summary>
+    public static class SampleJobHostConfigurationExtensions
```
mathewc: Copy paste error :)

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+                IBindingProvider outputProvider = factory.BindToAsyncCollector<TwilioSMSAttribute, SMSMessage>((attr) =>
+                {
+                    return new TwilioSMSMessageAsyncCollector(_twilioConfig, attr, _twilioClient);
+                });
+
+                IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
+                extensions.RegisterBindingRules<TwilioSMSAttribute>(outputProvider);
+            }
+
+            private SMSMessage CreateSmsMessage(JObject messageJson)
```
mathewc: nit: SMS

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                var message = new SMSMessage();
+                message.From = GetPropertyValue(messageJson, "from");
+                message.To = GetPropertyValue(messageJson, "to");
+                message.Body = GetPropertyValue(messageJson, "body");
```
mathewc: Isn't there a MediaUrl property as well for including pictures/video? If so, might be good to include

src/WebJobs.Extensions.Twilio/WebJobs.Extensions.Twilio.csproj
```diff
@@ -0,0 +1,121 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{65563C23-75B9-4F48-93E0-9A9FA2141CF9}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.Azure.WebJobs.Extensions</RootNamespace>
```
mathewc: Add the code analysis sections (see other projects)

mathewc:
As part of finishing this up, you're going to add examples to the samples project as well, right? Also update the wiki to include this awesome new binding :)

src/WebJobs.Extensions.Twilio/Bindings/TwilioSmsMessageAsyncCollector.cs
fabiocav: > Why Task.Run?

Kicking off a Task for now since the current Twilio SDK does not expose async methods. I didn't want to block here as clients may have issues with that. Their new SDK (in alpha, but I might actually use here) does expose async methods, so this is temporary until we move to that. 

> Also, AsyncCollectors usually collect all the items and only send them all in the Flush method (see SendGrid and other bindings). Add is supposed to just queue them up.

This was initially implemented as you're describing, but I switched it to follow a model similar to the Mobile Apps and DocumentDB implementations and send the messages on add. I'll change that back.

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
fabiocav: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.SendGrid/Config/SendGridJobHostConfigurationExtensions.cs#L18

A double one, too :)

src/WebJobs.Extensions.Twilio/Bindings/TwilioScriptBindingProvider.cs
fabiocav: Yes, that was the thought. I do want to support other Twilio services (e.g. voice, tasks, etc.), that's why I qualified this.

fabiocav:
Yes! I'll have some samples here. Just wanted to get this out for an early review :)

src/WebJobs.Extensions.Twilio/WebJobs.Extensions.Twilio.csproj
fabiocav: Done.

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
fabiocav: We can pass one or more media URLs into the API call, but unfortunately, their `SMSMessage` type does not expose that as a property. I considered having our own model defined (our own `SmsMessage`) to have just the relevant properties, including media URLs.

fabiocav:
@brettsam, let me know if you have any feedback on this.

fabiocav:
@mathewc , @brettsam this is pretty low priority, but when you have a chance, let me know if you have any additional feedback on this.

mathewc:
I don't have any further comments on this. If everything is all wrapped up from your point of view, it works and you have working samples integrated, I say get it in so we can start getting feedback on it :)

README.md
```diff
@@ -306,6 +306,39 @@ public static void SendTemplateNotification(
...
+        From = "{StorePhoneNumber}",
+        Body = "{CustomerName}, we've received your order ({OrderId}) and have begun processing it!")]
+    out SMSMessage message)
+{
+    // You can set additional message properties here
+    message = new SMSMessage();
+}
+```
+
+Here's another example showing how you can easily send yourself notification mails to your own admin address when your jobs complete. In this case, the default To/From addresses come from the global SendGridConfiguration object specified on startup, so don't need to be specified.
```
mathewc: nit: copy/paste error (SendGridConfiguration)

src/ExtensionsSample/Samples/TwilioSamples.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        /// <summary>
+        /// Demonstrates the IAsyncCollector binding. This works with JObject
+        /// or SMSMessage. Using IAsyncCollector is also a way to conditionally
+        /// send messages from a function.
+        /// </summary>
+        [Disable]
+        public static void ProcessOrder_JObjectAsyncCollector(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [TwilioSms] IAsyncCollector<JObject> messages)
```
mathewc: Shouldn't SMS be capitalized in the attribute Type name? SMS is capitalized in their SDK types

src/WebJobs.Extensions.Twilio/Bindings/TwilioScriptBindingProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <inheritdoc/>
+        public TwilioScriptBindingProvider(JobHostConfiguration config, JObject hostMetadata, TraceWriter traceWriter) 
+            : base(config, hostMetadata, traceWriter)
+        {
+            _twilioAssemblyName = typeof(SMSMessage).Assembly.GetName().Name;
+        }
+
+        internal static TwilioSmsConfiguration CreateConfiguration(JObject metadata)
+        {
+            var twilioConfig = new TwilioSmsConfiguration();
```
mathewc: Here as well - if we decide to capitalize SMS in Type names like their SDK does, recheck all the new types like this config type

christopheranderson:
:shipit:

mathewc:
:ship:

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Twilio
+{
+    /// <summary>
+    /// Defines the configuration options for the Twilio binding.
+    /// </summary>
+    public class TwilioSmsConfiguration
```
brettsam: >TwilioSmsConfiguration [](start = 17, length = 22)

Some of our more recent changes have made the Configuration implement IExtensionConfigProvider -- it simplifies the classes a bit. (see DocDB for an example).

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        internal const string AzureWebJobsTwilioAccountSidKeyName = "AzureWebJobsTwilioAccountSid";
+        internal const string AzureWebJobsTwilioAccountAuthTokenName = "AzureWebJobsTwilioAuthToken";
+
+        public TwilioSmsConfiguration()
+        {
+            var nameResolver = new DefaultNameResolver();
+            AccountSid = nameResolver.Resolve(AzureWebJobsTwilioAccountSidKeyName);
```
brettsam: Don't create your own NameResolver -- that prevents anyone from overriding it in the JobHostConfig. 

Building on my comment above -- if this class implements IExtensionConfigProvider, you can get the NameResolver in the Initialize() method and set things up. Again, see: [https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs)

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    ThrowMissingSettingException("AccountSID", TwilioSmsConfiguration.AzureWebJobsTwilioAccountSidKeyName, "AccountSID");
+                }
+
+                if (string.IsNullOrEmpty(_twilioConfig.AuthToken))
+                {
+                    ThrowMissingSettingException("AuthToken", TwilioSmsConfiguration.AzureWebJobsTwilioAccountAuthTokenName, "AuthToken");
+                }
+
+                _twilioClient = new TwilioRestClient(_twilioConfig.AccountSid, _twilioConfig.AuthToken);
```
brettsam: >_twilioClient = new TwilioRestClient(_twilioConfig.AccountSid, _twilioConfig.AuthToken); [](start = 16, length = 88)

This won't work properly in Script -- you'll want each script to be able to define its own AccountSid and AuthToken, but this will use the config's no matter what.

In DocDB/Mobile/NH, we maintain a 'ClientCache' dictionary of all the clients based on some lookup value (probably AccountSid here?). Then when it comes time to get one, you pull it out and use it.

src/WebJobs.Extensions.Twilio/TwilioSMSAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Attribute used to bind a parameter to a Twilio SMSMessage that will automatically be
+    /// sent when the function completes.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class TwilioSmsAttribute : Attribute
```
brettsam: >TwilioSmsAttribute [](start = 24, length = 18)

You're going to need to support AccountSid and AuthToken here as well so you can set them on each Script. 

AuthToken will definitely be one of the 'secret' properties that you'll want to require to be an app setting (resolved by the NameResolver). I'm not sure about AccountSid.

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+
```
brettsam: Also, you may want to create the TwilioRestClient via a Factory so you can replace it with a mocked version for testing. I'm not sure how complete the tests will be for this extension, but it's a thought.

src/ExtensionsSample/Samples/TwilioSamples.cs
fabiocav: Addressed offline.

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
fabiocav: Was following the pattern used by the sample and Sendgrid. I'll update. Do we need to go through some of the other extensions and do the same?

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
fabiocav: The `IExtensionConfigProvider` is implemented here:
https://github.com/Azure/azure-webjobs-sdk-extensions/pull/124/files#diff-39fe02541097467043407a687aaebdf3R40
This follows the pattern used by the Sendgrid extension.

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
fabiocav: The name resolver is used in the `IExtensionConfigProvider` implementation here:
https://github.com/Azure/azure-webjobs-sdk-extensions/pull/124/files#diff-39fe02541097467043407a687aaebdf3R72 
Again, following the pattern established by Sendgrig, but happy to move this type elsewhere if improves clarity.

src/WebJobs.Extensions.Twilio/TwilioSMSAttribute.cs
fabiocav: Will take a look at this.

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
fabiocav: Have some basic tests in place. Intend to add more, may postpone until then, but definitely a good idea.

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
fabiocav: Makes sense. We definitely want to support this per function (which I'm assuming is what you meant), so I'll make this change.

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
brettsam: Sendgrid probably isn't the best one to follow. The biggest issue is that it doesn't allow each function to set its own Account details and instead relies on the environment variables exclusively.

Once you have it set up to check the Attribute properties first, then look at the Config settings, then fallback to environment, it's a bit easier to do it all in one class. That allows you access to the registered NameResolver as well. 

---
In reply to: [75050079](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/124#discussion_r75050079) [](ancestors = 75050079)

src/WebJobs.Extensions.Twilio/Config/TwilioSMSConfiguration.cs
brettsam: That nameresolver is used for [AutoResolve] -- which is correct. But here you're using it to set the default values. If someone were to plug in a TestNameResolver, it wouldn't get used here. 

The 'Attribute wins' approach (that SendGrid does not use) made this more complicated. I'd check out something like NH (which is probably the simplest). 

---
In reply to: [75050475](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/124#discussion_r75050475) [](ancestors = 75050475)

fabiocav: Makes sense. I'll check NH and update this.

src/WebJobs.Extensions.Twilio/Config/TwilioJobHostConfigurationExtensions.cs
mathewc: I think generally people are going to use a single SendGrid / Twilio account. I see how it is more flexible to allow this to vary per function, but for SendGrid I didn't see that being a very important scenario, and kept things simple.

fabiocav: I could certainly keep things as is, but if the effort is not significant, I think it's worth spending a bit more time to get this working as Brett is describing, since there are some scenarios (e.g. test tokens) that would benefit from it.

src/WebJobs.Extensions.Twilio/Config/TwilioSmsConfiguration.cs
```diff
@@ -0,0 +1,122 @@
...
+                ThrowMissingSettingException("AuthToken", AzureWebJobsTwilioAccountSidKeyName, "AuthToken");
+            }
+        }
+
+        private TwilioSmsContext CreateContext(TwilioSmsAttribute attribute)
+        {
+            string accountSid = Utilities.GetValueOrDefault(attribute.AccountSidSetting, AccountSid, _defaultAccountSid);
+            string authToken = Utilities.GetValueOrDefault(attribute.AuthTokenSetting, AuthToken, _defaultAuthToken);
+
+            TwilioRestClient client = _twilioClientCache.GetOrAdd(new Tuple<string, string>(accountSid, authToken), t => new TwilioRestClient(t.Item1, t.Item2));
```
fabiocav: Will be introducing a factory here as I work on additional tests.

fabiocav:
@brettsam some things still in progress but I made the changes to support the scenarios we've discussed above. Now, AccountSID and AuthToken are supported from configuration and attribute. To, From and Subject from the attribute or the message itself.

fabiocav:
Just wanted to ping you guys again to see if you have any other concerns that haven't been addressed. If you're comfortable, we can add what we have here (as it wouldn't impact any other extensions) and iterate based on feedback.

brettsam:
:shipit:

Issue #132, opened 2016-07-28
Restarting/redeploy WebJob can cause DailySchedule to run twice


Thnslj:
Using  `<package id="Microsoft.Azure.WebJobs.Extensions" version="2.0.0-beta1-10332" targetFramework="net461" />`

I have a DailySchedule running each day at 1 AM.
```
internal class BackupScheduler : DailySchedule
{
   public BackupScheduler() : base("01:00:00")
      {
      }
}
```
Sometimes the System.Timers.Timer will trigger the job before 1AM
![Azure WebJobs Dashboard](http://i.imgur.com/pnJUiDK.png)

The status file written to  `azure-webjobs-hosts\timers\{hostid\{functionId}\status}` Looks like this:
`{"Last":"2016-07-28T00:59:59.8690904+00:00","Next":"2016-07-28T01:00:00"}`

The "Next" value is wrongly calculated to only a few milliseconds after the "Last" run. This is not a problem if the process keeps running, because (as I understand) the status file is only read on startup.

But if the WebJobs is redeployed or simply restarted. The `CheckPastDueAsync` in `ScheduleMonitor` will find that the timer IS past due and should be started.

I have fixed it locally with a hack; But this should be fixed in a proper way.
```
    internal class BackupScheduler : DailySchedule
    {
        public BackupScheduler() : base("01:00:00")
        {
        }

        public override DateTime GetNextOccurrence(DateTime currentTime)
        {
            TimeSpan threshold = TimeSpan.FromSeconds(10);
            DateTime nextOccurrence = base.GetNextOccurrence(currentTime);

            if (currentTime + threshold <= nextOccurrence)
            {
                return nextOccurrence;
            }
            return nextOccurrence.AddDays(1);
        }
    }

```

davidebbo:
@brettsam this looks a lot like the issue you were investigating, no?

brettsam:
Yes! Thank you for the details @Thnslj, this helps enormously. I think I know what's going on now -- it's possible for a timer to fire just before its 'officially' scheduled time (as we've written it in the blob). So if you've said you want it to run every day at midnight and we fire the timer just before midnight on the 1st, we'll calculate that the next run should still be at midnight on the 1st -- because that's the next midnight that we see. It doesn't affect anything if the site remains running but once it restarts, it consults the status blob and sees that it's past due so it runs again.

I'll look into a proper fix.

shyamal890:
[...]

Issue #40, opened 2016-02-19
Repeated executions


paolorr:
I'm using TimerTrigger to run a job daily at 18:00 (6pm UTC -2:00) but I noticed the job sometimes runs twice as shown in the log

In this case it runs twice due to website restart, which I don't know why, because it's set Always On
```
[02/15/2016 19:59:59 > f70569: INFO] Executing: 'Functions.SendPublishedInspections' - Reason: 'Timer fired at 2016-02-15T17:59:58.6655227-02:00'
[02/15/2016 20:00:00 > f70569: INFO] Executed: 'Functions.SendPublishedInspections' (Succeeded)
[02/15/2016 20:42:53 > 9dad03: SYS INFO] Status changed to Starting
...
[02/15/2016 20:43:08 > 9dad03: INFO] Job host started
[02/15/2016 20:45:45 > f70569: SYS INFO] WebJob is stopping due to website shutting down
[02/15/2016 20:45:45 > f70569: SYS INFO] Status changed to Stopping
[02/15/2016 20:45:46 > f70569: INFO] Job host stopped
[02/15/2016 20:45:47 > f70569: SYS INFO] Status changed to Success
[02/15/2016 20:45:47 > f70569: SYS INFO] Status changed to Stopped
[02/15/2016 20:46:09 > 9dad03: INFO] Executing: 'Functions.SendPublishedInspections' - Reason: 'Timer fired at 2016-02-15T18:46:09.0790525-02:00'
[02/15/2016 20:46:10 > 9dad03: INFO] Executed: 'Functions.SendPublishedInspections' (Succeeded)
```

In this case it runs twice due to webjob restart, which I also don't know why, because the website hasn't restarted. Notice that the first execution was ok, at 17:59, but the second execution was at 22:45, 4 hours after the scheduled time
```
[02/18/2016 19:59:59 > 2627e8: INFO] Executing: 'Functions.SendPublishedInspections' - Reason: 'Timer fired at 2016-02-18T17:59:59.7431098-02:00'
[02/18/2016 20:00:00 > 2627e8: INFO] Executed: 'Functions.SendPublishedInspections' (Succeeded)
[02/19/2016 00:44:44 > 2627e8: SYS INFO] Status changed to Starting
..
[02/19/2016 00:44:51 > 2627e8: INFO] Job host started
[...]

PR #133, opened 2016-07-28
fixing duplicate timer trigger issue

brettsam:
Will address #40 and #132.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -29,7 +29,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
         private string _timerName;
         private bool _disposed;
         private TimeSpan _remainingInterval;
-        private ScheduleStatus _scheduleStatus;
 
         public TimerListener(TimerTriggerAttribute attribute, TimerSchedule schedule, string timerName, TimersConfiguration config, ITriggeredFunctionExecutor executor, TraceWriter trace)
         {
@@ -43,7 +42,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
             ScheduleMonitor = _attribute.UseMonitor ? _config.ScheduleMonitor : null;
         }
 
-        internal static TimeSpan MaxTimerInterval 
+        internal static TimeSpan MaxTimerInterval
         {
             get
             {
@@ -59,6 +58,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 return _timer;
             }
         }
+
+        internal ScheduleStatus ScheduleStatus { get; set; }
 
         internal ScheduleMonitor ScheduleMonitor { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -176,20 +175,27 @@ private void OnTimer(object sender, ElapsedEventArgs e)
...
+        /// <param name="runOnStartup">True if the invocation is because the timer is configured to run on startup.</param>
+        internal async Task InvokeJobFunction(DateTime invocationTime, bool isPastDue = false, bool runOnStartup = false)
         {
             CancellationToken token = _cancellationTokenSource.Token;
-            TimerInfo timerInfo = new TimerInfo(_schedule, _scheduleStatus, isPastDue);
+            ScheduleStatus timerInfoStatus = null;
+            if (ScheduleMonitor != null)
+            {
+                timerInfoStatus = ScheduleStatus;
+            }
```
brettsam: Previously, we'd only maintain the ScheduleStatus if we were storing it in a ScheduleMonitor. But it's also handy even if it's just kept in-memory (it's how I prevent duplicate 'Next' values now).

The API for TimerInfo says that you will not get a ScheduleStatus if your interval is short -- but my change meant you'd get a ScheduleStatus no matter what, so I do this to keep it backwards-compatible.

Is there any reason that I need to do this? I can remove this 'back-compat' piece here if there's no objections.

brettsam:
Hey @fabiocav -- could you take a look at this for me?

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -176,20 +175,27 @@ private void OnTimer(object sender, ElapsedEventArgs e)
...
 
-            DateTime now = DateTime.Now;
-            await InvokeJobFunction(now, false);
-
             // restart the timer with the next schedule occurrence
-            now = DateTime.Now;
-            DateTime nextOccurrence = _schedule.GetNextOccurrence(now);
-            TimeSpan nextInterval = nextOccurrence - now;
+            await InvokeJobFunction(DateTime.Now, false);
+            TimeSpan nextInterval = ScheduleStatus.Next - DateTime.Now;
```
fabiocav: Is the change to use `DateTime.Now` (getting the time when this is calculated) instead of the cached value used when invoking the function intentional? This will behave differently as those could now be significantly different.

brettsam: That actually didn't change -- if you look at the previous iteration, we would update now with a new DateTime.Now after invoking. It confused me to cache that value but not actually re-use it so I switched it to use DateTime.Now directly.

In order for this interval to be correct, we need to calculate it after the function has run -- as close to the timer starting as possible. Example -- if you have an hour timer invoked at 12:00 and the invocation takes 1 minute, we want to calculate the interval for the next timer using 12:01 rather than at 12:00. Otherwise, you'd start a 1 hour timer at 12:01. But we really want it to be a 59-minute timer.

---
In reply to: [73020609](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/133#discussion_r73020609) [](ancestors = 73020609)

fabiocav: I see, yes, you're right. I misread the code above what you've changed and missed the line where `now` was being updated again. I agree that this was confusing, so a good change 👍

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -79,41 +80,39 @@ internal System.Timers.Timer Timer
...
 
             // log the next several occurrences to console for visibility
             string nextOccurrences = TimerInfo.FormatNextOccurrences(_schedule, 5);
             _trace.Verbose(nextOccurrences);
 
-            // start the timer
-            now = DateTime.Now;
-            DateTime nextOccurrence = _schedule.GetNextOccurrence(now);
-            TimeSpan nextInterval = nextOccurrence - now;
+            // start the timer
```
mathewc: Might as well add some detailed comments here, that we need to start the timer AFTER the function invocation (the comments you made below)

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
mathewc: If we're not doing schedule monitoring, then values like ScheduleStatus.Last will not be correct in your in memory case. So users coding against it in their function won't get valid values. So I'd say keep the back compat.

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -209,14 +215,25 @@ private void OnTimer(object sender, ElapsedEventArgs e)
                 // schedule. Errors will be logged to Dashboard already.
             }
 
+            // If the trigger fired before it was officially scheduled (likely under 1 second due to clock skew),
+            // adjust the invocation time forward for the purposes of calculating the next occurrence.
+            // Without this, it's possible to set the 'Next' value to the same time twice in a row, 
+            // which results in duplicate triggers if the site restarts.
+            DateTime adjustedInvocationTime = invocationTime;
+            if (!isPastDue && !runOnStartup && ScheduleStatus?.Next > invocationTime)
```
mathewc: This line uses C# 6 which will currently fail compilation on our TeamCity

mathewc: Why is `!runOnStartup` part of the clause? That would mean that any RunOnStartup functions won't get your fix.

mathewc:
Couple minor comments, then :ship:

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
brettsam: I was planning on having this be the first package released from the AppVeyor build, so it should be okay.

---
In reply to: [73402013](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/133#discussion_r73402013) [](ancestors = 73402013)

PR #140, opened 2016-08-04
fixing a timer bug that could cause negative intervals

brettsam:


src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -241,6 +235,21 @@ private void OnTimer(object sender, ElapsedEventArgs e)
...
+        {
+            // We need to calculate the next interval based on the current
+            // time as we don't know how long the previous function invocation took.
+            // Example: if you have an hourly timer invoked at 12:00 and the invocation takes 1 minute,
+            // we want to calculate the interval for the next timer using 12:01 rather than at 12:00.
+            // Otherwise, you'd start a 1-hour timer at 12:01 when we really want it to be a 59-minute timer.
+
+            // Recalculate the next occurrence as well. For short (1-second) intervals, it's possible that we've 
+            // missed the next invocation and do not want to end up with a negative interval.
+            ScheduleStatus.Next = _schedule.GetNextOccurrence(now);
```
mathewc: InvokeJobFunction is also updating ScheduleStatus.Next and persisting it to storage. So it's confusing to me that we'd recompute it again here. Is the situation you're talking about a scenario where the interval is so short that the time it takes us to persist it to storage causes us to miss the next expected? 

Perhaps what we should do is have the existing StartTimer(Interval) function interpret negative intervals as zero. That would handle this corner case, causing the function to run immediately.

brettsam: The Script tests hit this with RunOnStartup = true and without persisting (since short intervals don't persist). In that case, it may be that the verbose logging we do between that initial startup invocation and this calculation was enough to throw it into negative territory.

But you're right that persisting it to storage could cause this as well. If you have RunOnStartup = true (or are past due) and *very* close to the next interval, that persistence step will throw it over.

I actually first did the negative -> zero calculation and tested it, but decided to go back to this since this logic had been around a while. But I do think adjusting back to zero works as well, and maybe it's easier to follow.

Now that I think about it, it's probably 'more correct' as well. This logic I currently have (and that we previously had) could cause you to miss an interval. Setting it to zero would start your intervals immediately, but maybe *very* slightly behind schedule.


---
In reply to: [73600161](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/140#discussion_r73600161) [](ancestors = 73600161)

mathewc:
:ship:

PR #144, opened 2016-08-20
Exporting DocumentDB assembly

fabiocav:
Updating the DocumentDB extension to export the SDK assembly so users are no longer forced to add a Nuget reference.


mathewc:
:ship:

Issue #32, opened 2016-01-20
Timer Trigger: Is the time(s) GMT or local to the Azure VM region?


PureKrome:
If i have a Timer Trigger webjob and say: _run this at 2am, 6am and 3pm_ .. what times will this _really_ run at?

~~~eg. is that 2am @ GMT? or the system time of the azure VM the web job exists on (which I'm assuming is the local time of the region ...) ?~~~

AH! the commit [Use Local time (not UTC) in TimerTrigger scheduling](https://github.com/Azure/azure-webjobs-sdk-extensions/commit/f6c31de78492a8663f88d0873931bb48412148c7) says exactly that .. it's now **local time** :smile: (and all the `UtcNow`'s are replaced with `Now`) 

... so what is _local time_ (just to confirm). I know that, that code uses the machine's system time - but what time is this set to -- the time of the region or something?

mathewc:
I believe the local time in all Azure stamps is set to be the same as UTC, but don't quote me on that.

Regardless, the good thing about using Local time is that you can now control the actual timezone by changing the timezone of your WebApp ([details here](http://blogs.msdn.com/b/tomholl/archive/2015/04/07/changing-the-server-time-zone-on-azure-web-apps.aspx)).

PureKrome:
Hi @mathewc - sorry to cast 'ressurect dead' on this thread, but I've been watching my webjob with more scrutiny over the last few days / week (as my bugs are ironed out) and noticed that changing the timezone of the webjob is not working for me.

Or (more accurately) the CRON JOB is not working as expected .

These are my two methods I have in `functions.cs` :

- `public static async Task Method2Async([TimerTrigger("0 0 2 * * *", UseMonitor = false)] TimerInfo timer)`
- `public static async Task Method2Async([TimerTrigger("0 30 2 * * *", UseMonitor = false)] TimerInfo timer)

First one is for 2am, 2nd is 2:30am -> and by that, I'm hoping for 2am/2:30am Australia Eastern Standard Time (i.e. AESTD).

Looking at my logging, they are getting fired off at 1pm and 1:30pm AESTD. Which is 2am UTC time.

Following the instructions from your previous post, i did the following:

- Added a setting to `appsettings`
- Added an entry to the website appsettings

![image](https://cloud.githubusercontent.com/assets/899878/12837814/4cf6125a-cc1a-11e5-97f6-a94c67999be5.png)

[...]

PR #149, opened 2016-09-14
prevent TimerListener from firing after it's been stopped

brettsam:
Should address https://github.com/Azure/azure-webjobs-sdk-script/issues/558.

I've also included a commit to move our nugets forward.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -75,6 +75,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
             // if schedule monitoring is enabled, record (or initialize)
             // the current schedule status
             bool isPastDue = false;
+
+            // we use DateTime.Now rather than DateTime.UtcNow to allow the local machine to set the time zone. In Azure this will be
+            // UTC by default, but can be configured to use any time zone if it makes scheduling easier.
             DateTime now = DateTime.Now;
             if (ScheduleMonitor != null)
             {
@@ -173,7 +176,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 return;
             }
 
-            // restart the timer with the next schedule occurrence
             await InvokeJobFunction(DateTime.Now, false);
 
             StartTimer(DateTime.Now);
@@ -255,6 +257,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
 
         private void StartTimer(TimeSpan interval)
         {
+            // Restart the timer with the next schedule occurrence, but only 
+            // if Cancel, Stop, and Dispose have not been called.
+            if (_cancellationTokenSource.IsCancellationRequested)
+            {
+                return;
+            }
+
             _timer = new System.Timers.Timer
             {
                 AutoReset = false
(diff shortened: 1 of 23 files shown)


test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
```diff
@@ -270,7 +270,23 @@ public void Timer_VerifyMaxInterval()
             Assert.True(updateCalled);
         }
 
-        private void CreateTestListener(string expression, bool useMonitor = true)
+        [Fact]
+        public async Task StoppedListener_DoesNotContinueRunning()
```
mathewc: This test reliably failed before your fix?

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -173,10 +173,14 @@ private void OnTimer(object sender, ElapsedEventArgs e)
                 return;
             }
 
-            // restart the timer with the next schedule occurrence
             await InvokeJobFunction(DateTime.Now, false);
 
-            StartTimer(DateTime.Now);
+            // Restart the timer with the next schedule occurrence, but only 
+            // if Cancel, Stop, and Dispose have not been called.
+            if (!_cancellationTokenSource.IsCancellationRequested)
```
mathewc: Should this check be moved into StartTimer? I see another StartTimer call above that isn't guarded.

mathewc:
When you say "should address" - I thought you had a repro. So aren't you certain this addresses the issue? :)

mathewc:
Other than my one comment, looks good. :ship:

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -173,10 +173,14 @@ private void OnTimer(object sender, ElapsedEventArgs e)
...
+            {
+                StartTimer(DateTime.Now);
```
MikeStall: I'm looking at a similar bug with DateTime right now and so this caught my eye. Would .UtcNow be more correct here?   .Now is timezone specific and can have oddities in different timezones and around daylight savings. We should always use Utc (timezone agnostic)

mathewc: We need to use Local time not UTC time for TimerTrigger, to allow people to control their time zone in their Web App. See https://github.com/Azure/azure-webjobs-sdk-extensions/issues/32 for all the details :) We used to use UTC then fixed the issues by moving to Local.

test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
brettsam: I ran it ~15 times and it failed every time.

---
In reply to: [78817781](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/149#discussion_r78817781) [](ancestors = 78817781)

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
brettsam: Yeah that should work too.

---
In reply to: [78818151](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/149#discussion_r78818151) [](ancestors = 78818151)

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
fabiocav: Would be nice to have a comment describing why we're not using UTC here to avoid confusion.

Issue #142, opened 2016-08-15
Move to latest SendGrid packages


mathewc:
We're currently on a very old version of the SendGrid SDK (6.3.4) and the latest stable is 8.0.2.

gorillapower:
This is a big problem for us as half our projects use the new SendGrid SDK >7.x.x and WebJobs requires <= SendGrid SDK 6.3.4. Its causing us a bit of a dependency nightmare. Is there a time frame on when this is scheduled to be complete?

mathewc:
Yes we'll be getting to this soon. I tried before when moving other dependencies, and there were a lot of non trivial breaking changes, so I backed that out.

mathewc:
@gorillapower This has been done. Once @brettsam publishes the packages to myget, you can start using the beta packages if you wish. We've addressed all their breaking changes.

gorillapower:
Awesome thanks all.

brettsam:
They're up already. Anything after 10369: https://www.myget.org/feed/azure-appservice/package/nuget/Microsoft.Azure.WebJobs.Extensions.SendGrid/2.0.0-beta1-10369

mathewc:
Note that you will have to absorb all the breaking changes as well - you can see our updated samples in this repo for guidance.

mguinness:
Any word on when the dependency on SendGrid SDK 6.3.4 will be removed for the latest stable version of Microsoft.Azure.WebJobs.Extensions.SendGrid in NuGet?  I'm unable to use the beta version from MyGet in production and would like some guidance on when the new version will become a release.

PR #152, opened 2016-09-23
update to SendGrid to use their v3 API; add attribute ApiKey support

brettsam:
Addresses #142 and https://github.com/Azure/azure-webjobs-sdk-script/issues/669 (although this will need some work in Script first). Also sets us up to address https://github.com/Azure/azure-webjobs-sdk-script/issues/640.

I haven't done any testing of this in Script yet but will get all that work done before merging this.

Notable changes:

- Refactored `SendGridConfiguration` so that it now supports setting different `ApiKey` values on attributes.
- Moves to SendGrid's v3 API, which required a decent amount of changes: https://sendgrid.com/docs/API_Reference/Web_API_v3/index.html
- Adds `string` support to the output binding. Since the underlying SendGrid client uses a JSON string, it seems silly not to support it directly.
- Added some new factory/client interfaces so I could write a couple of mocked E2E tests.
- Moved TargetFramework up to 4.5.2, which the SendGrid nuget required.


src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>ExtensionsSample</RootNamespace>
     <AssemblyName>ExtensionsSample</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
```
mathewc: No, not if everything still continues to work with the core SDK, script, etc.

src/ExtensionsSample/Samples/SendGridSamples.cs
```diff
@@ -19,14 +19,15 @@ public static class SendGridSamples
         public static void ProcessOrder_Declarative(
             [QueueTrigger(@"samples-orders")] Order order,
             [SendGrid(
+                ApiKey = "Test",
```
mathewc: This is optional, right? It will default if not specified here?

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
```diff
@@ -33,110 +35,80 @@ internal static bool TryParseAddress(string value, out MailAddress mailAddress)
...
+            return CreateMessage(json);
+        }
 
-            if (input.TryGetValue("subject", StringComparison.OrdinalIgnoreCase, out value))
-            {
-                message.Subject = (string)value;
-            }
+        internal static Mail CreateMessage(JObject input)
+        {
+            return input.ToObject<Mail>();
```
mathewc: Previously we would emit some nice error messages - do we still do that somewhere if the inputs are bad here?

src/WebJobs.Extensions.SendGrid/Bindings/SendGridMessageAsyncCollector.cs
```diff
@@ -33,7 +34,7 @@ public Task AddAsync(SendGridMessage item, CancellationToken cancellationToken =
 
             SendGridHelpers.DefaultMessageProperties(item, _config, _attribute);
 
-            if (item.To == null || item.To.Length == 0)
+            if (item.Personalization?.Count == 0)
```
mathewc: This error message is specific to a To address - shouldn't we be checking personalization.To?

src/ExtensionsSample/ErrorNotifier.cs
```diff
@@ -71,15 +71,17 @@ public void WebNotify(TraceFilter filter)
...
+                Subject = "WebJob Error Notification"
             };
-            message.To = new MailAddress[] { _sendGridConfig.ToAddress };
+            message.AddContent(new Content("text/plain", filter.GetDetailedMessage(5)));
+            Personalization personalization = new Personalization();
+            personalization.AddTo(_sendGridConfig.ToAddress);
+            message.AddPersonalization(personalization);
 
-            _sendGrid.DeliverAsync(message);
+            _sendGrid.client.mail.send.post(requestBody: message.Get());
```
mathewc: In other places below, I see code like `_sendGrid.SendMessageAsync` being used. What is the difference here?

src/WebJobs.Extensions.SendGrid/Bindings/SendGridScriptBindingProvider.cs
```diff
@@ -57,7 +57,7 @@ public override bool TryResolveAssembly(string assemblyName, out Assembly assemb
 
             if (string.Compare(assemblyName, "SendGridMail", StringComparison.OrdinalIgnoreCase) == 0)
```
mathewc: I thought assembly "SendGridMail" no longer exists?

src/WebJobs.Extensions.SendGrid/Config/ISendGridClientFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Extensions.Clients;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Config
+{
+    internal interface ISendGridClientFactory
```
mathewc: this probably belongs under the Clients namespace?

mathewc: Also, does the folder name "Clients" mean you're going to have multiple, or should it just be "Client"?

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
```diff
@@ -38,12 +50,58 @@ public SendGridConfiguration()
...
+        }
+
+        private void ValidateBinding(SendGridAttribute attribute, Type type)
+        {
+            string apiKey = FirstOrDefault(attribute.ApiKey, ApiKey, _defaultApiKey);
+
+            if (string.IsNullOrEmpty(apiKey))
+            {
+                throw new InvalidOperationException(
+                    string.Format("The SendGrid ApiKey must be set either via a '{0}' app setting, via a '{0}' environment variable, or directly in code via SendGridConfiguration.ApiKey.",
```
mathewc: Should this error message also mention that it can be set via SendGridAttribute.ApiKey?

mathewc:
Thanks for taking care of all this Brett :) I just had some minor comments. Otherwise this looks great. If any of those comments need to be addressed please do so, otherwise, if all the samples/tests continue to work with these changes, just check it in :)

mathewc:
There are also some minor updates that we need to make to the readme page at the root of this repo - it shows a SendGrid example. The difficulty there is that users of non preview versions of the extension won't have these updates, however I think the approach we've been taking with readme/docs is to keep them up to date with the latest.

src/WebJobs.Extensions.SendGrid/Bindings/SendGridScriptBindingProvider.cs
fabiocav: To avoid issues like this, I went with a different approach with the Twilio implementation:
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.Twilio/Bindings/TwilioScriptBindingProvider.cs#L26
Might be worth doing the same here.

mathewc: Yes, we should do that across the board.

src/ExtensionsSample/ErrorNotifier.cs
brettsam: SendMessageAsync is our method on the client wrapper ISendGridClient (for mocking). This is the actual SendGrid API for sending the request. This sample has always used the SendGrid API directly (it's showing how to use a `TraceFilter`) so I kept it that way.

src/WebJobs.Extensions.SendGrid/Bindings/SendGridMessageAsyncCollector.cs
brettsam: I'm wondering if we want to be in the business of providing error messages like this. I was looking through their error docs, and they'll give you very specific reasons why something doesn't work: https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html#-Personalizations-Errors. That would prevent us from accidentally short-circuiting something that could work in the future (what if they decide to add server-side 'to' templates or something)?

I'll tweak this to match their rules, but I get wary about trying to keep up with the validation rules of a service we don't own. Thoughts?

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
brettsam: We were basically doing our own JSON deserialziation previously so we'd throw an error if something didn't parse as we expected. But now the `Mail` object is a lot more complex -- https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html#-Request-Body-Parameters and the object supports JSON.NET properties: https://github.com/sendgrid/sendgrid-csharp/blob/master/SendGrid/SendGrid/Helpers/Mail/Mail.cs#L46. So I decided to remove all of that and just let JSON.NET handle the deserialization. The errors won't be as concise but I'm not sure we want to write code to handle all of the possibilities here.

src/ExtensionsSample/Samples/SendGridSamples.cs
brettsam: Yes, I meant to clean that up. The new E2E tests verify that it's optional and fallbacks work as expected.

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
brettsam: Yup. And added a test to make sure it's thrown during indexing.

src/WebJobs.Extensions.SendGrid/Bindings/SendGridMessageAsyncCollector.cs
mathewc: I'm on board with you making whatever changes are necessary to ensure a good error experience. I agree that we don't want to duplicate their validations. I made this comment because the error message was about a missing "to" field, but your changes were such that it could happen in other cases, so the error message could be misleading.

If your experimentation indicates that we don't need to be doing these validations, then remove them.

brettsam: Yeah, your point was valid :-) I've fixed that up and added a test for it. I think if this validation starts getting too crazy or it's a pain to keep up, it's not a big deal to remove it since we can trust the SendGrid server to do the right thing even if we do no validation.

Issue #158, opened 2016-10-12
ApiHubFile triggers do not get fired if the path gets updated after inital setup


safihamid:
ApiHubFile triggers do not get fired if the path gets updated after initial setup

#### Repro steps
Create a SaaSFile trigger function on a path, add some files to the path and then go back to the portal and edit the path or connection. the code still listens to the old path/connection.

#### Known workarounds
Create a new function for the new path or connection or go to the function storage and manually delete the blob related to that function.

mathewc:
Can you elaborate? In functions, any change to function.json triggers a host restart, and that should pull in the updates.

safihamid:
This is only related to ApiHubFile. we keep the state(last checked time, ...) in a blob for each function in azure-webjobs-hosts/apihubs/{func app name}/{func name}/status. the content includes the PollUri which we get from the connector. however after initial setup if the users goes back to portal and changes the path or connection, this blob entry does not get reset and the trigger still listens to the old path.

One solution I am thinking is to include the path as well as connector name in the blob path so instead of azure-webjobs-hosts/apihubs/{func app name}/{func name}/status the blob name should be azure-webjobs-hosts/apihubs/{func app name}/{func name}/{connector-name}-{path}-status

PR #159, opened 2016-10-13
Fix for https://github.com/Azure/azure-webjobs-sdk-extensions/issues/158

safihamid:
ApiHubFile triggers do not get fired if the path gets updated after inital setup


mathewc:
Need to add some tests for this.

PR #161, opened 2016-10-13
[NotificaitonHubs] Add EnableTestSend

### LINKED ISSUE #156: [NotificationHub]Add option to enable debug send
Please provide a succinct description of the issue.

NotificationHubClient has an option to [EnableTestSend](https://azure.microsoft.com/en-us/documentation/articles/notification-hubs-push-notification-fixer/). If this property is set, NotificationHubs gives detailed error message if sending push fails for some reason. Exposing this property in NH extension would help customers debug most common config issues while sending Push

pragnagopa:
Fixes #156

### DIFF
--- a/src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
+++ b/src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
@@ -4,6 +4,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.NotificationHubs;
+using Microsoft.Azure.WebJobs.Host;
 
 namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
 {
@@ -11,16 +12,35 @@ namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
     {
         private INotificationHubClientService _notificationHubclientService;
         private string _tagExpression;
+        private bool _enableTestSend;
+        private TraceWriter _traceWriter;
 
-        public NotificationHubAsyncCollector(INotificationHubClientService clientService, string tagExpression)
+        public NotificationHubAsyncCollector(INotificationHubClientService clientService, string tagExpression, bool enableTestSend, TraceWriter traceWriter)
         {
             _notificationHubclientService = clientService;
             _tagExpression = tagExpression;
+            _enableTestSend = enableTestSend;
+            _traceWriter = traceWriter;
         }
 
         public async Task AddAsync(Notification item, CancellationToken cancellationToken = default(CancellationToken))
         {
-            await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            NotificationOutcome notificationOutcome = await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            if (_enableTestSend)
+            {
+                string debugLog = $"NotificationHubs Test Send\r\n" +
+                    $"  TrackingId = {notificationOutcome.TrackingId}\r\n" +
+                    $"  State = {notificationOutcome.State}\r\n" +
+                    $"  Results (Success = {notificationOutcome.Success}, Failure = {notificationOutcome.Failure})\r\n";
+                if (notificationOutcome.Results != null)
+                {
+                    foreach (RegistrationResult result in notificationOutcome.Results)
+                    {
+                        debugLog += $"    ApplicationPlatform:{result.ApplicationPlatform}, RegistrationId:{result.RegistrationId}, Outcome:{result.Outcome}\r\n";
+                    }
+                }
+                _traceWriter.Info(debugLog);
+            }
         }
 
         public Task FlushAsync(CancellationToken cancellationToken = default(CancellationToken))
--- a/src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
+++ b/src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
 
         private string _defaultConnectionString;
         private string _defaultHubName;
-
+        
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
@@ -62,32 +62,33 @@ namespace Microsoft.Azure.WebJobs.Extensions.NotificationHubs
             var bindingFactory = new BindingFactory(nameResolver, converterManager);
             IBindingProvider clientProvider = bindingFactory.BindToExactType<NotificationHubAttribute, NotificationHubClient>(BindForNotificationHubClient);
 
-            var ruleOutput = bindingFactory.BindToAsyncCollector<NotificationHubAttribute, Notification>(BuildFromAttribute);
+            var ruleOutput = bindingFactory.BindToAsyncCollector<NotificationHubAttribute, Notification>((attribute) => BuildFromAttribute(attribute, context.Trace));
 
             extensions.RegisterBindingRules<NotificationHubAttribute>(ruleOutput, clientProvider);
         }
 
-        internal IAsyncCollector<Notification> BuildFromAttribute(NotificationHubAttribute attribute)
+        internal IAsyncCollector<Notification> BuildFromAttribute(NotificationHubAttribute attribute, TraceWriter trace)
         {
             string resolvedConnectionString = ResolveConnectionString(attribute.ConnectionStringSetting);
             string resolvedHubName = ResolveHubName(attribute.HubName);
+            bool enableTestSend = attribute.EnableTestSend;
 
-            INotificationHubClientService service = GetService(resolvedConnectionString, resolvedHubName);
-            return new NotificationHubAsyncCollector(service, attribute.TagExpression);
+            INotificationHubClientService service = GetService(resolvedConnectionString, resolvedHubName, enableTestSend);
+            return new NotificationHubAsyncCollector(service, attribute.TagExpression, attribute.EnableTestSend, trace);
         }
 
         internal NotificationHubClient BindForNotificationHubClient(NotificationHubAttribute attribute)
         {
             string resolvedConnectionString = ResolveConnectionString(attribute.ConnectionStringSetting);
             string resolvedHubName = ResolveHubName(attribute.HubName);
-            INotificationHubClientService service = GetService(resolvedConnectionString, resolvedHubName);
+            INotificationHubClientService service = GetService(resolvedConnectionString, resolvedHubName, attribute.EnableTestSend);
 
             return service.GetNotificationHubClient();
         }
 
-        internal INotificationHubClientService GetService(string connectionString, string hubName)
+        internal INotificationHubClientService GetService(string connectionString, string hubName, bool enableTestSend)
         {
-            return ClientCache.GetOrAdd(new Tuple<string, string>(connectionString, hubName.ToLowerInvariant()), (c) => NotificationHubClientServiceFactory.CreateService(c.Item1, c.Item2));
+            return ClientCache.GetOrAdd(new Tuple<string, string>(connectionString, hubName.ToLowerInvariant()), (c) => NotificationHubClientServiceFactory.CreateService(c.Item1, c.Item2, enableTestSend));
         }
 
         /// <summary>
--- a/src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
+++ b/src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
@@ -47,5 +47,11 @@ namespace Microsoft.Azure.WebJobs
         /// <see cref="NotificationHubsConfiguration"/>.
         /// </summary>
         public string HubName { get; set; }
+
+        /// <summary>
+        /// Optional. Boolean value to enable debug send on NotificationHubClient
+        /// <see cref="NotificationHubClient"/>.
+        /// </summary>
+        public bool EnableTestSend { get; set; }
     }
 }
--- a/test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubAsyncCollectorTests.cs
+++ b/test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubAsyncCollectorTests.cs
@@ -2,9 +2,11 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.Azure.NotificationHubs;
 using Microsoft.Azure.WebJobs.Extensions.NotificationHubs;
+using Microsoft.Azure.WebJobs.Extensions.Tests.Common;
 using Moq;
 using Xunit;
 
@@ -12,6 +14,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Tests.NotificationHubs
 {
     public class NotificationHubAsyncCollectorTests
     {
+       public static string debugLogNoResults = "NotificationHubs Test Send\r\n" +
+                    "  TrackingId = \r\n" +
+                    "  State = Enqueued\r\n" +
+                    "  Results (Success = 0, Failure = 0)\r\n";
+
         [Fact]
         public async Task AddAsync_SendNotification_TagExpression_Null()
         {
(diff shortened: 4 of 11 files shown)


test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubAsyncCollectorTests.cs
```diff
@@ -41,12 +43,15 @@ public class NotificationHubAsyncCollectorTests
             mockNhClientService.Setup(x => x.SendNotificationAsync(notification, "foo||bar"))
                     .Returns(Task.FromResult(new NotificationOutcome()));
 
-            IAsyncCollector<Notification> collector = new NotificationHubAsyncCollector(mockNhClientService.Object, "foo||bar");
+            TestTraceWriter trace = new TestTraceWriter(TraceLevel.Info);
+
+            IAsyncCollector<Notification> collector = new NotificationHubAsyncCollector(mockNhClientService.Object, "foo||bar", true, trace);
```
brettsam: It looks like this is testing the scenario where `notificationOutcome.Results` is null. You should have a test where there are actual results there as well to make sure that output works as expected. And as I mentioned elsewhere, also test that you write something out when Results is not null but the Results collection is empty.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,39 @@
...
-            await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            NotificationOutcome notificationOutcome = await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            if (_enableTestSend)
+            {
+                _traceWriter.Info(string.Format("TrackingId:{0}", notificationOutcome.TrackingId));
+                if (notificationOutcome.Results != null)
+                {
+                    foreach (RegistrationResult result in notificationOutcome.Results)
+                    {
+                        _traceWriter.Info(string.Format("ApplicationPlatform:{0} RegistrationId:{1} \nOutcome:{2}", result.ApplicationPlatform, result.RegistrationId, result.Outcome));
```
brettsam: nit: use C# 6 string interpolation here rather than `string.Format`.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,39 @@
...
+            _traceWriter = traceWriter;
         }
 
         public async Task AddAsync(Notification item, CancellationToken cancellationToken = default(CancellationToken))
         {
-            await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            NotificationOutcome notificationOutcome = await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            if (_enableTestSend)
+            {
+                _traceWriter.Info(string.Format("TrackingId:{0}", notificationOutcome.TrackingId));
```
brettsam: A few suggestions

1. Since these logs can all be interleaved between each other, you may want to build up a single string and call `.Info()` once so that it's all in one place. It can get confusing otherwise. You can use newlines to break it up to look nicer.
2. You may want to start that string with something that indicates it's coming from Notification Hubs. Otherwise you'll just see a log with `TrackingId:` and that may not have enough context.
3. Right now you're only logging for the case where you have results. You should also log if results are null or if results are empty. This looks like a debugging scenario that will only be turned on occasionally so you can be verbose about exactly what's happening.

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -43,6 +43,8 @@ public NotificationHubsConfiguration()
         /// </summary>
         public string HubName { get; set; }
 
+        internal TraceWriter Trace { get; set; }
```
brettsam: Do you need this? Doesn't look like you use it anywhere.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
pragnagopa: Done.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
pragnagopa: Good point. Added [NotificaitonHubDebugLog] as prefix

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
pragnagopa: Getting this from Initialize method and using it to create NH service

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,44 @@
...
+            _traceWriter = traceWriter;
         }
 
         public async Task AddAsync(Notification item, CancellationToken cancellationToken = default(CancellationToken))
         {
-            await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            NotificationOutcome notificationOutcome = await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            if (_enableTestSend)
+            {
+                _traceWriter.Info($"[NotificationHubsDebugLog] TrackingId = {notificationOutcome.TrackingId}");
```
brettsam: @mathewc -- Is there any pattern for debug logs like this? Not sure if there's a standard way to do it.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,44 @@
...
+                if (notificationOutcome.Results == null || notificationOutcome.Results.Count == 0)
+                {
+                    _traceWriter.Info($"[NotificationHubsDebugLog] NotificationOutcome.Results is null or empty." +
+                                     "See https://azure.microsoft.com/en-us/documentation/articles/notification-hubs-push-notification-fixer/ for deatils on how to diagnose issues while sending push notifications");
```
brettsam: I'm not sure if you need to include a link like this -- it may break at some point. I'd rather see that in the docs.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
brettsam: @pragnagopa -- It's not a huge deal, but I still think you should build up a string with these `if` clauses and write out once at the end of the method. Otherwise, you may not be able to link the TrackingId to the log below it (it's possible for other logs to sneak in between).

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
brettsam: Can you either (a) make it private, or (b) just pass the context.Trace directly to your function like I do here: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs#L55. It cleans up the class a little bit.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
brettsam: Also, if you were to keep this, you're missing a space after the first sentence. I called that out in a test below.

test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubAsyncCollectorTests.cs
```diff
@@ -41,12 +43,111 @@ public class NotificationHubAsyncCollectorTests
...
+
+            // Act
+            await collector.AddAsync(notification);
+
+            // Assert
+            Assert.Equal(2, trace.Events.Count);
+            Assert.True(trace.Events[0].Message.Contains("[NotificationHubsDebugLog] TrackingId"));
```
brettsam: Rather than doing `.Contains` on all of these, can you do `.Equals()`? I think you had some bad spacing above, for example, and matching the whole string here would help catch that.

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
brettsam: Also -- :-) -- You can probably just say `No results`. Showing the class/property like that may be too specific for someone using Node, for example.

mathewc: Agree with all of the above comments :)

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
mathewc: No pattern. The [NotificationHubsDebugLog] might be a little strange - aren't these logs written to the function trace output file? If so, then why is a prefix needed?

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
mathewc: Yeah, we don't want consumers of this config class to have to worry about TraceWriters

src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
```diff
@@ -47,5 +47,11 @@ public sealed class NotificationHubAttribute : Attribute
         /// <see cref="NotificationHubsConfiguration"/>.
         /// </summary>
         public string HubName { get; set; }
+
+        /// <summary>
+        /// Optional. Boolean value to enable debug send on NotificationHubClient
+        /// <see cref="NotificationHubClient"/>.
+        /// </summary>
+        public bool EnableTestSend { get; set; }
```
mathewc: Shouldn't this be a global config level setting rather than per attribute?

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
brettsam: Good point. I was specifically thinking about the console logger, which now logs output from functions. However, a better approach may be to have that logger prefix messages from functions with some common text. Pragna, I'd suggest removing that [NotificationHubsDebugLog] that I suggested :-)

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,44 @@
...
+                }
+                else
+                {
+                    foreach (RegistrationResult result in notificationOutcome.Results)
+                    {
+                        _traceWriter.Info($"[NotificationHubsDebugLog] ApplicationPlatform:{result.ApplicationPlatform} RegistrationId:{result.RegistrationId} Outcome:{result.Outcome}");
```
brettsam: That's what I had assumed -- so using .Info() made sense since you're explicitly toggling it on/off for testing. Otherwise you'd have to EnableTestSend and then also change your logging to Verbose in order to see any results.

pragnagopa: Yes. You have to turn it on and then turn if off as NH as limitation on sending pushes with debug turned on.

src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
pragnagopa: Exposing this setting at attribute level is being more explicit about turning on/off to debug sending push just that notification. We could add this as global setting, but we would have to add at attribute level as well as we have the ability to override global NH connection string and hub name.

brettsam: I was thinking about this like `RunOnStartup` for timers -- which I always think of as a debugging switch. You're likely going to want to debug a single function with this setting rather than just run debugging on across-the-board. I think it makes sense to be attribute-only.

test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubAsyncCollectorTests.cs
```diff
@@ -2,16 +2,21 @@
...
 using Microsoft.Azure.WebJobs.Extensions.NotificationHubs;
+using Microsoft.Azure.WebJobs.Extensions.Tests.Common;
 using Moq;
 using Xunit;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Tests.NotificationHubs
 {
     public class NotificationHubAsyncCollectorTests
     {
+        public static string debugLogNoResults = "TrackingId = .Results is null or empty. Refer to NotificationHubs documentaion for more details on how to debug issues while sending push notifications.";
```
brettsam: Since this is meant to be a debugging tool, let's give all the info that we can. I looked at the types and I think we can do something like this:

```
Notification Hubs Test Send
  NotificationId = notificationid
  TrackingId = trackingid
  State = Completed
  Results (0 success, 0 failure)
```
and for success
```
Notification Hubs Test Send
  NotificationId = notificationid
  TrackingId = trackingid
  State = Completed
  Results (3 success, 0 failure)
    ApplicationPlatform = Windows, RegistrationId = guid-guid-guid, Outcome = outcome
    ApplicationPlatform = Windows, RegistrationId = guid-guid-guid, Outcome = outcome
    ApplicationPlatform = Windows, RegistrationId = guid-guid-guid, Outcome = outcome
```

src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
pragnagopa: @brettsam / @mathewc  SO?

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,47 @@
...
+            {
+                string debugLog = $"NotificationHubs Test Send\r\n" +
+                    $"TrackingId = {notificationOutcome.TrackingId}\r\n" +
+                    $"State = {notificationOutcome.State}\r\n" +
+                    $"Results (Success = {notificationOutcome.Success}, Failure = {notificationOutcome.Failure})\r\n";
+                if (notificationOutcome.Results == null || notificationOutcome.Results.Count == 0)
+                {
+                    debugLog += $"Results are null or empty. Refer to NotificationHubs documentaion for more details on how to debug issues while sending push notifications.";
+                }
+                else
```
brettsam: I'd remove this log. It'll already show that they have 0 success and 0 failure and if they've turned this on, they likely know they can look at the docs. It'll be cleaner without this. So just write out the details below if Results != null

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,47 @@
...
         }
 
         public async Task AddAsync(Notification item, CancellationToken cancellationToken = default(CancellationToken))
         {
-            await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            NotificationOutcome notificationOutcome = await _notificationHubclientService.SendNotificationAsync(item, _tagExpression);
+            if (_enableTestSend)
+            {
+                string debugLog = $"NotificationHubs Test Send\r\n" +
+                    $"TrackingId = {notificationOutcome.TrackingId}\r\n" +
```
brettsam: Indent all of these in by one tab (or just use two spaces if that's easier -- not sure how tabs will translate in every logging situation)

src/WebJobs.Extensions.NotificationHubs/Bindings/NotificationHubAsyncCollector.cs
```diff
@@ -4,23 +4,47 @@
...
+                if (notificationOutcome.Results == null || notificationOutcome.Results.Count == 0)
+                {
+                    debugLog += $"Results are null or empty. Refer to NotificationHubs documentaion for more details on how to debug issues while sending push notifications.";
+                }
+                else
+                {
+                    foreach (RegistrationResult result in notificationOutcome.Results)
+                    {
+                        debugLog += $"\tApplicationPlatform:{result.ApplicationPlatform}, RegistrationId:{result.RegistrationId}, Outcome:{result.Outcome}\r\n";
+                    }
```
brettsam: indent these by 2 tabs (or 4 spaces) so they sit under the 'Results' label

mathewc:
:ship:

PR #163, opened 2016-10-18
Fix for ApiHubFile triggers to support Paths starting with or without a forward slash

safihamid:
This was discovered while investigating a customer issue and is to avoid confusion.


src/WebJobs.Extensions.ApiHub/Common/GenericFileTriggerBindingProvider.cs
```diff
@@ -89,7 +89,7 @@ private class GenericTriggerbinding : ITriggerBinding
                 MethodInfo methodInfo = (MethodInfo)parameter.Member;
                 _functionName = methodInfo.Name;
 
-                _bindingDataProvider = BindingDataProvider.FromTemplate(_attribute.Path, ignoreCase: true);
+                _bindingDataProvider = BindingDataProvider.FromTemplate(_attribute.Path.TrimStart('/'), ignoreCase: true);
```
mathewc: Why only trimming from the start?

mathewc:
:ship:

safihamid:
Merged

PR #164, opened 2016-10-20
Ensure ScheduledStatus.Next is adjusted if invocation time is fired too early

Thnslj:
Related to #132 

Fixes a bug that can make a `DailySchedule` or  `WeeklySchedule` fire twice in a row if the timer (due to timing skew), is fired before the intented time.
image

The `adjustedInvocationTime` were set to the `ScheduledStatus.Next` before calling `_schedule.GetNextOccurrence(adjustedInvocationTime)`.

But if the _scheduler is of type `DailySchedule` or `WeeklySchedule` the result of `_schedule.GetNextOccurrence(adjustedInvocationTime)` would still be the same as `ScheduleStatus.Next`


mathewc:
Hey, thanks for the PR :) I thought about this a bit more and I wonder if instead we shouldn't just fix the schedule implementations so they behave properly when now is equal to a schedule occurrence. I've submitted a PR for this here: https://github.com/Azure/azure-webjobs-sdk-extensions/pull/165. Can you review and let me know if this fixes the issues you're seeing?

PR #165, opened 2016-10-22
GetNextOccurrence Scheduling fix

mathewc:
GetNextOccurrence should not return same occurrence when now is equal to an occurrence. This fix is inspired by the issue reported in https://github.com/Azure/azure-webjobs-sdk-extensions/pull/164.


mathewc:
@Thnslj Can you review this?

mathewc:
Good point on the change in TimerSchedule - I've made that fix as well, thanks :)

mathewc:
Packages with this fix are now available on our myget feed (version 2.0.0-beta1-10380). Thanks.

PR #167, opened 2016-10-25
Timer: fixing 'PastDue' bug with clock skew

brettsam:
There was a bug here with how we stored the ScheduleStatus when we found clock skew and had to adjust the next invocation forward. We'd write the `Last` value as the actual time the trigger fired. However, if the host restarted, it'd take that time into consideration and think that the schedule had changed ([this code determines that](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs#L74-L86)). That would then incorrectly report that the timer was past due.

Now we'll store the adjusted invocation time, meaning all the future calculations should work as if the trigger fired exactly on time.

These things are tough to debug, so I added a couple of extra lines of verbose logging as well. That will help in the future should we need to do more investigations here.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -84,6 +84,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // check to see if we've missed an occurrence since we last started.
                 // If we have, invoke it immediately.
                 ScheduleStatus = await ScheduleMonitor.GetStatusAsync(_timerName);
+                _trace.Verbose($"Function '{_timerName}' initial status: Last='{ScheduleStatus?.Last.ToString("o")}', Next='{ScheduleStatus?.Next.ToString("o")}'");
                 TimeSpan pastDueDuration = await ScheduleMonitor.CheckPastDueAsync(_timerName, now, _schedule, ScheduleStatus);
                 isPastDue = pastDueDuration != TimeSpan.Zero;
             }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -225,16 +226,20 @@ private void OnTimer(object sender, ElapsedEventArgs e)
...
+                Last = adjustedInvocationTime,
                 Next = _schedule.GetNextOccurrence(adjustedInvocationTime)
             };
 
             if (ScheduleMonitor != null)
             {
                 await ScheduleMonitor.UpdateStatusAsync(_timerName, ScheduleStatus);
             }
+
+            _trace.Verbose($"Function '{_timerName}' updated status: Last='{ScheduleStatus.Last.ToString("o")}', Next='{ScheduleStatus.Next.ToString("o")}'");
```
mathewc: The other log entry you added is only written when schedule monitoring is on. Let's make this consistent. I assume your intent is to log this info only if schedule monitoring is on? If so, move this into the if block directly above.

mathewc:
TimerTrigger... the gift that keeps on giving :) :ship:

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
brettsam: My intent was to log the 'interesting' scenarios to aid debugging. However, I think you're right that printing out the current status for non-monitor scenarios (which will typically be those with less than 1 minute timers) isn't really that interesting and will result in a lot of noise. We've seen most of the issues surrounding longer timers using the monitor. I'll move this inside the block above.

Thnslj:
Is this why I am seeing extra "schedules" runs when i restart my webjobs?

brettsam:
@Thnslj -- it's possible. What version of Extensions are you running?

The symptoms are:
- A timer fires just before it was officially scheduled (like at 23:59:59.5 for a midnight timer).
- Your WebJob host restarts before the next timer fires.
- You see a log that this extra run was "past due".

Thnslj:
(version 2.0.0-beta1-10380) - The version that was suppose to fix my scheduling issues reported and resolved in pr #165 
I will try to update the the package again..

JasonLConway:
I have a CRON schedule of (0 45 21 * * *) and my function ran on its own this morning at (11/6/2016 11:36:22 AM) with a timer of (2016-11-06T11:36:19.9605284+00:00).  

I don't know how to prevent this function from running outside of the schedule.  What do I do?

brettsam:
@JasonLConway -- what version of Extensions are you running?

Thnslj:
My imput:
DailySchedules have been running smoothly for me since the last update.
image

Even when restarting and after a 00:59:59 run.

JasonLConway:
@brettsam - when the function ran outside of the schedule is was .8, I have since updated to .9

PR #175, opened 2016-11-09
Timer: implementing ISingletonRenewalMonitor

brettsam:
This goes with https://github.com/Azure/azure-webjobs-sdk/pull/878 to show how a listener would implement `ISingletonRenewalMonitor`.


brettsam:
We're not going to do this -- closing.

PR #182, opened 2017-01-05
documentdb query binding support

brettsam:
Addresses https://github.com/Azure/azure-webjobs-sdk-script/issues/289

### DIFF
--- a/src/ExtensionsSample/Samples/DocumentDBSamples.cs
+++ b/src/ExtensionsSample/Samples/DocumentDBSamples.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using ExtensionsSample.Models;
 using Microsoft.Azure.Documents;
 using Microsoft.Azure.Documents.Client;
@@ -25,14 +26,15 @@ namespace ExtensionsSample
         // Other supported types:
         //   out T[]
         //   IAsyncCollector<T>
-        //   ICollector<T>
+        //   ICollector<T>        
         public static void InsertDocument(
             [TimerTrigger("00:01")] TimerInfo timer,
             [DocumentDB("ItemDb", "ItemCollection")] out ItemDoc newItem)
         {
             newItem = new ItemDoc()
             {
-                Text = new Random().Next().ToString()
+                Text = new Random().Next().ToString(),
+                IsCompleted = new Random().Next() % 2 == 0
             };
         }
 
@@ -50,6 +52,24 @@ namespace ExtensionsSample
             [DocumentDB("ItemDb", "ItemCollection", Id = "{QueueTrigger}")] JObject item)
         {
             item["text"] = "Text changed!";
+        }
+
+        // Document query input binding
+        //   This binding requires the 'SqlQuery' property to be specified. The binding uses
+        //   that string to issue a query against the specified collection. The resulting collection is passed to the 
+        //   function parameter.
+        //
+        // This example uses the binding template "{IsCompleted}" to specify that the value of 'c.isCompleted' should come
+        // from the IsCompleted proparty of the queued JSON message.
+        public static void QueryDocument(
+            [QueueTrigger("samples-documentdb-csharp")] CustomQueueInput input,
+            [DocumentDB("ItemDb", "ItemCollection", SqlQuery = "SELECT c.id, c.text FROM c where c.isCompleted = {IsCompleted}")] IEnumerable<string> items,
+            TraceWriter log)
+        {
+            foreach (var doc in items)
+            {
+                log.Info($"Text: {doc}");
+            }
         }
 
         // DocumentClient input binding
--- a/src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
@@ -1,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Documents;
+using Microsoft.Azure.Documents.Client;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
+{
+    internal class DocumentDBEnumerableBuilder<T> : IAsyncConverter<DocumentDBAttribute, IEnumerable<T>> where T : class
+    {
+        private readonly DocumentDBConfiguration _config;
+
+        public DocumentDBEnumerableBuilder(DocumentDBConfiguration config)
+        {
+            _config = config;
+        }
+
+        public async Task<IEnumerable<T>> ConvertAsync(DocumentDBAttribute attribute, CancellationToken cancellationToken)
+        {
+            DocumentDBContext context = _config.CreateContext(attribute);
+
+            Uri collectionUri = UriFactory.CreateDocumentCollectionUri(context.ResolvedAttribute.DatabaseName, context.ResolvedAttribute.CollectionName);
+
+            List<T> finalResults = new List<T>();
+
+            string continuation = null;
+
+            SqlQuerySpec sqlSpec = new SqlQuerySpec
+            {
+                QueryText = context.ResolvedAttribute.SqlQuery,
+                Parameters = context.ResolvedAttribute.SqlQueryParameters ?? new SqlParameterCollection()
+            };
+
+            do
+            {
+                DocumentQueryResponse<T> response = await DocumentDBUtility.RetryAsync(
+                    () => context.Service.ExecuteNextAsync<T>(collectionUri, sqlSpec, continuation),
+                    context.MaxThrottleRetries);
+
+                finalResults.AddRange(response.Results);
+                continuation = response.ResponseContinuation;
+            }
+            while (!string.IsNullOrEmpty(continuation));
+
+            return finalResults;
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBJArrayBuilder.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBJArrayBuilder.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB.Bindings
+{
+    internal class DocumentDBJArrayBuilder : IAsyncConverter<DocumentDBAttribute, JArray>
+    {
+        private DocumentDBEnumerableBuilder<JObject> _builder;
+
+        public DocumentDBJArrayBuilder(DocumentDBConfiguration config)
+        {
+            _builder = new DocumentDBEnumerableBuilder<JObject>(config);
+        }
+
+        public async Task<JArray> ConvertAsync(DocumentDBAttribute attribute, CancellationToken cancellationToken)
+        {
+            IEnumerable<JObject> results = await _builder.ConvertAsync(attribute, cancellationToken);
+            return JArray.FromObject(results);
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
+++ b/src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Documents;
 using Microsoft.Azure.WebJobs.Extensions.DocumentDB;
 
 namespace Microsoft.Azure.WebJobs
@@ -90,5 +92,15 @@ namespace Microsoft.Azure.WebJobs
         /// collection.
         /// </summary>
         public int CollectionThroughput { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// When specified on an input binding using an <see cref="IEnumerable{T}"/>, defines the query to run against the collection. 
+        /// May include binding parameters.
+        /// </summary>
+        [AutoResolve(ResolutionPolicyType = typeof(DocumentDBSqlResolutionPolicy))]
+        public string SqlQuery { get; set; }
+
+        internal SqlParameterCollection SqlQueryParameters { get; set; }
     }
 }
--- a/src/WebJobs.Extensions.DocumentDB/Services/DocumentDBService.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Services/DocumentDBService.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Azure.Documents;
 using Microsoft.Azure.Documents.Client;
+using Microsoft.Azure.Documents.Linq;
 using Microsoft.Azure.WebJobs.Extensions.DocumentDB.Config;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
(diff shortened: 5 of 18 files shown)


src/WebJobs.Extensions.DocumentDB/Services/DocumentDBService.cs
```diff
@@ -66,6 +67,19 @@ public IOrderedQueryable<DocumentCollection> CreateDocumentCollectionQuery(Uri c
             return response.Resource;
         }
 
+        public async Task<DocumentQueryResponse> ExecuteNextAsync(Uri documentCollectionUri, string sqlExpression, string continuation)
+        {
+            FeedOptions feedOptions = new FeedOptions { RequestContinuation = continuation, EnableCrossPartitionQuery = true };
+            IDocumentQuery<Document> query = _client.CreateDocumentQuery<Document>(documentCollectionUri, sqlExpression, feedOptions).AsDocumentQuery();
+            FeedResponse<Document> response = await query.ExecuteNextAsync<Document>();
```
brettsam: I get the below error when I run this in a 32-bit process (which xunit is by default). I've started a conversation with the DocDB team:

`Partition routing information cannot be extracted from the query when running in a 32-bit process. To complete your query and avoid this exception, ensure that your host process is 64-bit. For Executable projects, this can be done by unchecking the "Prefer 32-bit" option in the project properties window, on the Build tab.`

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -52,6 +53,19 @@ public static class DocumentDBSamples
             item["text"] = "Text changed!";
         }
 
+        // Document query input binding
+        //   This binding requires teh 'SqlQuery' property to be specified. The binding uses
```
mathewc: nit: teh

src/ExtensionsSample/Samples/DocumentDBSamples.cs
```diff
@@ -52,6 +53,19 @@ public static class DocumentDBSamples
...
+        //   that string to issue a query against the specified collection. The resulting collection is passed to the 
+        //   function parameter.
+        //
+        // This example uses the binding template "{QueueTrigger}" to specify that the value of 'c.isCompleted' should come
+        // from the string value of the queued item.
+        public static void QueryDocument(
+            [QueueTrigger("samples-documentdb-csharp")] string input,
+            [DocumentDB("ItemDb", "ItemCollection", SqlQuery = "SELECT c.id, c.text FROM c where c.isCompleted = {QueueTrigger}")] IEnumerable<ItemDoc> items)
```
mathewc: DocDB APIs handle SQL injection prevention I assume?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableValueBinder.cs
```diff
@@ -0,0 +1,70 @@
...
+            Uri collectionUri = UriFactory.CreateDocumentCollectionUri(_context.ResolvedAttribute.DatabaseName, _context.ResolvedAttribute.CollectionName);
+
+            List<Document> finalResults = new List<Document>();
+
+            string continuation = null;
+
+            do
+            {
+                DocumentQueryResponse response = DocumentDBUtility.RetryAsync(
+                    () => _context.Service.ExecuteNextAsync(collectionUri, _context.ResolvedAttribute.SqlQuery, continuation),
```
mathewc: This value provider works for both queries as well as non query enumerables? E.g. if no SqlQuery is specified.

brettsam: It doesn't currently, but it could. If you ask for an IEnumerable<T> and don't specify a query, we'll return the whole collection. I'll add that.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableValueBinder.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                if (response == null)
+                {
+                    return null;
+                }
+
+                finalResults.AddRange(response.Results);
+                continuation = response.ResponseContinuation;
+            }
+            while (!string.IsNullOrEmpty(continuation));
```
mathewc: Wondering how Take/Top is handled? E.g. if the user IEnumerable binds to a table with a billion documents, we'll pull all of those into memory? We have some similar issues with our table binding query support (TableAttributeBindingProvider) https://github.com/Azure/azure-webjobs-sdk/issues/903.

src/ExtensionsSample/Samples/DocumentDBSamples.cs
brettsam: Good point. They do support SqlParameters. I'll have to think through how we make that work.

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
```diff
@@ -90,5 +91,13 @@ public DocumentDBAttribute(string databaseName, string collectionName)
         /// collection.
         /// </summary>
         public int CollectionThroughput { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// When specified on an input binding using an <see cref="IEnumerable{T}"/>, defines the query to run against the collection.
```
mathewc: What happens if someone specifies a SqlQuery that returns a single row, and binds it to a singleton type? Do we somehow prevent this?

brettsam: The plan was that if you supply `SqlQuery`, you must use an IEnumerable -- I'm adding tests to make sure that's the case. As well as make sure if you specify both `Id` and `SqlQuery`, we throw, as that is unclear.

So if you expect a single result from a query, you'll need to pull it from the returned enumerable.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableValueBinder.cs
brettsam: If you want to restrict it, you'd need to put a `SELECT TOP X` in your query. So yeah it's possible to pull a billion rows.

If someone wanted to page through a massive collection, they'd have to write that themselves with DocumentClient directly. We could probably come up with a solution (for a future release) that simplifies this -- we'd probably use the pattern for Tables as well.

src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs
mathewc: Yeah, that's the behavior I'd expect - just wanted to make sure we had the proper validations/user guidance in place

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableValueBinder.cs
mathewc: Sounds good

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableValueBinder.cs
```diff
@@ -0,0 +1,70 @@
...
+    internal class DocumentDBEnumerableValueBinder<T> : IValueBinder
+    {
+        private DocumentDBContext _context;
+
+        public DocumentDBEnumerableValueBinder(DocumentDBContext context)
+        {
+            _context = context;
+        }
+
+        public Type Type
```
mamaso: could use an expression bodied member to get a little more compact

`public Type Type => typeof(IEnumerable<T>)`

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableValueBinder.cs
mamaso: It might be worthwhile to add some tracing around this / enhanced errors.  

For script languages (which include binding time in overall timeout time) we could probably decrease confusion if, instead of throwing a generic timeout error, we indicate that the timeout occurred due to the query and not user code.

Not in the scope of this PR, think it's more of an invoker issue

brettsam:
@mamaso / @mathewc -- this is ready for a final review now that all the pieces in place.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBScriptBindingProvider.cs
```diff
@@ -74,7 +75,14 @@ public override Type DefaultType
                 {
                     if (Context.Access == FileAccess.Read)
                     {
-                        return typeof(JObject);
+                        if (string.Compare("many", Context.Cardinality, StringComparison.OrdinalIgnoreCase) == 0)
+                        {
+                            return typeof(IEnumerable<JObject>);
```
mamaso: Feels stranger than JArray, but makes sense in the context of that email thread. Hopefully we can light up JArray out of the box. For DocDB converters you only need to support IEnumerable<T>, correct?

brettsam: Yeah, only IEnumerable. If I could come up with an easy converter to JArray, I would support it, but all my attempts were failing.

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/DocumentDBConfigurationTests.cs
```diff
@@ -86,6 +118,33 @@ public void CreateContext_UsesDefaultRetryValue()
             Assert.Equal(DocumentDBContext.DefaultMaxThrottleRetries, context.MaxThrottleRetries);
         }
 
+        [Theory]
+        [InlineData(typeof(IEnumerable<string>), true)]
+        [InlineData(typeof(IEnumerable<Document>), true)]
+        [InlineData(typeof(IEnumerable<JObject>), true)]
+        [InlineData(typeof(JArray), false)]
+        [InlineData(typeof(string), false)]
+        [InlineData(typeof(List<Document>), false)]
```
mamaso: could make the argument we want to support types that implement IEnumerable but it's probably better to keep it simple

mamaso:
:shipit:

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/DocumentDBConfigurationTests.cs
brettsam: I was thinking that too, but then realized that string implements IEnumerable<char>, which wouldn't make sense to support. Once I see one special-case, it makes me nervous that there are a bunch of others, too.

Plus, how would I construct the types? If someone wanted a `MyCustomEnumerable<Document>`, I'd need to do some reflection to create it before returning it. It felt like a lot of work and testing for not a lot of value.

Plus, we can always expand that later.

mathewc:
@brettsam Is the plan to get this into the 2.0 release, or do you want to wait until after to give it bake time?

brettsam:
The plan is to get it in ASAP. We're having some discussions about the best way to enable this in Script.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Documents;
+using Microsoft.Azure.Documents.Client;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
+{
+    internal class DocumentDBEnumerableBuilder<T> : IAsyncConverter<DocumentDBAttribute, IEnumerable<T>> where T : class
+    {
+        private DocumentDBConfiguration _config;
```
MikeStall: nit: readonly?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            do
+            {
+                DocumentQueryResponse response = await DocumentDBUtility.RetryAsync(
+                    () => context.Service.ExecuteNextAsync(collectionUri, sqlSpec, continuation),
+                    context.MaxThrottleRetries, codesToIgnore: HttpStatusCode.NotFound);
+
+                if (response == null)
+                {
+                    return null;
```
MikeStall: >return null; [](start = 20, length = 12)

null  for an IEnumerable  seems wrong - I'd expect the caller to soon null-ref. Is this an error case that should be an exception? Or should this be 0-length array?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+                }
+
+                finalResults.AddRange(response.Results);
+                continuation = response.ResponseContinuation;
+            }
+            while (!string.IsNullOrEmpty(continuation));
+
+            return finalResults.Select(doc => (T)(dynamic)doc);
```
MikeStall: >(T)(dynamic) [](start = 46, length = 12)

What is this doing?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBJArrayBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB.Bindings
+{
+    internal class DocumentDBJArrayBuilder : IAsyncConverter<DocumentDBAttribute, JArray>
```
MikeStall: >DocumentDBJArrayBuilder [](start = 19, length = 23)

This is Matt's temp work around?

brettsam: Yes

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBScriptBindingProvider.cs
MikeStall: Is this the place that should be JArray instead? We don't want to require a corresponding script  change. 

---
In reply to: [102275264](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/182#discussion_r102275264) [](ancestors = 102275264)

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
brettsam: Changes a Document into any type. A non-generic example from the DocDB docs is here: https://github.com/Azure/azure-documentdb-dotnet/blob/master/samples/code-samples/DocumentManagement/Program.cs#L175

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
brettsam: I don't think this can ever happen -- I'll change it to throw an exception here as it would indicate something is wrong.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBScriptBindingProvider.cs
brettsam: Yeah, good catch. I forgot to change that from my initial implementation.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
MikeStall: but what's happening? Is doc already typed as T? Is this applying some deserialization magic? 

---
In reply to: [102564906](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/182#discussion_r102564906) [](ancestors = 102564906)

brettsam: `doc` is a `Document` (a DocumentDB Type), which implements `IDynamicMetaObjectProvider`, allowing it to be cast as a `dynamic`. Once you have a `dynamic`, you cast it T. There's a chance it fails (if your properties don't match).

brettsam: @MikeStall -- I went ahead and removed this completely in my latest PR. We now use the generic version from DocDB so they own all the deserialization and casting.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
brettsam: It turns out that 'null' was only used if the collection was not found. I looked at how Table works, and it just rethrows the exception, which I think is more discoverable. 'Null' doesn't really help there. So I've changed this to rethrow instead of using null.

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                QueryText = context.ResolvedAttribute.SqlQuery,
+                Parameters = context.ResolvedAttribute.SqlQueryParameters ?? new SqlParameterCollection()
+            };
+
+            do
+            {
+                DocumentQueryResponse<T> response = await DocumentDBUtility.RetryAsync(
+                    () => context.Service.ExecuteNextAsync<T>(collectionUri, sqlSpec, continuation),
+                    context.MaxThrottleRetries);
```
mamaso: how does removing 'codesToIgnore' affect this? I'm assuming the not found will surface as a function error.. not sure you can get a not found for a query in any event.

One thing to consider is handling 429s - I know Parse had a ton of issues with too many requests errors causing strange bugs. Probably not in the scope of this initial codeflow, but may want to chat with the docdb team about it.

mamaso:
current iteration looks good to me

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBEnumerableBuilder.cs
brettsam: 404s -- which will only happen if the collection or db don't exist -- will surface as a function error. I checked how Table works in this regard, and it matches that behavior.

Yeah, there were 429 issues that I worked through a while ago. Although I may need to revisit as it appears newest DocDB packages will handle 429s internally now (with a few knobs). So I may not even need any of this code.

Issue #194, opened 2017-02-24
TimerTrigger can miss IsPastDue


brettsam:
Investigating this issue: http://stackoverflow.com/questions/42386299/azure-function-on-time-trigger-not-firing

The scenario I'm seeing:

| Time| Message           | 
| ------------- |-------------| 
| 2017-02-21 00:59:11.4454545 | Function 'Host.Functions.UpdateData' initial status: Last='0001-01-01T00:00:00.0000000', Next='2017-02-21T01:00:00.0000000+00:00' |
| 2017-02-21 00:59:11.4454758 | The next 5 occurrences of the schedule will be: 2/21/2017 1:00:00 AM, *etc* | 

... then a quick restart and:

| Time| Message           | 
| ------------- |-------------| 
| 2017-02-21 01:00:12.0984654 | Function 'Host.Functions.UpdateData' initial status: Last='0001-01-01T00:00:00.0000000', Next='2017-02-21T01:00:00.0000000+00:00' |
| 2017-02-21 01:00:12.1164413 | The next 5 occurrences of the schedule will be: 2/22/2017 1:00:00 AM, *etc* |

I think this logic is flawed: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs#L67-L72

In this case, b/c `now` is slightly past the execution time, we're re-calculating it and ignoring the stored `Next` value. I believe if there's a stored `Next`, we should take that into account. We'll need to look through the tests and see what scenario the current logic is enabling before tweaking anything.

brettsam:
This test looks flawed: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/FileSystemScheduleMonitorTests.cs#L159-L196

Because the schedule is mocked, it's returning `GetNextOccurence()` times *in the past*, which gives us the 'past due' calculation that we expect. In reality, that will never happen.

I believe we check for a 'never run' function is because we don't want to rely on the stored `Next` value. If someone has changed the schedule since that was stored, then the `Next` value is invalid and we need to re-calculate it.

I'm thinking that storing an `LastUpdated` value along with the `Next` and `Last` values would give us all the details we want. Then we could check whether the calculation from that time has changed -- if it has, we know the schedule has changed. This needs to work with existing stored schedules, though.

[...]

PR #195, opened 2017-02-27
Pastdue

brettsam:
Addressing #194.

The problem is that if the function has never run, we have no `Last` value to use for calculating `Next` when the host starts. So instead, we calculate from `now`. If the timer is past due, that means we'll miss it. Some customers have seen this happen when their site wakes up and restarts for one reason or another -- the timer doesn't get around to evaluating until a couple seconds after the scheduled execution. At that point, we recalculate the next execution and don't know that we've missed one.

This change adds a `LastUpdated` value to our serialized status. Even if the function has never run, we'll update `LastUpdated` to be the last time that we used when calculating `Next`. So when the host restarts, we have something to work with. 

I've added tests to make sure that the 'legacy' behavior still works as expected -- since no one will have a `LastUpdated` value when this code starts running.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
@@ -56,7 +56,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
                 lastStatus = new ScheduleStatus
                 {
                     Last = default(DateTime),
-                    Next = nextOccurrence
+                    Next = nextOccurrence,
+                    LastUpdated = now
                 };
                 await UpdateStatusAsync(timerName, lastStatus);
                 recordedNextOccurrence = nextOccurrence;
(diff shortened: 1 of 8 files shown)


brettsam:
Also now addresses https://github.com/Azure/azure-webjobs-sdk-script/issues/1246.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
```diff
@@ -56,32 +56,57 @@ public abstract class ScheduleMonitor
...
             }
             else
             {
                 DateTime expectedNextOccurrence;
-                if (lastStatus.Last == default(DateTime))
+
+                // Track the time that was used to create 'expectedNextOccurrence'.
+                DateTime lastUpdated;
+
+                if (lastStatus.LastUpdated != default(DateTime))
```
mamaso: nit: could collapse this whole if else to something like:

```
DateTime lastUpdated = lastStatus?.LastUpdated ?? lastStatus?.Last ?? now;
expectedNextOccurrence = schedule.GetNextOccurrence(lastUpdated);
```

I like the comments, def helps the grok.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
```diff
@@ -56,32 +56,57 @@ public abstract class ScheduleMonitor
...
                 }
 
                 // ensure that the schedule hasn't been updated since the last
                 // time we checked, and if it has, update the status to use the new schedule
                 if (lastStatus.Next != expectedNextOccurrence)
                 {
+                    // if the schedule has changed and the next occurrence is in the past,
+                    // recalculate it based on the current time as we don't want it to register
+                    // immediately as 'past due'.
+                    if (now > expectedNextOccurrence)
```
mamaso: Just to be clear, we are treating 'now' as if it is the exact moment of schedule change, correct? 

nit: Maybe we could route this via a function call to consolidate this to a invalid/no last status codepath, as this moment basically says 'throw out everything we know and start over'.

mamaso: May want to change `CheckPastDueAsync` comments above to make sure they're consistent with this new delay behavior

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
```diff
@@ -56,32 +56,57 @@ public abstract class ScheduleMonitor
...
-                    // there have been no executions of the function yet, so compute
-                    // from now
-                    expectedNextOccurrence = schedule.GetNextOccurrence(now);
+                    // If we know when we last calculated the stored 'Next', use the current schedule
+                    // to recalculate. It may have changed, which will be checked below.
+                    expectedNextOccurrence = schedule.GetNextOccurrence(lastStatus.LastUpdated);
+                    lastUpdated = lastStatus.LastUpdated;
                 }
-                else
+                else if (lastStatus.Last != default(DateTime))
```
mathewc: Shouldn't use of Last if present take precedence over use of the new LastUpdated?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
brettsam: I agree -- but this is really confusing when you come back to it several months later (like always happens :-)), so I wanted to make sure the code and comments were verbose enough to be understood.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
brettsam: If LastUpdated exists, it will always be the same as Last. I'll look and see if I can easily swap them around and still have it be readable (I think I can).

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
brettsam: Yes, think of LastUpdated as "the time we used when we last calculated Next". It's real value is allowing you to look at the current schedule and recalculate 'Next' to see if it has changed. We don't trust 'Next' unless we recalculate it ourselves on startup with the current schedule.

Does that make sense? I know it can get confusing -- hence the verbose comments :-)

mathewc:
Shipit

PR #196, opened 2017-02-27
Removing dependency on Microsoft.Azure.WebJobs.Script.Extensibility

mathewc:
Required to 2.0 RTM this project. Since Microsoft.Azure.WebJobs.Script.Extensibility is a prerelease package currently undergoing redesign, we need to remove this dependency. A non-prerelease package cannot have a prerelease dependency.

Also removing the prerelease tag and moving to the latest 2.0 SDK package versions.


mathewc:
Merged

PR #204, opened 2017-03-22
Update webjobs nuget 10543

mamaso:


fabiocav:
What's included in this update? Is this just bringing the latest bug fix that was made in the SDK?

mamaso:
yep. I need the nuget for a bot apps update

PR #208, opened 2017-04-06
Update SendGrid extension to lightup with BYOB.

MikeStall:
This lets us remove the SendGrindScriptBinding in the Script repro.


src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
```diff
@@ -68,28 +66,35 @@ public void Initialize(ExtensionConfigContext context)
                 throw new ArgumentNullException("context");
             }
 
-            INameResolver nameResolver = context.Config.NameResolver;
-            _defaultApiKey = nameResolver.Resolve(AzureWebJobsSendGridApiKeyName);
+            var metadata = new ConfigMetadata();
```
MikeStall: This is pulling the config from Host.json. This is code that used to be in SendGridScriptBinding.Initialize.  The ConfigMetadata class matches the existing SendGrid config schema (which has slightly different naming than this config object)

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
```diff
@@ -102,5 +107,15 @@ private static string FirstOrDefault(params string[] values)
         {
             return values.FirstOrDefault(v => !string.IsNullOrEmpty(v));
         }
+
+        // Schema for host.json 
+        private class ConfigMetadata
+        {
+            [JsonProperty("to")]
```
MikeStall: Note that there's a slight name mismatch. The host.json says "to" and "from"; whereas the config says "toAddress" and "fromAddress". Having an intermediate metadata object like this clearly captures that.

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
```diff
@@ -68,28 +66,35 @@ public void Initialize(ExtensionConfigContext context)
...
-            IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
-            extensions.RegisterBindingRules<SendGridAttribute>(ValidateBinding, nameResolver, outputProvider);
+            context
+                .AddConverter<string, Mail>(SendGridHelpers.CreateMessage)
+                .AddConverter<JObject, Mail>(SendGridHelpers.CreateMessage)
+                .AddBindingRule<SendGridAttribute>()
+                    .BindToCollector<Mail>(CreateCollector);
+        }
+
+        private IAsyncCollector<Mail> CreateCollector(SendGridAttribute attr)
```
mamaso: out of curiosity, is the BindToCollector function executed per function invocation or per indexed attribute?

MikeStall: Per invocation. That's because the attribute can have { } which are resolved to invocation specific items.

PR #210, opened 2017-04-11
ApiHub Files to support Imperative Binding even if there are no declarative ApiHubs.

safihamid:
ApiHub Files to support Imperative Binding even if there are no declarative ApiHubs.

Found the issue while replying to this post:

http://stackoverflow.com/questions/43348985/dynamic-output-file-name-of-apihubfile-azure-function-binding-one-drive-drop-b


src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
```diff
@@ -154,7 +154,25 @@ private Task<ApiHubFile> BuildFromAttribute(ApiHubFileAttribute attribute)
...
         private IFolderItem GetFileSource(string key)
         {
-            return _map[key];
+            if (_map.ContainsKey(key))
+            {
+                return _map[key];
+            }
+            else
+            {
+                // If the key doesn't exist, see if it is available as an environment variable.
```
MikeStall: What is an "imperative binding scenario"?

safihamid: Basically pass the Binder object and do the binding inside the method/function itself.
Found the issue by this post:
http://stackoverflow.com/questions/43348985/dynamic-output-file-name-of-apihubfile-azure-function-binding-one-drive-drop-b

src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
```diff
@@ -154,7 +154,25 @@ private Task<ApiHubFile> BuildFromAttribute(ApiHubFileAttribute attribute)
...
+                // This might be the case for imperative binding scenarios.
+                var connectionString = Environment.GetEnvironmentVariable(key);
+                if (!string.IsNullOrWhiteSpace(connectionString))
+                {
+                    AddConnection(key, connectionString);
```
MikeStall: >AddConnection [](start = 20, length = 13)

Is the bug that AddConnection is currently only called  when there's an entry in Function.json? So for IBinder scenarios, it's skipped.  ([https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/src/WebJobs.Script/Binding/ApiHubScriptBindingProvider.cs#L105](https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/src/WebJobs.Script/Binding/ApiHubScriptBindingProvider.cs#L105) )

safihamid: Yep

src/WebJobs.Extensions.ApiHub/Config/ApiHubConfiguration.cs
```diff
@@ -154,7 +154,25 @@ private Task<ApiHubFile> BuildFromAttribute(ApiHubFileAttribute attribute)
...
-            return _map[key];
+            if (_map.ContainsKey(key))
+            {
+                return _map[key];
+            }
+            else
+            {
+                // If the key doesn't exist, see if it is available as an environment variable.
+                // This might be the case for imperative binding scenarios.
+                var connectionString = Environment.GetEnvironmentVariable(key);
```
MikeStall: >Environment.GetEnvironmentVariable [](start = 39, length = 34)

We should never call this directly and instead go through the INameResolver. (like we do in Script)

PR #211, opened 2017-04-12
Adding HttpTriggerAttribute

ahmelsayed:


WebJobs.Extensions.Http/HttpTriggerAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        /// <summary>
+        /// The function HTTP authorization level.
+        /// </summary>
+        public AuthLevel AuthLevel { get; set; } = AuthLevel.Function;
+
+        /// <summary>
+        /// Allowed HTTP methods.
+        /// </summary>
+        public IEnumerable<HttpMethod> Methods { get; set; };
```
ahmelsayed: Ideally I'd have this as 
```csharp
public IEnumerable<HttpMethod> Methods { get; set; } = Enumerable.Empty<HttpMethod>();
```
To avoid a null collection, but I think the semantic is different between an empty array and not defined, right?

WebJobs.Extensions.Http/HttpTriggerAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        /// <summary>
+        /// The type of WebHook handled by the trigger (if handling a pre-defined WebHook).
+        /// </summary>
+        public string WebHookType { get; set; }
+
+        /// <summary>
+        /// The function HTTP authorization level.
+        /// </summary>
+        public AuthLevel AuthLevel { get; set; } = AuthLevel.Function;
```
ahmelsayed: Not sure if I should have a default, or if I should require it in a constructor.

WebJobs.Extensions.Http/HttpTriggerAttribute.cs
mamaso: Correct - empty array allows no methods, null collection allows all methods. Null's the better default.

https://github.com/Azure/azure-webjobs-sdk-script/blob/3b3e8cb8312060a475d4eb0418fafb4d80f12326/src/WebJobs.Script/Binding/Http/HttpRouteFactory.cs#L40

mathewc:
Replaced by https://github.com/Azure/azure-webjobs-sdk-extensions/pull/213

PR #212, opened 2017-04-17
IoTHub extension with byob changes

mamaso:


src/WebJobs.Extensions.IoTHub/IoTHubConfigProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <inheritdoc/>
+        protected virtual void Dispose(bool disposing)
```
mamaso: @MikeStall do we have a story for handling extensions which need to dispose of resources? I can imagine that this could leak resources as a script binding with host restarts

src/WebJobs.Extensions.IoTHub/MessageAsyncCollector.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public Task AddAsync(Message item, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return _client.SendAsync(_deviceId, item);
+        }
+
+        public Task FlushAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return Task.FromResult(true);
```
MikeStall: >FromResult [](start = 24, length = 10)

nit: we now have Task.CompletedTask.

src/WebJobs.Extensions.IoTHub/IoTHubConfigProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+                .BindToCollector(attr => new MessageAsyncCollector(GetClient(attr), attr.DeviceId));
+            context.AddBindingRule<IoTHubAttribute>()
+                .BindToInput(attr => GetClient(attr));
+        }
+
+        private ServiceClient GetClient(IoTHubAttribute input)
+        {
+            return ClientCache.GetOrAdd(input.ConnectionString, (cs) => {
+                var client = ServiceClient.CreateFromConnectionString(cs);
+                client.OpenAsync().GetAwaiter().GetResult();
```
MikeStall: >.GetAwaiter().GetResult() [](start = 34, length = 25)

I guess we need an BindToCollector overload that accepts an IAsyncConverter so that we can avoid this?

src/WebJobs.Extensions.IoTHub/IoTHubConfigProvider.cs
MikeStall: We have one for the listeners; but not for extensions in general.  The extensions bindings are generally stateless and cleaned up when the function returns. 
Can we do that pattern here? 


---
In reply to: [111744252](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/212#discussion_r111744252) [](ancestors = 111744252)

src/WebJobs.Extensions.IoTHub/IoTHubConfigProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// Binding provider for IoTHub
+    /// </summary>
+    public class IoTHubConfigProvider : IExtensionConfigProvider, IDisposable
+    {
+        private readonly ConcurrentDictionary<string, ServiceClient> ClientCache = new ConcurrentDictionary<string, ServiceClient>();
+        private readonly static JsonSerializer _serializer = new JsonSerializer();
+        private bool _disposed = false;
+
+        /// <inheritdoc/>
+        public void Initialize(ExtensionConfigContext context)
```
MikeStall: Does this need any configuration from host.json? 
You can now pull that via context.ApplyConfig(this, "iot");

MikeStall: I guess not needed since there aren't any public properties on this config.  

---
In reply to: [111806252](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/212#discussion_r111806252) [](ancestors = 111806252)

src/WebJobs.Extensions.IoTHub/IoTHubConfigProvider.cs
mamaso: Would a finalizer which handles disposal work in the case of host restarts?

I think that this is something we should consider - an extension maintaining connections between invocations seems like desired behavior, and appears to be the case here & with document db.  I need to investigate how service bus / event hubs handles this

MikeStall: This is tricky:
1. We definitely don't want finalizers. Deterministic dispose would be better than finalizers. 
2. Configs are currently potentially used across multiple hosts; so diu
3. This is potentially long running, so the state could hold things open for a long time. 


---
In reply to: [111821779](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/212#discussion_r111821779) [](ancestors = 111821779)

mamaso: @MikeStall what do you mean by 'diu?'?

Also, do we have any 'after execution' hooks for bindings?

mamaso:
Added #1137 and #1138, closing this for now

PR #214, opened 2017-04-20
Updating ExecutionContext with function name and directory

fabiocav:


MikeStall:
:shipit:

src/WebJobs.Extensions/Extensions/Core/Bindings/ExecutionContextBindingProvider.cs
```diff
@@ -63,10 +60,17 @@ public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
...
+            }
+
+            private ExecutionContext CreateContext(Guid invocationId)
+            {
+                return new ExecutionContext
                 {
-                    InvocationId = context.FunctionInstanceId
-                });
+                    InvocationId = invocationId,
+                    FunctionName = _parameter.Member.Name,
```
mamaso: I think we have a few different ways of getting FunctionName in the core sdk (FullName, ShortName, etc) - is this compatible?  What's the intended usage?

fabiocav: This will just be passed as some context that may be consumed by the function. The function name here will match the method name, which will be the function name in the context of WebJobs. Happy to change this to something else, but based on the conversations I've had with @MikeStall and @mathewc, this seemed correct.

PR #216, opened 2017-04-21
Add rich request information to binding data.

MikeStall:
Flows authorization information to other bindings.
https://github.com/Azure/azure-webjobs-sdk-script/issues/1418

Requires updated SDK from https://github.com/Azure/azure-webjobs-sdk/pull/1117

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -15,6 +15,7 @@ using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Azure.WebJobs.Host.Triggers;
+using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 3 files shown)


MikeStall:
Also fixes https://github.com/Azure/azure-webjobs-sdk-script/issues/591

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -215,12 +220,33 @@ internal static void ApplyBindingData(object target, IDictionary<string, object>
 
             internal static async Task<IReadOnlyDictionary<string, object>> GetRequestBindingDataAsync(HttpRequestMessage request, Dictionary<string, Type> bindingDataContract = null)
             {
+                var requestBinding = new JObject();
```
mathewc: We need to think how this relates to the existing request data we make available in functions. Currently on the context object we have:

```javascript
context.req
context.req.headers
context.req.body
context.req.originalUrl
context.req.method
context.query
```

We are translating the request in to an IDictionary<string, object> already for context. This is just not part of binding data. I'm wondering if we should move some of this logic to the http binding...

mathewc:
Replaced by https://github.com/Azure/azure-webjobs-sdk-extensions/pull/225

PR #217, opened 2017-04-21
Adding AttributeTargets.ReturnValue to bindings

soninaren:


fabiocav:
Same concern as what I brought up in https://github.com/Azure/azure-webjobs-sdk/pull/1118

Since it won't work for WebJobs SDK users, so the attribute usage will be misleading in those scenarios.

I would recommend not having this in place until this is properly implemented and supported there.

PR #221, opened 2017-05-01
Adding a constructor to the HttpTriggerAttribute to specify only methods

soninaren:
Right now through HttpTrigger forces you to add authtype if you want to restrict httpTrigger to a certain http methods. However the if the HttpTrigger is of type webhook Authlevel does not apply. Adding another constructor to specify only methods parameter.

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttribute.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttribute.cs
@@ -24,6 +24,15 @@ namespace Microsoft.Azure.WebJobs
 
         /// <summary>
         /// Constructs a new instance.
+        /// </summary>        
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.Http/HttpTriggerAttribute.cs
```diff
@@ -24,6 +24,16 @@ public HttpTriggerAttribute()
 
         /// <summary>
         /// Constructs a new instance.
+        /// </summary>        
+        /// <param name="methods">The http methods to allow.</param>
+        public HttpTriggerAttribute(params string[] methods)
+        {
+            AuthLevel = AuthorizationLevel.Function;
```
mathewc: Nit: rather than duplicate this line here, just call base ctor this()

PR #225, opened 2017-05-08
Expanding HttpTrigger binding data support

mathewc:
Adding `IDictionary<string, string>` bindings for request query and header collections. Adding dictionary bindings for these collections is consistent with what we do for other bindings (e.g. Blob bindings expose dictionaries for Metadata/Properties). See [binding doc](https://github.com/Azure/azure-webjobs-sdk/wiki/Trigger-Binding-Data#blobtrigger) for more details.

In conjunction with the recent addition of path expression support to the core binding pipeline, the following binding expressions for http are enabled:

```csharp
public static void TestFunction(
    [HttpTrigger("post")] HttpRequestMessage req,
    [Blob("test-output/test-{query.testId}-{headers.testSuffix}")] out string blob,
    IDictionary<string, string> headers,
    IDictionary<string, string> query)
{
    blob = headers["testValue"];
}
```

### DIFF
--- a/src/ExtensionsSample/ExtensionsSample.csproj
+++ b/src/ExtensionsSample/ExtensionsSample.csproj
@@ -70,10 +70,10 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Azure.WebJobs, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.Azure.WebJobs.Core.2.1.0-beta1-10711\lib\net45\Microsoft.Azure.WebJobs.dll</HintPath>
+      <HintPath>..\..\packages\Microsoft.Azure.WebJobs.Core.2.1.0-beta1-10729\lib\net45\Microsoft.Azure.WebJobs.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Azure.WebJobs.Host, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.Azure.WebJobs.2.1.0-beta1-10711\lib\net45\Microsoft.Azure.WebJobs.Host.dll</HintPath>
+      <HintPath>..\..\packages\Microsoft.Azure.WebJobs.2.1.0-beta1-10729\lib\net45\Microsoft.Azure.WebJobs.Host.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Data.Edm, Version=5.8.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Microsoft.Data.Edm.5.8.2\lib\net40\Microsoft.Data.Edm.dll</HintPath>
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -20,6 +20,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
 {
     internal class HttpTriggerAttributeBindingProvider : ITriggerBindingProvider
     {
+        internal const string HttpQueryKey = "Query";
+        internal const string HttpHeadersKey = "Headers";
+
         public HttpTriggerAttributeBindingProvider()
         {
         }
@@ -71,7 +74,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                 _parameter = parameter;
                 _isUserTypeBinding = isUserTypeBinding;
 
-                Dictionary<string, Type> aggregateDataContract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
                 if (_isUserTypeBinding)
                 {
                     // Create the BindingDataProvider from the user Type. The BindingDataProvider
@@ -79,35 +81,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                     // bindings (i.e. the properties of the POCO can be bound to by other bindings).
                     // It is also used to extract the binding data from an instance of the Type.
                     _bindingDataProvider = BindingDataProvider.FromType(parameter.ParameterType);
-                    if (_bindingDataProvider.Contract != null)
-                    {
-                        aggregateDataContract.AddRange(_bindingDataProvider.Contract);
-                    }
                 }
 
-                // add any route parameters to the contract
-                if (!string.IsNullOrEmpty(attribute.Route))
-                {
-                    var routeParameters = _httpRouteFactory.GetRouteParameters(attribute.Route);
-                    var parameters = ((MethodInfo)parameter.Member).GetParameters().ToDictionary(p => p.Name, p => p.ParameterType, StringComparer.OrdinalIgnoreCase);
-                    foreach (string parameterName in routeParameters)
-                    {
-                        // don't override if the contract already includes a name
-                        if (!aggregateDataContract.ContainsKey(parameterName))
-                        {
-                            // if there is a method parameter mapped to this parameter
-                            // derive the Type from that
-                            Type type;
-                            if (!parameters.TryGetValue(parameterName, out type))
-                            {
-                                type = typeof(string);
-                            }
-                            aggregateDataContract[parameterName] = type;
-                        }
-                    }
-                }
-
-                _bindingDataContract = aggregateDataContract;
+                _bindingDataContract = GetBindingDataContract(attribute, parameter);
             }
 
             public IReadOnlyDictionary<string, Type> BindingDataContract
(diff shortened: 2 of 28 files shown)


src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -70,10 +70,10 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Azure.WebJobs, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.Azure.WebJobs.Core.2.1.0-beta1-10711\lib\net45\Microsoft.Azure.WebJobs.dll</HintPath>
+      <HintPath>..\..\packages\Microsoft.Azure.WebJobs.Core.2.1.0-beta1-10729\lib\net45\Microsoft.Azure.WebJobs.dll</HintPath>
```
mathewc: Moving to latest to get @MikeStall's new binding path expression support

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -71,43 +74,16 @@ public HttpTriggerBinding(HttpTriggerAttribute attribute, ParameterInfo paramete
...
 
-                Dictionary<string, Type> aggregateDataContract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
                 if (_isUserTypeBinding)
                 {
                     // Create the BindingDataProvider from the user Type. The BindingDataProvider
                     // is used to define the binding parameters that the binding exposes to other
                     // bindings (i.e. the properties of the POCO can be bound to by other bindings).
                     // It is also used to extract the binding data from an instance of the Type.
                     _bindingDataProvider = BindingDataProvider.FromType(parameter.ParameterType);
-                    if (_bindingDataProvider.Contract != null)
```
mathewc: Factoring all this static contract init from the ctor to keep it cohesive

test/WebJobs.Extensions.Tests/Extensions/Http/HttpTriggerEndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Tests.Common;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Tests.Extensions.Http
+{
+    public class HttpTriggerEndToEndTests
```
mathewc: Getting some e2e tests in place in this repo for the binding

test/WebJobs.Extensions.Tests/Extensions/Http/HttpTriggerEndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+                Assert.Equal("123", query["q1"]);
+                Assert.Equal("two", query["q2"]);
+
+                Assert.Equal("value1", headers["h1"]);
+                Assert.Equal("value2", headers["h2"]); 
+            }
+
+            public static void TestFunction2(
+                [HttpTrigger("post")] HttpRequestMessage req,
+                [Blob("test-output/test-{query.testId}-{headers.testSuffix}")] out string blob,
```
mathewc: Here is the main test demonstrating binding expressions using query/headers

brettsam (approved):
This looks great. I don't see any issues.

mamaso (approved):

MikeStall:
>     public class HttpTriggerBindingTests

I added some new tests ([https://github.com/Azure/azure-webjobs-sdk-extensions/pull/216/files#diff-301e809733ae6ed936c955d395fca605](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/216/files#diff-301e809733ae6ed936c955d395fca605) ). Can you include those here to make sure we're covering those cases? Specifically GetRequestBindingDataAsync_ReadsFrom_Duplicates,  and GetRequestBindingDataAsync_Auth_Header 

---
Refers to: test/WebJobs.Extensions.Tests/Extensions/Http/HttpTriggerBindingTests.cs:21 in d1c23fc. [](commit_id = d1c23fc21ba96f25e1a686e13518446d24a06bbb, deletion_comment = False)

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -213,12 +189,62 @@ internal static void ApplyBindingData(object target, IDictionary<string, object>
...
+                // add query parameter collection to binding contract
+                if (!aggregateDataContract.ContainsKey(HttpQueryKey))
+                {
+                    aggregateDataContract.Add(HttpQueryKey, typeof(IDictionary<string, string>));
+                }
+
+                return aggregateDataContract;
+            }
+
             internal static async Task<IReadOnlyDictionary<string, object>> GetRequestBindingDataAsync(HttpRequestMessage request, Dictionary<string, Type> bindingDataContract = null)
```
MikeStall: Our Node conversion logic does this same HttpRequest-->JObject. Can these paths share?

mathewc: We're not converting anything to JObject here. We're exposing dictionaries, and in Functions, we have the inverse problem of exposing these dictionaries as ExpandoObjects to Node. When I pull these updates into Functions I'll consider if there are any code sharing possibilities or simplifications that can be made.

MikeStall:
:shipit:

TheeJamesLee:
Hi, was this accidentally closed?

seniorquico:
@TheeJamesLee I'm a little confused trying to trace the history of this change. I got here via discussions on [#591](https://github.com/Azure/azure-webjobs-sdk-script/issues/591) and #216. I wasn't able to find what merge actually makes this work, but here's what I just tried successfully (I'm using the Visual Studio tools to create pre-compiled Functions):

```csharp
public static HttpResponseMessage Debug(
    // HTTP Request (Input)
    [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "users/{userId}")] HttpRequestMessage request,
    IDictionary<string, string> headers,
    IDictionary<string, string> query,
    string userId,
    // Azure Blob Storage (Output)
    [Blob("users/{userId}.json", FileAccess.ReadWrite)] ICloudBlob blob,
    // Misc
    CancellationToken cancellationToken)
{
    ...
}
```

`headers`, `query`, and `userId` are all populated with HTTP request values.

I was able to make use of the headers dictionary to create a [custom binding extension](https://github.com/Azure/azure-webjobs-sdk/wiki/Extensibility) that parses the authorization token into a custom object (by using the [AutoResolve] attribute with the expression "{headers.authorization}").

TheeJamesLee:
(Whilst writing this reply, another question appeared - could my issue just be a limitation of the DocumentDB extension? I would have thought that the "{bindingValueRef}" would be common, but thought I'd openly ask)

> public static void TestFunction(
    [HttpTrigger("post")] HttpRequestMessage req,
    **[Blob("test-output/test-{query.testId}-{headers.testSuffix}")] out string blob,**
    IDictionary<string, string> headers,
    IDictionary<string, string> query)
{
    blob = headers["testValue"];
}

(Bold line is what I am looking to achieve in a DocumentDB SQL statement)

-------------------------------------------------------------------------

@seniorquico Thanks for the reply. I've had a quick go using the method you've described, but it doesn't achieve this PR's function from what I have tried (Header & query dictionaries are populated, however they are not usable in bindings)... As a work around on Friday before your reply I'm just initializing an instance of a DocDB client in the function method that needs the header value and take it from the headers dictionary, so the AutoResolve doesn't seem to give me anything 'earlier' (ie, before the function method and usable in the binding) and makes me put a dependency (Service locator pattern time..) in the method itself...

> "I need to access the {headers.headerKeyValue} in another input binding".

Now, I believe that the ability to do this was achieved by this specific pull request and proved by this E2E test?  https://github.com/Azure/azure-webjobs-sdk-extensions/pull/225/files#r115345510 
Which leads me back to the original question,
>  Hi, was this accidentally closed? or replaced by another request? Thanks :)

Seeing that @MikeStall commented with ship it above, then @mathewc closed it, rather than merging it? Forgive me if I have completely missed something here... but it does just look like someone mistakenly clicked the wrong button. :)


However, @seniorquico, here is an exact sample of the problem... I am trying to use a header value in a DocumentDB binding SQL statement, like so:

(VS preview 15.3, Azure compiled functions)
Packages:
Microsoft.Azure.Webjobs (2.1.0-beta1-10811)
Microsoft.Azure.Webjobs.Extensions (2.1.0-beta1-10450)
Microsoft.Azure.Webjobs.Extensions.DocumentDB (1.1.0-beta1-10450)
Microsoft.Azure.Webjobs.Extensions.Http (1.0.0-beta1-10450)

```
public static async Task<HttpResponseMessage> Run(
            [HttpTrigger("get", Route = "test")]HttpRequestMessage req,
            IDictionary<string, string> headers,
            IDictionary<string, string> query,
            [DocumentDB("dbName", "dbColName",
            ConnectionStringSetting = "DocumentDbConnectionString", 
            SqlQuery = "SELECT * FROM c WHERE c.SomeValue = {headers.testHeaderValue}")]
        IEnumerable<object> documents,
            TraceWriter log)
        {

            return req.CreateResponse(HttpStatusCode.OK);
        }
```

```
HTTP GET  api/test Header: testHeaderValue : "tv"
```

> "statusCode": 500, "messsage": "Exception while executing function: Functions.GMCDH -> Exception binding parameter 'documents' -> Error while accessing 'testHeaderValue': property doesn't exist.",
    "errorDetails": "Microsoft.Azure.WebJobs.Host.FunctionInvocationException : Exception while executing function: Functions.GMCDH ---> System.InvalidOperationException : Exception binding parameter 'documents' ---> System.InvalidOperationException : Error while accessing 'testHeaderValue': property doesn't exist.\r\n   at Microsoft.Azure.WebJobs.Host.Bindings.Path.BindingTemplateToken.ExpressionToken.Evaluate(IReadOnlyDictionary`2 bindingData)\r\n 

If you remove the DocumentDB binding line, you can see that the header property is available within the function, which was looks to be the initial part of this pull request.

Thanks again for your help,

seniorquico:
@TheeJamesLee Got it. Not sure if there's a difference between our dependency versions:

Microsoft.Azure.WebJobs (2.1.0-beta1)
Microsoft.Azure.WebJobs.Extensions (2.1.0-beta1)
Microsoft.NET.Sdk.Functions (1.0.0-alpha6)

However, an expression like "{headers.name}" is working in my custom binding attribute:

```csharp
[AttributeUsage(AttributeTargets.Parameter)]
[Binding]
public sealed class CustomAttribute : Attribute
{
    [AutoResolve] 
    public string Tag { get; set; }
}

public sealed class ExtensionConfigProvider : IExtensionConfigProvider
{
    public void Initialize(ExtensionConfigContext context)
    {
        var rule = context.AddBindingRule<CustomAttribute>();
        rule.BindToInput(attribute => attribute.Tag);
    }
}

[FunctionName("GetUser")]
public static Task<HttpResponseMessage> Debug(
    // HTTP Request (Input)
    [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "users/{userId}")] HttpRequestMessage request,
    IDictionary<string, string> headers,
    IDictionary<string, string> query,
    string userId,
    // Azure Blob Storage (Output)
    [Custom(Tag = "{headers.tag}")] string tag,
    // Misc
    CancellationToken cancellationToken)
{
    // tag == headers["Tag"]
}
```

My custom attribute is [*not* annotated the same way](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.DocumentDB/DocumentDBAttribute.cs#L103) as `DocumentDBAttribute.SqlQuery`. Maybe the resolution policy is the culprit?

TheeJamesLee:
@seniorquico I had a look around the resolution policies and think you're spot on and my tests seem to prove it.
Thanks,

KirillOsenkov:
@mathewc I love that this is supported, but where is this documented? It took a long time for me to figure out the syntax (by reading through this PR).

We should really document this syntax over at:

 * https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-expressions-patterns
 * https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=csharp
 * https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings?tabs=csharp
 * https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-blob-output?tabs=csharp
 * https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook

We should also have a sample of an Azure Function that is run by an Http Trigger, pushes the request content to a new blob in blob storage, and binds the blob name to the current date/time and a custom request header value. It was highly non-trivial to figure out.

(see related issue: https://github.com/Azure/azure-functions-host/issues/591)

mathewc:
Thanks for the feedback. You can use the feedback links on those doc pages to provide feedback and request changes/improvements.

PR #227, opened 2017-05-12
Allowing objects to flow through binding data

mathewc:
Required to support new path based binding expressions for non strongly typed languages in Functions (e.g. Node.js)

### DIFF
--- a/src/WebJobs.Extensions.Http/Utility.cs
+++ b/src/WebJobs.Extensions.Http/Utility.cs
@@ -23,13 +23,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                 string json = value as string;
                 if (!string.IsNullOrEmpty(json) && Utility.IsJson(json))
                 {
-                    // parse the object skipping any nested objects (binding data
-                    // only includes top level properties)
+                    // parse the object adding top level properties
                     JObject parsed = JObject.Parse(json);
                     var additionalBindingData = parsed.Children<JProperty>()
-                        .Where(p => p.Value != null &&
-                        (p.Value.Type != JTokenType.Object & p.Value.Type != JTokenType.Array))
-                        .ToDictionary(p => p.Name, p => (string)p);
+                        .Where(p => p.Value != null && (p.Value.Type != JTokenType.Array))
+                        .ToDictionary(p => p.Name, p => ConvertPropertyValue(p));
 
                     if (additionalBindingData != null)
                     {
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.Http/Utility.cs
```diff
@@ -23,13 +23,11 @@ internal static void ApplyBindingData(object value, Dictionary<string, object> b
                 string json = value as string;
                 if (!string.IsNullOrEmpty(json) && Utility.IsJson(json))
                 {
-                    // parse the object skipping any nested objects (binding data
-                    // only includes top level properties)
+                    // parse the object adding top level properties
                     JObject parsed = JObject.Parse(json);
                     var additionalBindingData = parsed.Children<JProperty>()
-                        .Where(p => p.Value != null &&
-                        (p.Value.Type != JTokenType.Object & p.Value.Type != JTokenType.Array))
```
mathewc: Previously we were filtering out nested objects from binding data.

test/WebJobs.Extensions.Tests/Extensions/Http/HttpTriggerBindingTests.cs
```diff
@@ -29,9 +29,12 @@ public class HttpTriggerBindingTests
             request.Content = new StringContent(input);
             var bindingData = await HttpTriggerAttributeBindingProvider.HttpTriggerBinding.GetRequestBindingDataAsync(request);
 
-            Assert.Equal(4, bindingData.Count);
+            Assert.Equal(5, bindingData.Count);
             Assert.Equal("testing", bindingData["test"]);
             Assert.Equal("123", bindingData["baz"]);
+
+            dynamic nestedObject = bindingData["nestedObject"];
```
MikeStall: >dynamic [](start = 12, length = 7)

Can we make this a static type instead of dynamic.  Is it JObject? Dict? Expando? 
This is a test, so we can enforce the strict behavior and catch it if it changes.  Dynamic is too loose.

MikeStall:
:shipit:

src/WebJobs.Extensions.Http/Utility.cs
```diff
@@ -51,6 +49,18 @@ internal static void ApplyBindingData(object value, Dictionary<string, object> b
...
 
+        private static object ConvertPropertyValue(JProperty property)
+        {
+            if (property.Value != null && property.Value.Type == JTokenType.Object)
+            {
+                return (JObject)property.Value;
+            }
+            else
+            {
+                return (string)property.Value;
```
MikeStall: >string [](start = 24, length = 6)

What about JArray?   (contrast to .NET's object model, that's not a JOBject)

test/WebJobs.Extensions.Tests/Extensions/Http/HttpTriggerBindingTests.cs
mathewc: It's a JObject. I'm using dynamic because it makes the property access easier

MikeStall: Or this would be fine too: 
   dynamic nestedObject = (JObject) bindingData["nestedObject"]; 
Think of the  extra (Jobj) cast as a micro-test that the return type is really JObj and doesn't get changed to IDict<string,obj> later on

PR #228, opened 2017-05-15
Fix docdb issue with duplicate SQL query parameters

anthonychu:
The DocDB binding currently does not handle duplicate parameters in `SqlQuery`. For instance:

    SELECT * FROM c WHERE c.id = {foo} AND c.value = {foo}

This results in an exception: `mscorlib: An item with the same key has already been added.`

- Changed behavior such that only a single `SqlParameter` is created for duplicate parameter names
- Fixed "same key" exception
- Added tests


brettsam:
Thanks @anthonychu. Great fix and tests. It looks like we have some of our tests mis-categorized, which is why there's a failure in AppVeyor. I'll make sure we can get a full clean test run and then I'll merge this.

PR #229, opened 2017-05-16
adding missing E2E test traits

brettsam:
PR #228 showed that we're missing a couple of traits so community test runs in AppVeyor don't complete successfully.


brettsam:
This is straightforward and low-risk enough that I'm going to merge it.

Issue #160, opened 2016-10-13
Add support for DocumentDB trigger binding


Biment:
DocumentDB trigger binding can be a great feature for the use cases which require reacting to changes in the collections.

christopheranderson:
This is something we've started investigating. Will likely land after GA of Functions and 2.0.0 of WebJobs SDK.

stephlocke:
There is now an API of deltas for DocDB that could be used for this:
https://azure.microsoft.com/en-us/blog/introducing-change-feed-support-in-azure-documentdb/

mathewc:
Yep - we'll be working with the DocDB team to get a trigger binding in place :)

vip32:
any ETA?

PR #244, opened 2017-06-29
CosmosDB Change Feed trigger

ealsur:
This change adds the `[CosmosDBTrigger]` extension that allows users to subscribe to collection changes.

When changes are detected, the trigger will fire and provide the user with a `IReadOnlyList<Document>` to read and use the documents. The Trigger can also be chained with the `DocumentDB` output binding as demonstrated in one of the samples.

Due to this ability to chain Trigger and Binding, the Trigger has a separate explicit set of Connection Strings to avoid infinite loops. Support for AppSettings, Configuration and Attribute set of the Connection Strings has also been added.

Internally it uses the [ChangeFeedProcessor](https://github.com/Azure/azure-documentdb-dotnet/tree/master/samples/ChangeFeedProcessor) distributed as a Nuget package. Package is still in publishing process.

Samples were added also to the ExtensionsSample Project.

### DIFF
--- a/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
@@ -6,6 +6,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Threading.Tasks;
+using Microsoft.Azure.Documents.ChangeFeedProcessor;
 using Microsoft.Azure.Documents.Client;
 using Microsoft.Azure.WebJobs.Extensions.DocumentDB.Bindings;
 using Microsoft.Azure.WebJobs.Host;
@@ -40,6 +41,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
         /// </summary>
         public string ConnectionString { get; set; }
 
+        /// <summary>
+        /// Gets or sets the lease options for the DocumentDB Trigger. 
+        /// </summary>
+        public ChangeFeedHostOptions LeaseOptions { get; set; }
+
         /// <inheritdoc />
         public void Initialize(ExtensionConfigContext context)
         {
@@ -73,6 +79,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 
             IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
             extensions.RegisterBindingRules<DocumentDBAttribute>(ValidateConnection, nameResolver, outputProvider, clientProvider, jArrayProvider, enumerableProvider, inputProvider);
+
+            context.Config.RegisterBindingExtensions(new CosmosDBTriggerAttributeBindingProvider(ResolveConnectionString(null), ResolveConnectionString(null), LeaseOptions));
         }
 
         internal static void ValidateInputBinding(DocumentDBAttribute attribute, Type parameterType)
@@ -151,7 +159,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
             // Finally, fall back to the default.
             return _defaultConnectionString;
         }
-
+        
         internal IDocumentDBService GetService(string connectionString)
         {
             return ClientCache.GetOrAdd(connectionString, (c) => DocumentDBServiceFactory.CreateService(c));
--- a/CustomDictionary.xml
+++ b/CustomDictionary.xml
@@ -65,6 +65,9 @@
       <Word>Twilio</Word>
       <Word>IEnumerable</Word>
       <Word>sqlquery</Word>
+      <Word>DocumentDB</Word>
+      <Word>AccountEndpoint</Word>
+      <Word>AccountKey</Word>
     </Recognized>
     <Deprecated/>
     <Compound>
--- a/WebJobs.Extensions.sln
+++ b/WebJobs.Extensions.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.12
+VisualStudioVersion = 15.0.26430.6
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{1118E0AD-83C8-4B67-859B-FCDEF54338C1}"
 EndProject
--- a/src/ExtensionsSample/App.config
+++ b/src/ExtensionsSample/App.config
@@ -53,6 +53,10 @@
         <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-4.0.1.0" newVersion="4.0.1.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.Azure.Documents.Client" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-1.13.0.0" newVersion="1.13.0.0" />
+      </dependentAssembly>
     </assemblyBinding>
   </runtime>
   <startup>
--- a/src/ExtensionsSample/ExtensionsSample.csproj
+++ b/src/ExtensionsSample/ExtensionsSample.csproj
@@ -57,9 +57,8 @@
       <HintPath>..\..\packages\Microsoft.Azure.ApiHub.Sdk.0.7.2-alpha\lib\net45\Microsoft.Azure.ApiHub.Sdk.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Microsoft.Azure.Documents.Client, Version=1.11.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.Azure.DocumentDB.1.11.4\lib\net45\Microsoft.Azure.Documents.Client.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="Microsoft.Azure.Documents.Client, Version=1.13.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Microsoft.Azure.DocumentDB.1.13.2\lib\net45\Microsoft.Azure.Documents.Client.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Azure.KeyVault.Core">
       <HintPath>..\..\packages\Microsoft.Azure.KeyVault.Core.1.0.0\lib\net40\Microsoft.Azure.KeyVault.Core.dll</HintPath>
@@ -113,7 +112,6 @@
     </Reference>
     <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Newtonsoft.Json.9.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
-      <Private>True</Private>
     </Reference>
     <Reference Include="RestSharp, Version=105.2.3.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\..\packages\RestSharp.105.2.3\lib\net46\RestSharp.dll</HintPath>
@@ -241,6 +239,7 @@
     <Compile Include="SamplesTypeLocator.cs" />
     <Compile Include="Samples\ApiHubTableSamples.cs" />
     <Compile Include="Samples\ApiHubSamples.cs" />
+    <Compile Include="Samples\CosmosDBTriggerSamples.cs" />
     <Compile Include="Samples\MobileAppSamples.cs" />
     <Compile Include="Samples\DocumentDBSamples.cs" />
     <Compile Include="GlobalSuppressions.cs" />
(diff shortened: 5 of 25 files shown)


src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -21,8 +21,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
     public class DocumentDBConfiguration : IExtensionConfigProvider
     {
         internal const string AzureWebJobsDocumentDBConnectionStringName = "AzureWebJobsDocumentDBConnectionString";
+        // We separate the Trigger connection string in case of a mix of Trigger & output Attribute, otherwise it would generate an infinite loop
```
ealsur: If you set the DocumentDBTrigger to have the same connection string as the DocumentDB binding, it creates an infinite loop. You write a document, the trigger fires, you work with the document that changed and use the output binding so save, since both point to the same service, it might trigger again (if both work on the same collection). That happens if I reuse the same setting as the output binding on the trigger.

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -152,6 +176,30 @@ internal string ResolveConnectionString(string attributeConnectionString)
             return _defaultConnectionString;
         }
 
+        internal string ResolveTriggerConnectionString()
```
ealsur: I could fall back to the `ConnectionString`, I just want to avoid people accidentally creating an infinite loop and not realizing why.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerLeaseOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
+{
+    using System;
+    using Microsoft.Azure.Documents.ChangeFeedProcessor;
+
+    /// <summary>
+    /// Configuration options for the Leases used by the trigger mechanism
+    /// </summary>
+    public class DocumentDBTriggerLeaseOptions
```
ealsur: I tried to expose it directly. The effect was that it needed to download the ChangeFeedProcessor Nuget package also on the consumer project. If a user is creating a Webjob, they would need to also download that other Nuget to customize. I felt like it was another dependency. But I can get rid of it and let the user take the dependency hit.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerLeaseOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+        public DocumentDBTriggerLeaseOptionsCheckpointFrequency CheckpointFrequency { get; set; }
+
+        internal ChangeFeedHostOptions ToChangeFeedHostOptions()
+        {
+            ChangeFeedHostOptions options = new ChangeFeedHostOptions()
+            {
+                LeaseAcquireInterval = this.LeaseAcquireInterval,
+                FeedPollDelay = this.FeedPollDelay,
+                LeaseExpirationInterval = this.LeaseExpirationInterval,
+                LeaseRenewInterval = this.LeaseRenewInterval
```
ealsur: Sorry! I'll fix it, StyleCop didn't complain but I'll change it nonetheless.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerConstants.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
+{
+    internal static class DocumentDBTriggerConstants
+    {
+        public const string DefaultLeaseCollectionName = "leases";
+
+        public const string TriggerName = "DocumentDBTrigger";
+
+        public const string TriggerDescription = "New changes on collection {0} at {1}";
```
ealsur: Totally doable, will push it.

ealsur: I tried to implement it, but I don't seem to be able to read the data within the `ToParameterDescriptor` implementation.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerBinding.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private static IReadOnlyDictionary<string, object> GetBindingData(IReadOnlyList<Document> value)
+        {
+            Dictionary<string, object> bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)
+            {
+                { "Documents", value }
+            };
+            return bindingData;
+        }
```
ealsur: They went away on my last commit that added support for `JArray`.

ealsur:
@brettsam Pushed support for `JArray` binding, that should help with JS and other languages 😄

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
brettsam: Ok, that makes sense. We (kind of) have the same issue with other triggers/bindings (like Queue -> Queue, for example), but I think it's a little more obvious here b/c you're likely using the same Collection. I was just trying to keep the configuration a little smaller, if possible. Let me play with it and see if anything jumps out.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerLeaseOptions.cs
brettsam: It should be fine to expose that -- we do something similar for other bindings (we expose `EventProcessorOptions` and `PartitionManagerOptions` in Event Hubs, for example). It'd reduce the wrapper code a little.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerConstants.cs
brettsam: Okay, that was just a little "nice-to-have". We can always revisit later if we figure out a way.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerLeaseOptions.cs
ealsur: Done in the last commit.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerAttribute.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <summary>
+        /// Connection string for the service containing the lease collection
+        /// </summary>
+        [AppSetting]
+        public string LeaseConnectionString { get; private set; }
+
+        /// <summary>
+        /// Name of the lease collection
+        /// </summary>
+        public string LeaseCollection { get; private set; }
```
brettsam: `LeaseCollectionName` and `[AutoResolve]`

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerAttribute.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        /// <summary>
+        /// Name of the database containing the lease collection
+        /// </summary>
+        public string LeaseDatabase { get; private set; }
```
brettsam: `LeaseDatabaseName` and `[AutoResolve]`

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -40,6 +46,21 @@ public DocumentDBConfiguration()
...
 
+        /// <summary>
+        /// Gets or sets the DocumentDB connection string for the monitored service.
+        /// </summary>
+        public string TriggerConnectionString { get; set; }
+
+        /// <summary>
+        /// Optional. Gets or sets the DocumentDB connection string for the leases service.
+        /// </summary>
+        public string LeasesConnectionString { get; set; }
```
brettsam: I'm curious what @MikeStall thinks, but I think we can probably get rid of the `Trigger` and `Lease` connection string properties here and have the most common scenario work like:

- You use the config.ConnectionString value (which falls back to the environment variable) as the fallback for any missing connection string.
- If you want to override any of these in a specific function, you use the `triggerAttribute.ConnectionString` or `triggerAttribute.LeaseConnectionString` to do so.

That keeps the concepts a little cleaner in the configuration and still allows you to achieve everything you need to. Yes, there's still a chance that you accidentally wire up a loop, but that's true of a lot of our bindings that have triggers.

Plus, if we leave them off for now, we can always add the properties in the future if we decide we need them (while taking them away is harder).

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerAttribute.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+    using System;
+    using Microsoft.Azure.WebJobs.Extensions.DocumentDB;
+
+    /// <summary>
+    /// Defines the [DocumentDBTrigger] attribute
+    /// </summary>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1019:DefineAccessorsForAttributeArguments")]
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class DocumentDBTriggerAttribute : Attribute
```
ealsur: Just curious, without the multiple constructors, how would you specify the other parameters in a Functions signature?

For example:

`public static void CosmosSample([DocumentDBTrigger("connectionString", "db", "collection", "leaseConnectionString", "db", "collection")] JArray docs)`

How would I pass the lease data if I take away the constructor that receives it?

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
ealsur: Refactored based on your comments.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerAttribute.cs
brettsam: Make the `ConnectionString`, `LeaseConnectionString`, etc, have public setters, then we'll use that. An example from WebJobs here: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/ExtensionsSample/Samples/DocumentDBSamples.cs#L52 (note that `Id` isn't a part of the constructor).

In Functions, we'll have to write some 'glue' code that creates the attribute based on the function.json; we'll use the properties there as well.

ealsur: Done in the last commit.

src/ExtensionsSample/Samples/DocumentDBTriggerSamples.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        // Sample implementation of the CosmosDBTrigger that listens for changes in a collection.
+        // The trigger uses an auxiliary collection for leases for multiple partitions.
+        public static void Listen(
+            [DocumentDBTrigger("ItemDb", "ItemCollection", LeaseDatabaseName = "ItemDb", LeaseCollectionName = "Leases")] IReadOnlyList<Document> modifiedDocuments,
+            TraceWriter log)
+        {
+            foreach (Document modifiedDocument in modifiedDocuments)
+            {
+                log.Info(modifiedDocument.Id);
```
brettsam: Is there any way to know whether the document was added or modified?

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerListener.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public void Dispose()
+        {
+            //Nothing to dispose
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            return host.RegisterObserverFactoryAsync(this);
```
brettsam: A couple of clarifications now that I'm thinking through this (we may have discussed this already):

1. Will there only ever be one instance that is hooked up to this trigger? I can't remember what we said -- or was it one per partition?
2. Is there a maximum per-batch document count that you can configure?
3. Are these observers executed synchronously? For (a silly) example, if I have `Task.Delay(60000)` in my function code so it takes a while for me to execute each batch -- will my observer be called again on a separate thread while the previous one is waiting? Or will it wait until the previous is done?

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerAttribute.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            if (string.IsNullOrWhiteSpace(monitorCollectionDatabase))
+            {
+                throw new ArgumentException("Missing information for the collection to monitor", "monitorCollectionDatabase");
+            }
+
+            CollectionName = monitorCollection;
+            DatabaseName = monitorCollectionDatabase;
+            LeaseCollectionName = DocumentDBTriggerConstants.DefaultLeaseCollectionName;
```
ealsur: No, they have to be stored in a different collection always. And it's recommended that it's a different service if possible. That's the reason of the second connection string. Using it is the recommended scenario.

src/ExtensionsSample/Samples/DocumentDBTriggerSamples.cs
ealsur: No, according to the [docs](https://docs.microsoft.com/en-us/azure/cosmos-db/change-feed#how-change-feed-works-in-azure-cosmos-db) what it gets published are the documents that are either inserted or modified. Deletions are not published. It contains just the document, not the reason of why it was published in the feed.

src/WebJobs.Extensions.DocumentDB/Trigger/DocumentDBTriggerListener.cs
ealsur: 1. Partitions are handled dynamically, that's the reason for the Leases collection. If one instance of the Host is active, it will monitor all partitions, as more instances of the host are created (by multiplying instances in an App Service for example), partition ranges start to be distributed and leased. 
2. Yes, it's defined on the `leaseOptions` parameter of the constructor and exposed in the `DocumentDBConfiguration`. You can configure other things, like lease duration.
3. I'm not 100% sure of that scenario, from my tests, the processor waits until one batch is finished although there are configurable times in the `leaseOptions`. Since it's using a ContinuationToken to keep track of which was the last batch processed, you never lose changes if you are slow processing them, you just will get notified slower of the next batch.

brettsam:
FYI -- this will address #160

PR #258, opened 2017-07-31
Fixups to ExecutionContext

MikeStall:
Add FunctionAppDirectory path, set to  something determined by the host. Script runtime will set this to ScriptConfig.RootScriptPath

Also fix ExecutionContext.FunctionName to properly honor [FunctionName] attribute.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Core/Bindings/ExecutionContextBindingProvider.cs
+++ b/src/WebJobs.Extensions/Extensions/Core/Bindings/ExecutionContextBindingProvider.cs
@@ -2,10 +2,12 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Protocols;
+using static Microsoft.Azure.WebJobs.CoreJobHostConfigurationExtensions;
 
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions/Extensions/Core/Bindings/ExecutionContextBindingProvider.cs
```diff
@@ -60,17 +71,24 @@ public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
...
             {
-                return new ExecutionContext
+                var result = new ExecutionContext
                 {
-                    InvocationId = invocationId,
-                    FunctionName = _parameter.Member.Name,
-                    FunctionDirectory = Environment.CurrentDirectory
+                    InvocationId = context.FunctionInstanceId,
+                    FunctionName = context.FunctionContext.MethodName,
+                    FunctionDirectory = Environment.CurrentDirectory,
```
davidebbo: Seems quirky to rely on Current Dir, as that's fairly random in IIS scenarios (e.g. I think it's system32 folder). Though I take it that in real scenarios, the code below that overrides it will always run, making this one moot?

MikeStall: Correct.  
We have 2 scenarios: 
1. SDK-only outside of script. Existing behavior here is Env.CurDir. Not sure this is actually useful at all; but I'm leaving that behavior untouched. 
2. Running inside of script. In that case, app dir will always be set (subsequent PR coming), and that will override env.curdir to provide the values we discussed.

PR #260, opened 2017-08-02
Added support for Direct Send in Notification Hubs

tomasmcguinness:
This pull request adds a new NotificationHubDirectSendAttribute which allows a connection to an Azure Notification Hub to send notifications directly to a specific device. A new type called DirectNotification has been added that holds both a notification and a DeviceHandle.


tomasmcguinness:
Rookie mistake. Forgot to test existing unit tests and to add more.

PR #262, opened 2017-08-03
EasyAuth Support

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk-script/issues/33

I'm still in the process of adding some more tests.


src/WebJobs.Extensions.Http/Bindings/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+            {
+                if (context == null)
+                {
+                    throw new ArgumentNullException("context");
+                }
+
+                var identity = GetPrimaryIdentity(ClaimsPrincipal.Current);
```
mathewc: @fabiocav @cgillum  Any issues with accessing ClaimsPrincipal.Current on .NET Core? What is the recommended way for users/us to get the EasyAuth identity on .NET Core?

cgillum: Yes, this won't work in .NET Core because .NET Core doesn't play nice with IIS modules.  You would need to write some logic which manually creates the identity from the HTTP request headers.

cgillum: To get the claims, I'll need to update Easy Auth to flow them as a new HTTP request header.  I believe we had some conversation about this before, but I think you had some pushback and don't remember what the conclusion was. :(

mathewc:
Needs to be rebased and modified after .NET Core migration. Will resume this work in the future.

PR #267, opened 2017-08-11
Migrate Event Hub Extension to .Net Core

mamaso:
resolves https://github.com/Azure/azure-webjobs-sdk/issues/1291

Added code verbatim from Microsoft.Azure.WebJobs.ServiceBus in https://github.com/Azure/azure-webjobs-sdk-extensions/commit/0d0fdecae4198301444f8ff7bdcb5b204110505f

Changes to build on new event hubs nugets added in https://github.com/Azure/azure-webjobs-sdk-extensions/commit/92c82d821fd91cb04dbf8e1f6c34a4ae3acee150

The only major api change is that `PartitionManagerOptions` is no longer a thing, which doesn't have a large effect on functions.

We may want to take the opportunity to make `EventHubConfiguration` behave more like most extension configs, fix service bus duplicate namespace issue, etc. I think we could clean up pieces like `AddEventHubClient` and `AddEventHubSender`. @MikeStall any opinions on this?

Also, may consider moving this to multi-target in order to support webjobs 2.x.

### DIFF
--- a/WebJobs.Extensions.sln
+++ b/WebJobs.Extensions.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26724.1
+VisualStudioVersion = 15.0.26730.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{1118E0AD-83C8-4B67-859B-FCDEF54338C1}"
 EndProject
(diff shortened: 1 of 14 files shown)


WebJobs.Extensions.sln
```diff
@@ -22,12 +22,14 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ExtensionsSample", "src\Ext
...
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions.Tests", "test\WebJobs.Extensions.Tests\WebJobs.Extensions.Tests.csproj", "{1729561F-D598-4208-9154-3A5497523D54}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sample.Extension", "src\Sample.Extension\Sample.Extension.csproj", "{55EDABA0-8D31-43C9-AB3C-2A40D0EFF28A}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sample.Extension", "src\Sample.Extension\Sample.Extension.csproj", "{55EDABA0-8D31-43C9-AB3C-2A40D0EFF28A}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions", "src\WebJobs.Extensions\WebJobs.Extensions.csproj", "{7719584E-9863-405A-9D26-2CC02AFC2427}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions.Http", "src\WebJobs.Extensions.Http\WebJobs.Extensions.Http.csproj", "{47B3ECC3-DDDB-4C80-8A13-E6279A2EB835}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions.EventHubs", "src\WebJobs.Extensions.EventHubs\WebJobs.Extensions.EventHubs.csproj", "{37B71EB7-54DF-45FC-B25C-C80A797BB765}"
```
MikeStall: >"WebJobs.Extensions.EventHubs", [](start = 52, length = 31)

I'm concerned that:
- this is getting separate from ServiceBus
- this is moved from the SDK repro to this repro.  
these should be explicit choices.

mamaso: Bullet 1, I think it makes sense judging as event hubs is its own nuget, allowing for more granularity.

Bullet 2, I don't have an opinion. Moved here so it was easy to consume new extensions

mamaso:
moved to https://github.com/Azure/azure-webjobs-sdk/pull/1306

PR #269, opened 2017-08-14
Fixing CosmosDBTrigger ConnectionStringSetting binding

ealsur:
CosmosDBTrigger was not correctly reading ConnectionStringSetting from AppSettings and was treating it as a normal string.

### DIFF
--- a/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
             IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
             extensions.RegisterBindingRules<DocumentDBAttribute>(ValidateConnection, nameResolver, outputProvider, clientProvider, jArrayProvider, enumerableProvider, inputProvider);
 
-            context.Config.RegisterBindingExtensions(new CosmosDBTriggerAttributeBindingProvider(ResolveConnectionString(null), ResolveConnectionString(null), LeaseOptions));
+            context.Config.RegisterBindingExtensions(new CosmosDBTriggerAttributeBindingProvider(nameResolver, LeaseOptions));
         }
 
         internal static void ValidateInputBinding(DocumentDBAttribute attribute, Type parameterType)
--- a/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
@@ -13,14 +13,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 
     internal class CosmosDBTriggerAttributeBindingProvider : ITriggerBindingProvider
     {
-        private readonly string _monitorConnectionString;
-        private readonly string _leasesConnectionString;
         private readonly ChangeFeedHostOptions _leasesOptions;
+        private readonly INameResolver _nameResolver;
+        private string _monitorConnectionString;
+        private string _leasesConnectionString;
 
-        public CosmosDBTriggerAttributeBindingProvider(string monitorConnectionString, string leasesConnectionString, ChangeFeedHostOptions leasesOptions = null)
+        public CosmosDBTriggerAttributeBindingProvider(INameResolver nameResolver, ChangeFeedHostOptions leasesOptions = null)
         {
-            _monitorConnectionString = monitorConnectionString;
-            _leasesConnectionString = leasesConnectionString;
+            _nameResolver = nameResolver;
             _leasesOptions = leasesOptions ?? new ChangeFeedHostOptions();
         }
 
@@ -38,7 +38,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
             {
                 return Task.FromResult<ITriggerBinding>(null);
             }
-          
+
+            _monitorConnectionString = _nameResolver.Resolve(DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName);
+            _leasesConnectionString = _nameResolver.Resolve(DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName);
+
             DocumentCollectionInfo documentCollectionLocation;
             DocumentCollectionInfo leaseCollectionLocation;
             ChangeFeedHostOptions leaseHostOptions = ResolveLeaseOptions(attribute);
@@ -91,27 +94,55 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 
         private string ResolveAttributeConnectionString(CosmosDBTriggerAttribute attribute)
         {
-            if (string.IsNullOrEmpty(_monitorConnectionString) &&
-                string.IsNullOrEmpty(attribute.ConnectionStringSetting))
+            string connectionString = this._monitorConnectionString;
+            if (!TryReadFromSettings(attribute.ConnectionStringSetting, connectionString, out connectionString))
             {
                 throw new InvalidOperationException(
                     string.Format(CultureInfo.CurrentCulture,
-                    "The DocumentDBTrigger connection string must be set either via a '{0}' app setting, via the DocumentDBTriggerAttribute.ConnectionString property or via DocumentDBConfiguration.TriggerConnectionString.",
+                    "The CosmosDBTriggerAttribute.ConnectionStringSetting '{0}' property specified does not exist in the Application Settings.",
+                    attribute.ConnectionStringSetting));
+            }
+
+            if (string.IsNullOrEmpty(connectionString))
+            {
+                throw new InvalidOperationException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    "The CosmosDBTrigger connection string must be set either via a '{0}' app setting, via the CosmosDBTriggerAttribute.ConnectionStringSetting property or via DocumentDBConfiguration.ConnectionString.",
                     DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName));
             }
 
-            return attribute.ConnectionStringSetting ?? _monitorConnectionString;
+            return connectionString;
         }
 
         private string ResolveAttributeLeasesConnectionString(CosmosDBTriggerAttribute attribute, string triggerConnectionString)
         {
             // If the leases connection is not specified, it connects to the monitored service
-            return attribute.LeaseConnectionStringSetting ?? _leasesConnectionString ?? triggerConnectionString;
+            string connectionString = string.IsNullOrEmpty(this._leasesConnectionString) ? triggerConnectionString : this._leasesConnectionString;
+            if (!TryReadFromSettings(attribute.LeaseConnectionStringSetting, connectionString, out connectionString))
+            {
+                throw new InvalidOperationException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    "The CosmosDBTriggerAttribute.LeaseConnectionStringSetting '{0}' property specified does not exist in the Application Settings.",
+                    attribute.ConnectionStringSetting));
+            }
+
+            return connectionString;
         }
 
         private ChangeFeedHostOptions ResolveLeaseOptions(CosmosDBTriggerAttribute attribute)
         {
             return attribute.LeaseOptions ?? _leasesOptions;
+        }
+
+        private bool TryReadFromSettings(string settingsKey, string defaultValue, out string settingsValue)
+        {
+            if (string.IsNullOrEmpty(settingsKey))
+            {
+                settingsValue = defaultValue;
+                return true;
+            }
+            settingsValue = _nameResolver.Resolve(settingsKey);
+            return !string.IsNullOrEmpty(settingsValue);
         }
     }
 }
--- a/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerBinding.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerBinding.cs
@@ -36,6 +36,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
         /// </summary>
         public Type TriggerValueType => typeof(IReadOnlyList<Document>);
 
+        internal DocumentCollectionInfo DocumentCollectionLocation => _documentCollectionLocation;
+
+        internal DocumentCollectionInfo LeaseCollectionLocation => _leaseCollectionLocation;
+
         public IReadOnlyDictionary<string, Type> BindingDataContract
         {
             get { return _bindingDataProvider.Contract; }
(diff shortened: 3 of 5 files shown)


src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -91,27 +92,55 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
         }
+
+        private static bool TryReadFromSettings(string settingsKey, string defaultValue, out string settingsValue)
+        {
+            if (string.IsNullOrEmpty(settingsKey))
+            {
+                settingsValue = defaultValue;
+                return true;
+            }
+            settingsValue = ConfigurationManager.AppSettings[settingsKey];
```
brettsam: Instead of reading from `ConfigurationManager` directly, use `INameResolver.Resolve(settingsKey)`.  The DefaultNameResolver pulls from App Settings, then falls back to Environment variables. It also makes it easier to test as you can pass in your own test `INameResolver` without needing app settings or environment variables set up.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -91,27 +92,55 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+            string connectionString = this._monitorConnectionString;
+            if (!TryReadFromSettings(attribute.ConnectionStringSetting, connectionString, out connectionString))
+            {
+                throw new InvalidOperationException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    "The CosmosDBTriggerAttribute.ConnectionStringSetting '{0}' property specified does not exist in the Application Settings.",
+                    attribute.ConnectionStringSetting));
+            }
+
+            if (string.IsNullOrEmpty(connectionString))
```
brettsam: Will you ever get a null or empty here? Doesn't `TryReadFromSettings()` check that above? Also, this error message is more complete -- if you remove this `if` block completely, can you re-use the message here in the check above?

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -91,27 +92,55 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
             // If the leases connection is not specified, it connects to the monitored service
-            return attribute.LeaseConnectionStringSetting ?? _leasesConnectionString ?? triggerConnectionString;
+            string connectionString = string.IsNullOrEmpty(this._leasesConnectionString) ? triggerConnectionString : this._leasesConnectionString;
+            if (!TryReadFromSettings(attribute.LeaseConnectionStringSetting, connectionString, out connectionString))
+            {
+                throw new InvalidOperationException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    "The CosmosDBTriggerAttribute.LeaseConnectionStringSetting '{0}' property specified does not exist in the Application Settings.",
+                    attribute.ConnectionStringSetting));
+            }
```
brettsam: Can you use the more 'complete' error message here?

test/WebJobs.Extensions.Tests/app.config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <appSettings>
     <add key="NotificationHubsConnectionString" value="ConnectionString_FromAppSettings" />
+    <add key="CosmosDBConnectionString" value="AccountEndpoint=https://someEndpoint;AccountKey=someKey;" />
```
brettsam: When you switch to `INameResolver`, you won't need this.

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -98,31 +88,29 @@ public static IEnumerable<ParameterInfo[]> GetParameters()
...
                     new[] { GetFirstParameter(type, "Func3") },
-                    new[] { GetFirstParameter(type, "Func4") },
-                    new[] { GetFirstParameter(type, "Func5") },
-                    new[] { GetFirstParameter(type, "Func6") }
+                    new[] { GetFirstParameter(type, "Func4") }
                 };
             }
         }
 
         private static class ValidCosmosDBTriggerBindigs
```
brettsam: Can you add another valid binding -- where there's no `ConnectionStringSetting` used?  It should fall back to use the `AzureWebJobsDocumentDBConnectionString` app setting or environment variable.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: I was trying to catch a certain scenario: If the Attribute is set but it's pointing to an un-existing setting. The more complete message is a bit generic, as a user, I would thing "But I did specify the Connection setting".

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: Same as the other comment, it's meant to catch the specify scenario of an Attribute with an non-existant Setting in the environment.
The more complete message sounds aimed at a scenario where no Connection string was specified anywhere.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: Done

test/WebJobs.Extensions.Tests/app.config
ealsur: What if I wanted to test that the extraction of the value from the AppSettings worked specifically? Not just the defaulting of the environment variable.

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
ealsur: Done

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
brettsam: Ahh, okay, that does make sense then.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -91,27 +93,55 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
         }
+
+        private bool TryReadFromSettings(string settingsKey, string defaultValue, out string settingsValue)
+        {
+            if (string.IsNullOrEmpty(settingsKey))
+            {
+                settingsValue = defaultValue;
+                return true;
+            }
+            settingsValue = this._nameResolver.Resolve(settingsKey);
```
brettsam: nit: we don't use `this.` in our code :-) I searched for an easy way to catch this in StyleCop but couldn't find it, so we're a little mixed throughout the code base. I try to point it out when I see it, though.

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
ealsur: After doing it I discovered that the test failed in AppVeyor since there was no environment or setting available. Is that right?

test/WebJobs.Extensions.Tests/app.config
brettsam: You can add a test that uses the `DefaultNameResolver` as the name resolver. Mocking the `INameResolver` lets you make sure all of the fallbacks work without needing to change environment variables a bunch. I think for the most part, we trust that the `DefaultNameResolver` does what we expect (I see a lot of usage of `TestNameResolver`s in our tests) and try to ensure that we're using the `INameResolver` interface correctly.

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
brettsam: It's likely because you're not a part of our team directly -- we don't let non-team members have access to our connection strings for security reasons. That's another reason to mock out `INameResolver` -- you can use Moq to make sure that you're asking for the `AzureWebJobsDocumentDBConnectionString` as the fallback without actually needing the environment variable set up.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: Done 😄

test/WebJobs.Extensions.Tests/app.config
ealsur: Done 😄

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
ealsur: Done 😄  Finally understood how to use the mock

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -80,7 +80,7 @@ public void Initialize(ExtensionConfigContext context)
             IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
             extensions.RegisterBindingRules<DocumentDBAttribute>(ValidateConnection, nameResolver, outputProvider, clientProvider, jArrayProvider, enumerableProvider, inputProvider);
 
-            context.Config.RegisterBindingExtensions(new CosmosDBTriggerAttributeBindingProvider(ResolveConnectionString(null), ResolveConnectionString(null), LeaseOptions));
+            context.Config.RegisterBindingExtensions(new CosmosDBTriggerAttributeBindingProvider(nameResolver, ResolveConnectionString(null), ResolveConnectionString(null), LeaseOptions));
```
brettsam: Since ultimately what this is trying to do is make sure both the monitor and lease connection strings fall back to the default, I think we should bake that logic into the BindingProvider itself. It'll allow you to test the fallbacks more easily (see other comments). Can we remove the 2nd and 3rd parameters here? Then in TryCreateAsync you can do:

```
_monitorConnectionString = _nameResolver.Resolve(DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName);
_leasesConnectionString = _nameResolver.Resolve(DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName);
```

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerBinding.cs
```diff
@@ -81,4 +81,4 @@ public ParameterDescriptor ToParameterDescriptor()
             };
         }
     }
-}
+}
```
brettsam: In this class, you can make `DocumentCollectionLocation` and `LeaseCollectionLocation` `internal`, which will give you access to them from the tests. This will let you verify that things were set up as you expected.

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -21,18 +22,22 @@ public class CosmosDBTriggerAttributeBindingProviderTests
...
         [Theory]
         [MemberData("ValidCosmosDBTriggerParameters")]
         public async Task ValidParameters_Succeed(ParameterInfo parameter)
         {
-            CosmosDBTriggerAttributeBindingProvider provider = new CosmosDBTriggerAttributeBindingProvider("AccountEndpoint=https://someEndpoint;AccountKey=someKey;", string.Empty);
+            var nameResolver = new TestNameResolver();
+            nameResolver.Values[DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName] = "AccountEndpoint=https://someEndpoint;AccountKey=someKey;";
+            nameResolver.Values["CosmosDBConnectionString"] = "AccountEndpoint =https://someEndpoint;AccountKey=someKey;";
+
+            CosmosDBTriggerAttributeBindingProvider provider = new CosmosDBTriggerAttributeBindingProvider(nameResolver, "AccountEndpoint=https://someEndpoint;AccountKey=someKey;", string.Empty);
```
brettsam: In my previous comments, I suggested reducing this down to just take an `INameResolver` and the `ChangeFeedHostOptions` as well as exposing some extra properties on your Binding for testing. If you do that, now you should be able to:

1. Give each value in your `TestNameResolver` a unique connection string/ key value.
2. Look at the exposed `DocumentCollectionLocation` (and `LeaseCollectionLocation`) to make sure that they have the correct values.

You may have to split up your test since the validation will vary based on the scenario -- or you can add 'expected' parameters in your MemberData, which we do elsewhere as well.

But this way you'll know that you properly set the right fallback values based on what the attribute asked for.

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
ealsur: Makes sense, I didn't fully understand the nameResolver back when I created this 😄

test/WebJobs.Extensions.Tests/Extensions/DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
ealsur: Done 😄 Splitted tests into different scenarios

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerBinding.cs
ealsur: Done

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -16,11 +16,13 @@ internal class CosmosDBTriggerAttributeBindingProvider : ITriggerBindingProvider
...
+        private readonly INameResolver _nameResolver;
 
-        public CosmosDBTriggerAttributeBindingProvider(string monitorConnectionString, string leasesConnectionString, ChangeFeedHostOptions leasesOptions = null)
+        public CosmosDBTriggerAttributeBindingProvider(INameResolver nameResolver, ChangeFeedHostOptions leasesOptions = null)
         {
-            _monitorConnectionString = monitorConnectionString;
-            _leasesConnectionString = leasesConnectionString;
+            _nameResolver = nameResolver;
+            _monitorConnectionString = _nameResolver.Resolve(DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName);
+            _leasesConnectionString = _nameResolver.Resolve(DocumentDBConfiguration.AzureWebJobsDocumentDBConnectionStringName);
```
brettsam: Can you move these two to `TryCreateAsync`? I don't think it's guaranteed that this call won't throw, and we wouldn't want that to happen in a constructor.

ealsur: Sounds logical

Issue #273, opened 2017-08-18
Allow HTTP Trigger to bind to both request and poco 


MikeStall:
Support this: https://github.com/Azure/Azure-Functions/issues/401

ScriptBinding has some extra binding magic to support this. That should be in the http extension instead so that it works in all cases.  It currently does not work in direct  SDK cases, which includes Functions VS Tooling.

NicolasHumann:
+1 !!!
I really need this feature as well !

TheeJamesLee:
For reference:

Assuming that you are using the HttpRequestMessage to get headers/query parameters and/or for creating the response back. If so, this will achieve what you need:

`public static async Task<HttpResponseMessage> Run([HttpTrigger(...)]POCO req, IDictionary<string, string> headers, IDictionary<string, string> query)`

(then just build a response back rather then using the helpers off HttpRequestMessage)

The HttpTrigger will bind the Header and Query dictionaries to the 2 dictionaries in the signature above, which is what we do for what we need.

If we need anything more detailed then we just use the main http request object and in the function deserialize the object.

MikeStall:
We need to fix this (and document it so people know about it!). Script + CSX supports it.

MikeStall:
In the CSX path, this is handled here: https://github.com/Azure/azure-webjobs-sdk-script/blob/918b057707acfb842659c9dad3cef0193fae1330/src/WebJobs.Script.WebHost/WebScriptHostManager.cs#L354

MikeStall:
Resolved with https://github.com/Azure/azure-webjobs-sdk-extensions/pull/284

baio:
btw It still not working with core at the moment (

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="RegisterOrgFun.fs" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.0.0" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions" Version="3.0.0-beta4" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.8" />
    <PackageReference Include="System.Configuration" Version="2.0.5" />
    <PackageReference Include="FSharp.Core" Version="4.2.3" />
[...]

PR #284, opened 2017-08-29
Allow http binding to both poco and HttpRequestMessage.

MikeStall:
Fix https://github.com/Azure/azure-webjobs-sdk-extensions/issues/273

Script had special logic to allow this for CSX; but VS tooling and direct SDK users didn't have the same logic. Add the httprequest binding logic in the binding.


src/WebJobs.Extensions.Http/HttpDirectRequestBindingProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            ParameterInfo parameter = context.Parameter;
+            if (parameter.ParameterType == typeof(HttpRequestMessage))
+            {
+                // Not already claimed by another trigger?
+                if (parameter.GetCustomAttributes(false).Length == 0)
```
fabiocav: Should we be more specific on the type of attribute here (verify this is indeed a binding attribute so we still allow for other attributes)?

src/WebJobs.Extensions.Http/HttpDirectRequestBindingProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+                }
+            }
+
+            public Task<IValueProvider> BindAsync(BindingContext context)
+            {
+                if (context == null)
+                {
+                    throw new ArgumentNullException(nameof(context));
+                }
+                var request = context.BindingData["$request"];
```
fabiocav: Didn't we have constants with special binding data names somewhere? Would be good to move this there.

Issue #242, opened 2017-06-28
[.NET Core] - Migrate Twilio extension project


fabiocav:
Issue to track migration of the WebJobs.Extension.Twilio project to .NET Standard 2.0

 - [x] Migrate project
 - [x] Setup build / artifacts publishing

PR #287, opened 2017-08-30
Migrating Twilio to .NET Core

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk-extensions/issues/242.

There were a bunch of changes required to the object model of the extension, because Twilio changed their object model quite a bit. See https://www.twilio.com/docs/libraries/csharp/migrating-your-csharp-dot-net-application-twilio-sdk-4x-5x.

### DIFF
--- a/src/ExtensionsSample/Samples/TwilioSamples.cs
+++ b/src/ExtensionsSample/Samples/TwilioSamples.cs
@@ -1,0 +1,102 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs;
+using Newtonsoft.Json.Linq;
+using Twilio.Rest.Api.V2010.Account;
+using Twilio.Types;
+
+namespace ExtensionsSample.Samples
+{
+    // To use the TwilioSamples:
+    // 1. Configure your Twilio Account SID via the 'AzureWebJobsTwilioAccountSid' App Setting or Environment variable
+    // 2. Configure your Twilio Auth Token via the 'AzureWebJobsTwilioAuthToken' App Setting or Environment variable
+    // 3. Add typeof(TwilioSamples) to the SamplesTypeLocator in Program.cs
+    public static class TwilioSamples
+    {
+        /// <summary>
+        /// Demonstrates declaratively SMS message properties with parameter binding
+        /// to message properties.
+        /// </summary>
+        public static void ProcessOrder_Declarative(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [TwilioSms(
+                From = "{StorePhoneNumber}",
+                Body = "{CustomerName}, we've received your order ({OrderId}) and have begun processing it!")]
+            out CreateMessageOptions messageOptions)
+        {
+            // You can set additional message properties here
+            messageOptions = new CreateMessageOptions(new PhoneNumber(order.CustomerPhoneNumber));
+        }
+
+        /// <summary>
+        /// Demonstrates imperatively setting SMS message properties inline in the function.
+        /// </summary>
+        [Disable]
+        public static void ProcessOrder_Imperative(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [TwilioSms] out CreateMessageOptions messageOptions)
+        {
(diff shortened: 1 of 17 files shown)


src/ExtensionsSample/Samples/TwilioSamples.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary>
+        /// Demonstrates declaratively SMS message properties with parameter binding
+        /// to message properties.
+        /// </summary>
+        public static void ProcessOrder_Declarative(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [TwilioSms(
+                From = "{StorePhoneNumber}",
+                Body = "{CustomerName}, we've received your order ({OrderId}) and have begun processing it!")]
+            out CreateMessageOptions messageOptions)
```
mathewc: Main change is that we now use the mutable CreateMessageOptions, whereas it was SMSMessage before. They overhauled their message APIs. The replacement for SMSMessage is MessageResource, and it is no longer mutable.

src/ExtensionsSample/Samples/TwilioSamples.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            // You can set additional message properties here
+            messageOptions = new CreateMessageOptions(new PhoneNumber(order.CustomerPhoneNumber));
```
mathewc: The To property is read only (its a ctor param), so this example had to be changed. The To property can't be defaulted later.

src/Sample.Extension/SampleJobHostConfigurationExtensions.cs
```diff
@@ -33,7 +35,19 @@ public void Initialize(ExtensionConfigContext context)
                 context.Config.RegisterBindingExtensions(
                     new SampleAttributeBindingProvider(),
                     new SampleTriggerAttributeBindingProvider());
+
+                IConverterManager converterManager = context.Config.GetService<IConverterManager>();
```
mathewc: Fixing the broken sample app. It doesn't run currently in either branch.

src/WebJobs.Extensions.Twilio/Config/TwilioSmsConfiguration.cs
```diff
@@ -0,0 +1,157 @@
...
+                Client = client,
+                Body = Utility.FirstOrDefault(attribute.Body, Body),
+                From = Utility.FirstOrDefault(attribute.From, From),
+                To = Utility.FirstOrDefault(attribute.To, To)
+            };
+
+            return context;
+        }
+
+        internal static CreateMessageOptions CreateMessageOptions(JObject messageOptions)
```
mathewc: I'm doing manual deserialization here. I couldn't get Json.NET to deserialize directly into these types for me. I tried lots of different ways, specified the converter array, etc. Open to any suggestions...

src/WebJobs.Extensions.Twilio/Config/TwilioSmsConfiguration.cs
```diff
@@ -0,0 +1,157 @@
...
+        public void Initialize(ExtensionConfigContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            INameResolver nameResolver = context.Config.GetService<INameResolver>();
+
+            _defaultAccountSid = nameResolver.Resolve(AzureWebJobsTwilioAccountSidKeyName);
```
mamaso: instead of using this flow, can we use `[AppSetting(Default="AzureWebJobsTwilioAccountSid")]` instead?

src/WebJobs.Extensions.Twilio/Config/TwilioSmsConfiguration.cs
mamaso: Does `JsonConvert.PopulateObject` work? Construct CreateMessageOptions via phone number constructor & populate it.

This is the popular answer which looks like manual deserialization to me: https://stackoverflow.com/a/28155770/5915331

mathewc: Yes, that's the first thing I tried. Their custom types for PhoneNumber, enums (that aren't actually enums) were failing to deserialize, even with custom converters added.

Issue #243, opened 2017-06-28
[.NET Core] - Migrate SendGrid extension project


fabiocav:
Issue to track migration of the WebJobs.Extension.SendGrid project to .NET Standard 2.0

 - [x] Migrate project
 - [x] Setup build / artifacts publishing

xt0rted:
I've been working with this and the SendGrid library (the newer version that changed the api yet again, ref: #191). Once this is ready to be worked on, assuming I have the time, I could take a shot at doing this task. It'd be useful preparation for migrating my own projects over which depend on this.

PR #288, opened 2017-08-31
Migrating SendGrid extension to .NET core (#243)

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk-extensions/issues/243

### DIFF
--- a/src/ExtensionsSample/Samples/SendGridSamples.cs
+++ b/src/ExtensionsSample/Samples/SendGridSamples.cs
@@ -1,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+using Newtonsoft.Json.Linq;
+using SendGrid.Helpers.Mail;
+
+namespace ExtensionsSample
+{
+    // To use the SendGridSamples:
+    // 1. Configure your SendGrid API Key via the 'AzureWebJobsSendGridApiKey' App Setting or Environment variable
+    // 4. Add typeof(SendGridSamples) to the SamplesTypeLocator in Program.cs
+    public static class SendGridSamples
+    {
+        /// <summary>
+        /// Demonstrates declaratively defining all email properties with parameter binding
+        /// to message properties.
+        /// </summary>
+        public static void ProcessOrder_Declarative(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [SendGrid(
+                To = "{CustomerEmail}",
+                Subject = "Thanks for your order (#{OrderId})!",
+                Text = "{CustomerName}, we've received your order ({OrderId}) and have begun processing it!")]
+            out SendGridMessage message)
+        {
+            // You can set additional message properties here
+            message = new SendGridMessage();
+            message.AddHeader("MyHeader", "MyValue");
+        }
+
+        /// <summary>
+        /// Demonstrates imperatively setting email properties inline in the function.
+        /// </summary>
+        [Disable]
+        public static void ProcessOrder_Imperative(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [SendGrid] out SendGridMessage message)
+        {
+            message = new SendGridMessage
+            {
+                Subject = $"Thanks for your order (#{order.OrderId})!"
+            };
+            message.AddContent("text/plain", $"{order.CustomerName}, we've received your order ({order.OrderId}) and have begun processing it!");
+            message.AddTo(order.CustomerEmail);
+        }
+
+        /// <summary>
+        /// Demonstrates the JObject binding.
+        /// </summary>  
+        [Disable]
+        public static void ProcessOrder_JObject(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [SendGrid] out JObject message)
+        {
+            message = JObject.Parse(GetEmailJson(order));
+        }
+
+        /// <summary>
+        /// Demonstrates the string binding.
+        /// </summary>
+        [Disable]
+        public static void ProcessOrder_String(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [SendGrid] out string message)
+        {
+            message = GetEmailJson(order);
+        }
+
+        /// <summary>
+        /// Demonstrates the IAsyncCollector binding. This works with JObject
+        /// or SendGridMessage. Using IAsyncCollector is also a way to conditionally
+        /// send messages from a function. If your function is not async, you can
+        /// use ICollector.
+        /// </summary>
+        [Disable]
+        public static async Task ProcessOrder_JObjectAsyncCollector(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [SendGrid] IAsyncCollector<JObject> messages)
+        {
+            JObject message = JObject.Parse(GetEmailJson(order));
+            await messages.AddAsync(message);
+        }
+
+        private static string GetEmailJson(Order order)
+        {
+            // Mail reference can be found at: https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html
+            return $@"{{
+              'personalizations': [
+                {{
+                  'to': [
+                    {{
+                      'email': '{order.CustomerEmail}'
+                    }}
+                  ]                  
+                }}
+              ],
+              'subject': 'Thanks for your order (#{order.OrderId})!',
+              'content': [
+                {{
+                  'type': 'text/plain',
+                  'value': '{order.CustomerName}, we\'ve received your order ({order.OrderId}) and have begun processing it!'
+                }}
+              ]
+            }}";
+        }
+    }
+}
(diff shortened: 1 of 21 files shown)


src/ExtensionsSample/Samples/SendGridSamples.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// Demonstrates declaratively defining all email properties with parameter binding
+        /// to message properties.
+        /// </summary>
+        public static void ProcessOrder_Declarative(
+            [QueueTrigger(@"samples-orders")] Order order,
+            [SendGrid(
+                To = "{CustomerEmail}",
+                Subject = "Thanks for your order (#{OrderId})!",
+                Text = "{CustomerName}, we've received your order ({OrderId}) and have begun processing it!")]
+            out SendGridMessage message)
```
mathewc: Main object model change was from type "Mail" to "SendGridMessage"

src/Sample.Extension/SampleJobHostConfigurationExtensions.cs
```diff
@@ -33,6 +35,18 @@ public void Initialize(ExtensionConfigContext context)
                 context.Config.RegisterBindingExtensions(
                     new SampleAttributeBindingProvider(),
                     new SampleTriggerAttributeBindingProvider());
+
+                IConverterManager converterManager = context.Config.GetService<IConverterManager>();
+                converterManager.AddConverter<CloudTable, Table<OpenType>, TableAttribute>(typeof(CustomTableBindingConverter<>));
+            }
+        }
+
+        private class CustomTableBindingConverter<T>
```
mathewc: Allows the samples to run. I made the same change in the Twilio migration review. Will reconcile.

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+using Newtonsoft.Json.Linq;
+using SendGrid.Helpers.Mail;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Bindings
+{
+    internal class SendGridHelpers
+    {
+        internal static EmailAddress Apply(EmailAddress current, string value)
+        {
+            EmailAddress mail;
```
mathewc: Lots of their previous types changed names. Email => EmailAddress, etc.

src/WebJobs.Extensions.SendGrid/Client/ISendGridClient.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using SendGrid;
+using SendGrid.Helpers.Mail;
+
+namespace Client
```
mathewc: I changed this namespace to avoid conflicts we were having with their SendGrid namespace and their SendGridClient. .NET core doesn't support extern alias, which would have been another way to resolve.

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        internal static void DefaultMessageProperties(SendGridMessage mail, SendGridConfiguration config, SendGridAttribute attribute)
+        {
+            // Apply message defaulting
+            if (mail.From == null)
+            {
+                if (!string.IsNullOrEmpty(attribute.From))
+                {
+                    EmailAddress from = null;
+                    if (!TryParseAddress(attribute.From, out from))
```
mamaso: nit `out EmailAddress from`

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+                    }
+                    mail.From = from;
+                }
+                else if (config.FromAddress != null)
+                {
+                    mail.From = config.FromAddress;
+                }
+            }
+
+            if (!IsToValid(mail))
```
mamaso: a lot of this looks like it could be simplified with something like the Twilio Utility.FirstOrDefault() that does all the null/empty coalescing

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
```diff
@@ -0,0 +1,153 @@
...
+            return item.Personalizations != null &&
+                item.Personalizations.Count > 0 &&
+                item.Personalizations.All(p => p.Tos != null && !p.Tos.Any(t => string.IsNullOrEmpty(t.Email)));
+        }
+
+        internal static SendGridConfiguration CreateConfiguration(JObject metadata)
+        {
+            SendGridConfiguration sendGridConfig = new SendGridConfiguration();
+
+            JObject configSection = (JObject)metadata.GetValue("sendGrid", StringComparison.OrdinalIgnoreCase);
```
mamaso: I think the ApplyConfig utility might help here - I think we should allow the properties `fromAddress` and `toAddress` so that there's not a disconnect between host.json and config values (while still supporting the old to and from values)

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
```diff
@@ -0,0 +1,126 @@
...
+using SendGrid.Helpers.Mail;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SendGrid
+{
+    /// <summary>
+    /// Defines the configuration options for the SendGrid binding.
+    /// </summary>
+    public class SendGridConfiguration : IExtensionConfigProvider
+    {
+        internal const string AzureWebJobsSendGridApiKeyName = "AzureWebJobsSendGridApiKey";
```
mamaso: same `[AppSetting(Default=...)]` comment as twilio

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+
+        private static string FirstOrDefault(params string[] values)
+        {
+            return values.FirstOrDefault(v => !string.IsNullOrEmpty(v));
+        }
+
+        // Schema for host.json 
+        private class ConfigMetadata
```
mamaso: think we should also support fromAddress and toAddress in json

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
mathewc: Why do you think that helps?

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
mathewc: Agreed. However my primary focus in these migrations is to structurally migrate w/o too much uneccessary refactoring. We can improve later :)

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
mamaso: it doesn't, especially in the context of how TryParseAddress is used (you could flip around the if statement to be cleaner but eh)
carry on

src/WebJobs.Extensions.SendGrid/Bindings/SendGridHelpers.cs
mamaso: Agreed, sounds good

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
mathewc: Good suggestion. I think I'll get the initial migrations checked in before refactoring things.

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
mathewc: You mean just add additional JsonProperty annotations for those alternate names?

mamaso:
@mathewc @MikeStall @brettsam @fabiocav 

I feel like these extension prs will be quite a bit easier to read if we can see the diff from full framework.

With that in mind, if the extension has been deleted from the repo it'll be easier if you have 2 commits 
1. add the old extension back to the repo
2. migrate the extension

Then we can scope reviews down to mostly commit 2

src/WebJobs.Extensions.SendGrid/Config/SendGridConfiguration.cs
mamaso: Yeah, I think that might do the trick

mathewc:
Well, we made an early decision to delete everything from the repo first, which put us on this path.

mamaso:
@mathewc yep, it'd be nice if we could restore the full history somehow

PR #290, opened 2017-09-07
Migrating DocumentDB Extension to .NETStandard and renaming CosmosDB

### LINKED ISSUE #240: [.NET Core] - Migrate DocumentDB extension project
Issue to track migration of the WebJobs.Extension.DocumentDB project to .NET Standard 2.0

 - [ ] Migrate project
 - [ ] Setup build / artifacts publishing

### LINKED ISSUE #280: Remove RetryAsync from DocumentDBUtility
Since DocDB 1.8, the client has handled retries internally. There's no need for this utility anymore and we should remove it and the related tests. We should sync with the DocDB team to find out if the defaults will work best for us.

brettsam:
Fixes #280.
Fixes #240.

Note that I removed the `CosmosDBTrigger` for now as there's currently no .NETStandard version of the ChangeFeedProcessor.


src/ExtensionsSample/Program.cs
```diff
@@ -54,19 +55,14 @@ public static void Main(string[] args)
...
-            host.Call(typeof(MiscellaneousSamples).GetMethod("ExecutionContext"));
-            host.Call(typeof(FileSamples).GetMethod("ReadWrite"));
-            host.Call(typeof(SampleSamples).GetMethod("Sample_BindToStream"));
-            host.Call(typeof(SampleSamples).GetMethod("Sample_BindToString"));
-            host.Call(typeof(TableSamples).GetMethod("CustomBinding"));
+            //host.Call(typeof(MiscellaneousSamples).GetMethod("ExecutionContext"));
+            //host.Call(typeof(FileSamples).GetMethod("ReadWrite"));
+            //host.Call(typeof(SampleSamples).GetMethod("Sample_BindToStream"));
+            //host.Call(typeof(SampleSamples).GetMethod("Sample_BindToString"));
+            //host.Call(typeof(TableSamples).GetMethod("CustomBinding"));
```
mamaso: why are these commented out?

src/WebJobs.Extensions.DocumentDB/Bindings/DocumentDBItemValueBinder.cs
```diff
@@ -51,8 +51,16 @@ public Type Type
...
 
-            Document document = await DocumentDBUtility.RetryAsync(() => _context.Service.ReadDocumentAsync(documentUri, options),
-                    _context.MaxThrottleRetries, codesToIgnore: HttpStatusCode.NotFound);
+            Document document = null;
+
+            try
+            {
+                document = await _context.Service.ReadDocumentAsync(documentUri, options);
+            }
+            catch (DocumentClientException ex) when (ex.StatusCode == HttpStatusCode.NotFound)
```
mamaso: hm, nifty sugar

src/WebJobs.Extensions.CosmosDB/Bindings/CosmosDBAsyncCollector.cs
```diff
@@ -10,13 +10,13 @@
 using Microsoft.Azure.Documents.Client;
 using Newtonsoft.Json.Linq;
 
-namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
+namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
```
mamaso: This is one that I think we should avoid - I know the name has changed, but unfortunately this will break customers. It would be nice if we could have the old namespace as well as the new cosmosdb namespace & classes

src/ExtensionsSample/Program.cs
brettsam: Oh that's just for my testing -- I usually change the loaded sample to only use the class that I'm testing and so these always fail while I'm I'm playing around. I usually back that out in the last second, but must have missed it this time.

src/WebJobs.Extensions.CosmosDB/Config/IDocumentDBServiceFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
fabiocav: Rename file to `ICosmosDBServiceFactory`

brettsam:
@fabiocav / @mamaso -- I've also ported over the 2 changes that I have ready for the `v2.x` branch in this PR: #292. Would you guys mind looking at either of these PRs and I'll consider them both approved?

brettsam:
I'm going to go ahead and get this in -- can address other comments as they come.

PR #304, opened 2017-09-25
NotificationHub name property (hubName) in binding cannot be mapped t…

alrod:
…o app setting. Fixes #1907.

### DIFF
--- a/src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
+++ b/src/WebJobs.Extensions.NotificationHubs/NotificationHubAttribute.cs
@@ -48,6 +48,7 @@ namespace Microsoft.Azure.WebJobs
         /// Optional. The Notification Hub Name to use, if different than the one specified in the
         /// <see cref="NotificationHubsConfiguration"/>.
         /// </summary>
+        [AutoResolve]
         public string HubName { get; set; }
 
         /// <summary>


src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -127,13 +128,13 @@ internal string ResolveHubName(string attributeHubName)
             // First, try the Attribute's string.
             if (!string.IsNullOrEmpty(attributeHubName))
             {
-                return attributeHubName;
+                return _nameResolver.ResolveWholeString(attributeHubName);
```
mamaso: Instead of using the name resolver directly, I wonder if applying the `[AppSetting]` or `[AutoResolve]` attribute to the HubName property in NotificationHubAttribute may make more sense.

AppSetting will resolve from app settings only, AutoResolve will resolve %% and {} within names to app settings and binding data respectively. In this case, I think that AppSetting makes the most sense.

https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs#L155

One potential issue is if `AppSetting` requires the property to exist - perhaps the attribute could have an 'optional' flag.

src/WebJobs.Extensions.NotificationHubs/Config/NotificationHubsConfiguration.cs
```diff
@@ -127,13 +128,13 @@ internal string ResolveHubName(string attributeHubName)
...
             }
 
             // Second, try the config's HubName
             if (!string.IsNullOrEmpty(HubName))
             {
-                return HubName;
+                return _nameResolver.ResolveWholeString(HubName);
```
mamaso: What happens if _nameResolver.ResolveWholeString fails to find an app setting to resolve? Does it return string.Empty?

We probably want to check that a result was found - if not, return the original HubName.

test/WebJobs.Extensions.Tests/Extensions/NotificationHubs/NotificationHubConfigurationTests.cs
```diff
@@ -97,12 +91,29 @@ public void Resolve_UsesDefault_Last()
             Assert.Equal("Default", connString);
         }
 
-        private NotificationHubsConfiguration InitializeConfig(string defaultConnStr)
+        [Fact]
```
mamaso: I'd like to see a few more tests - maybe setting HubName on attributes, both attribute & config, existing and non-existing app settings, %% signs and no percent signs.

Issue #309, opened 2017-10-04
TimerTrigger to support cron expressions of 5 parts


kevbite:
Is there any reason why the TimerTrigger does not support a 5 part cron expression (missing the seconds part)? Lots of the online cron expression builders only build the expressions with 5 parts and I've been caught out a few times by this.

fabiocav:
Resolved by #310

PR #310, opened 2017-10-04
Support 5 part cron expressions

kevbite:
This is related to this issue #309.

I've also altered the tests so they don't have to run in an en-us culture.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
@@ -47,11 +47,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
         internal static bool TryCreate(string cronExpression, out CronSchedule cronSchedule)
         {
             cronSchedule = null;
-            var isSixPart = cronExpression.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries).Length == 6;
+            
             CrontabSchedule.ParseOptions options = new CrontabSchedule.ParseOptions()
             {
-                IncludingSeconds = isSixPart
+                IncludingSeconds = HasSeconds(cronExpression)
             };
+
             CrontabSchedule crontabSchedule = CrontabSchedule.TryParse(cronExpression, options);
             if (crontabSchedule != null)
             {
@@ -59,6 +60,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
                 return true;
             }
             return false;
+        }
+
+        private static bool HasSeconds(string cronExpression)
+        {
+            return cronExpression.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries).Length != 5;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -60,9 +47,10 @@ public override string ToString()
         internal static bool TryCreate(string cronExpression, out CronSchedule cronSchedule)
         {
             cronSchedule = null;
+            var isSixPart = cronExpression.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries).Length == 6;
```
mamaso: A couple things:

Can we convert this to match 5 part cron expressions so that the default is 6 (in the event that the matching logic is off)

Also, I believe that the expressions support tabs and newlines in whitespace - can we change the match logic to be a regex with 5 parts and whitespace?

kevbite: Sure I'll update the PR

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -60,10 +47,12 @@ public override string ToString()
         internal static bool TryCreate(string cronExpression, out CronSchedule cronSchedule)
         {
             cronSchedule = null;
```
fabiocav: With the previous implementation, if `cronExpression` was `null`, that would just result in `TryCreate` returning `false` (not super clear, and a bit fragile as it just relied on the `TryParse` behavior), but with this change, it will result in a `null` reference exception when it hits the `Split` in the new `HasSeconds` method.

Can we update this to maintain the same behavior and return `false` when a `null` expression is passed in?

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/ScheduleMonitorTests.cs
```diff
@@ -17,21 +20,21 @@ public class ScheduleMonitorTests
...
+            _hourlySchedule = new CronSchedule(CrontabSchedule.Parse("0 * * * *"));
+            _halfHourlySchedule = new CronSchedule(CrontabSchedule.Parse("*/30 * * * *"));
+            _dailySchedule = new CronSchedule(CrontabSchedule.Parse("0 0 * * *"));
         }
 
         [Fact]
         public async Task CheckPastDue_NullStatus()
         {
-            DateTime now = DateTime.Parse("1/1/2017 9:35");
+            DateTime now = new DateTime(2017, 1, 1, 9, 35, 0);
```
fabiocav: 👍

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
kevbite: Updated to replicate behaviour and wrapped in tests.

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/CronScheduleTests.cs
```diff
@@ -53,8 +53,19 @@ public void GetNextOccurrence_ThreeDaySchedule_MultipleScheduleIterations()
...
         {
-            CronSchedule.TryCreate("0 59 11 * * 1-5", out CronSchedule schedule);
+            var result = CronSchedule.TryCreate("0 59 11 * * 1-5", out CronSchedule schedule);
+
+            Assert.True(result);
             Assert.Equal("Cron: '0 59 11 * * 1-5'", schedule.ToString());
         }
+
+        [Fact]
+        public void NullExpression_ReturnsFalseAndNullcronSchedule()
```
fabiocav: Thank you for the added test!

PR #314, opened 2017-10-16
Plumb [File] on SDK's new BindToInput rule.

MikeStall:
FileTrigger is not changed.
The key difference here is remove the file binding provider classses and using the rules. See FilesExtensionConfig.Initialize

This relies on https://github.com/Azure/azure-webjobs-sdk/pull/1387 from SDK.


src/WebJobs.Extensions/Extensions/Files/Config/FilesJobHostConfigurationExtensions.cs
```diff
@@ -57,16 +59,45 @@ public FilesExtensionConfig(FilesConfiguration filesConfig)
                 _filesConfig = filesConfig;
             }
 
+            private FileInfo GetFileInfo(FileAttribute attribute)
+            {
+                string boundFileName = attribute.Path;
+                string filePath = Path.Combine(_filesConfig.RootPath, boundFileName);
```
ConnorMcMahon: Should we be checking for null values on the attribute.Path value or the RootPath of the fileconfig?

src/WebJobs.Extensions/Extensions/Files/Config/FilesJobHostConfigurationExtensions.cs
```diff
@@ -57,16 +59,45 @@ public FilesExtensionConfig(FilesConfiguration filesConfig)
...
+                FileInfo fileInfo = new FileInfo(filePath);
+                return fileInfo;
+            }
+
+            public FileStream GetFileStream(FileAttribute attribute)
+            {
+                var fileInfo = GetFileInfo(attribute);
+                if ((attribute.Access == FileAccess.Read) && !fileInfo.Exists)
+                {
+                    return null;
```
ConnorMcMahon: Letting the FileInfo.Open() call below call an exception below may provide a more clear error message than returning a null value for the Stream.

ConnorMcMahon: Looking at your unit tests on the other PR, it appears we want to return null here since there isn't a reliable exception for the file not existing. Adding a comment here would help make that more clear.

MikeStall: Right, null is the contract for "does not exist"

src/WebJobs.Extensions/Extensions/Files/Config/FilesJobHostConfigurationExtensions.cs
```diff
@@ -57,16 +59,45 @@ public FilesExtensionConfig(FilesConfiguration filesConfig)
...
 
+                var rule = context.AddBindingRule<FileAttribute>();
+                rule.BindToInput<FileInfo>(this.GetFileInfo);
+                rule.BindToInput<FileStream>(this.GetFileStream);
+                rule.BindToStream(this, FileAccess.ReadWrite);
+
                 context.Config.RegisterBindingExtensions(
-                    new FileTriggerAttributeBindingProvider(_filesConfig, context.Trace),
-                    new FileAttributeBindingProvider(_filesConfig, context.Config.NameResolver));
+                    new FileTriggerAttributeBindingProvider(_filesConfig, context.Trace)
```
mathewc: Does this binding provider now handle trigger as well as non-trigger bindings (e.g. the input/output file bindings). If so, then it might make sense to remove "trigger" from its name.

MikeStall: No change to  triggers yet.  So we still need the TriggerBindingProviders. I had a separate idea that may improve this, but will tackle later.

MikeStall:
Superseded by https://github.com/Azure/azure-webjobs-sdk-extensions/pull/331

PR #305, opened 2017-09-27
Adding support for JSON string values in Cosmos DB Trigger binding

ealsur:
This change enables executing the Trigger with a test payload from the Functions UI.

PR #321, opened 2017-11-01
[Cosmos] adding string support for trigger to support JavaScript

brettsam:


brettsam:
@ealsur -- I had to change some things here for v2 JavaScript support. There may be some bug fixes in the core binding stuff to make this nicer, but right now JavaScript uses a 'string' parameter, which this didn't support. That also caused issues with the BindingContract b/c strings don't have one. Since we don't really use that, I decided to make it empty. We can always revisit later if there is a good use for binding data.

brettsam:
Going to merge this again to create a build to unblock others. Will still accept comments :-)

ealsur:
@brettsam I added this on a PR that was never merged :( https://github.com/Azure/azure-webjobs-sdk-extensions/pull/305

PR #324, opened 2017-11-02
[CosmosDB] fixing javascript output binding

brettsam:


brettsam:
Note: I'll leave this up for a bit but it may not be necessary as some new binding fixes come from the SDK.

PR #329, opened 2017-11-07
updating appveyor.yml to respect IncludeBuildNumberInVersion

brettsam:


brettsam:
@mathewc FYI -- I'm gonna commit this. I verified that it worked in both scenarios on AppVeyor.

PR #331, opened 2017-11-10
Update to latest extensibility interfaces

MikeStall:
Move away from deprecated BindingFactory.
User OpenType. 
Use Bind ToStream for [File]
Use converter rules for [FileTrigger]


src/ExtensionsSample/Program.cs
```diff
@@ -20,32 +20,21 @@ public static class Program
         public static void Main(string[] args)
         {
             JobHostConfiguration config = new JobHostConfiguration();
+
```
MikeStall: I'll revert these

src/ExtensionsSample/Samples/FileSamples.cs
```diff
@@ -14,19 +14,19 @@ namespace ExtensionsSample
...
-        // container then deleted.
-        public static async Task ImportFile(
-            [FileTrigger(@"import/{name}", "*.dat", autoDelete: true)] Stream file,
-            [Blob(@"processed/{name}")] CloudBlockBlob output,
-            string name,
-            TextWriter log)
-        {
-            await output.UploadFromStreamAsync(file);
-            file.Close();
+        //// When new files arrive in the "import" directory, they are uploaded to a blob
```
MikeStall: I'll revert these.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs
```diff
@@ -58,34 +60,30 @@ public void Initialize(ExtensionConfigContext context)
...
             _defaultConnectionString = nameResolver.Resolve(AzureWebJobsCosmosDBConnectionStringName);
 
-            BindingFactory factory = new BindingFactory(nameResolver, converterManager);
-
-            IBindingProvider outputProvider = factory.BindToCollector<CosmosDBAttribute, OpenType>(typeof(CosmosDBCollectorBuilder<>), this);
-
-            IBindingProvider outputProviderJObject = factory.BindToCollector<CosmosDBAttribute, JObject>(typeof(CosmosDBCollectorBuilder<>), this);
-
-            IBindingProvider clientProvider = factory.BindToInput<CosmosDBAttribute, DocumentClient>(new CosmosDBClientBuilder(this));
+            // Apply ValidateConnection to all on this rule.
```
MikeStall: No more BindingFactory! All rules based.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs
```diff
@@ -58,34 +60,30 @@ public void Initialize(ExtensionConfigContext context)
...
+            rule.BindToCollector<OpenType>(typeof(CosmosDBCollectorBuilder<>), this);
+            rule.BindToCollector<JObject>(typeof(CosmosDBCollectorBuilder<>), this);
 
-            IBindingProvider jArrayProvider = factory.BindToInput<CosmosDBAttribute, JArray>(typeof(CosmosDBJArrayBuilder), this);
+            rule.BindToInput<DocumentClient>(new CosmosDBClientBuilder(this));
+            rule.BindToInput<JArray>(typeof(CosmosDBJArrayBuilder), this);
 
-            IBindingProvider enumerableProvider = factory.BindToInput<CosmosDBAttribute, IEnumerable<OpenType>>(typeof(CosmosDBEnumerableBuilder<>), this);
-            enumerableProvider = factory.AddValidator<CosmosDBAttribute>(ValidateInputBinding, enumerableProvider);
+            rule.BindToInput<IEnumerable<OpenType>>(typeof(CosmosDBEnumerableBuilder<>), this).AddValidator(ValidateInputBinding);
```
MikeStall: Same semantics here... the validator runs *After* the rule is matched (this is important so that we don't attempt validation on other rules that would obviously fail). So in the fluent interface, it's Bind().AddValidate()

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
```diff
@@ -64,84 +67,40 @@ public void Initialize(ExtensionConfigContext context)
...
+            
+            var rule = context.AddBindingRule<MobileTableAttribute>();
+            rule.AddValidator(ValidateMobileAppUri);
 
-            BindingFactory factory = new BindingFactory(_nameResolver, converterManager);
+            rule.BindToCollector<OpenType>(typeof(MobileTableCollectorBuilder<>), this);
+            rule.BindToInput<IMobileServiceClient>(new MobileTableClientBuilder(this));
 
-            IBindingProvider outputProvider = factory.BindToCollector<MobileTableAttribute, OpenType>(typeof(MobileTableCollectorBuilder<>), this);
+            rule.WhenIsNotNull(nameof(MobileTableAttribute.TableName)).
```
MikeStall: Mobile has some pretty aggressive rules here that combine Attribute and parameter type (which I think . This is using the fluent declarative filtering - it detangles it and should be preserving the same semantics as the original. 

Note that  we no longer need per-rule predicates. We can use the OpenType system and just say IMobileServiceTableQuery<MobileType> and IMobileServiceTable<MobileType>

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs
```diff
@@ -58,34 +60,30 @@ public void Initialize(ExtensionConfigContext context)
...
-            IBindingProvider outputProvider = factory.BindToCollector<CosmosDBAttribute, OpenType>(typeof(CosmosDBCollectorBuilder<>), this);
-
-            IBindingProvider outputProviderJObject = factory.BindToCollector<CosmosDBAttribute, JObject>(typeof(CosmosDBCollectorBuilder<>), this);
-
-            IBindingProvider clientProvider = factory.BindToInput<CosmosDBAttribute, DocumentClient>(new CosmosDBClientBuilder(this));
+            // Apply ValidateConnection to all on this rule. 
+            var rule = context.AddBindingRule<CosmosDBAttribute>();
+            rule.AddValidator(ValidateConnection);
+            rule.BindToCollector<OpenType>(typeof(CosmosDBCollectorBuilder<>), this);
+            rule.BindToCollector<JObject>(typeof(CosmosDBCollectorBuilder<>), this);
```
brettsam: We shouldn't need this `JObject` one anymore, right? I added this as the workaround for the 'default type' issue in javascript.

MikeStall: Correct. I can remove it. I also added a test in SDK to verify this. 

---
In reply to: [151225522](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/331#discussion_r151225522) [](ancestors = 151225522)

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerBinding.cs
```diff
@@ -48,14 +48,8 @@ public CosmosDBTriggerBinding(ParameterInfo parameter, DocumentCollectionInfo do
...
-            IReadOnlyList<Document> triggerValue;
-            if (!TryAndConvertToDocumentList(value, out triggerValue))
-            {
-                throw new InvalidOperationException("Unable to convert trigger to CosmosDBTrigger.");
-            }
-
-            var valueBinder = new CosmosDBTriggerValueBinder(_parameter.ParameterType, triggerValue);
-            return Task.FromResult<ITriggerData>(new TriggerData(valueBinder, _emptyBindingData));
+            // ValueProvider is via binding rules. 
+            return Task.FromResult<ITriggerData>(new TriggerData(null, _emptyBindingData));
```
brettsam: So does this get called at all? Or is the value merged with this TriggerData somehow?

MikeStall: It's still called to produce the binding data. But then we impose a  IValueProvider separately 

---
In reply to: [151226742](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/331#discussion_r151226742) [](ancestors = 151226742)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
```diff
@@ -64,84 +67,40 @@ public void Initialize(ExtensionConfigContext context)
...
-
-            return false;
+            rule.WhenIsNotNull(nameof(MobileTableAttribute.TableName)).
+                BindToValueProvider<JObject>(BindForItemAsync);
+            rule.BindToValueProvider<MobileType>(BindForItemAsync);
         }
 
-        internal bool IsTableType(MobileTableAttribute attribute, Type paramType)
+        // Must have an single 'string Id' property 
+        internal class MobileType : OpenType
```
brettsam: Cool :-)

src/WebJobs.Extensions.MobileApps/Config/MobileAppsConfiguration.cs
brettsam: I believe there are enough tests for all of this, so if they pass, we should feel confident that the rules match.

brettsam:
@MikeStall -- now that your other changes are in SDK, can this go in?

PR #335, opened 2017-11-21
Sending 400 RUs as default

ealsur:
This change sends 400 RU as the default Throughput when creating a Cosmos DB collection instead of not sending anything.

The reason for this is that the default can vary between Subscription Types at this point and people were expecting 400 RU collections (the minimum) and saw 1000 RU collections being created.


ealsur:
AppVeyor tests seem to be failing on ApiHub related tests, but they were not modified on this PR.

brettsam:
I see some `DocumentDBTestUtility` tests failing. It looks like you'll need to update `SetupCollectionMock` to default to 400 -- it currently expects 0 as the default. This should simplify that method since 0 was a special case.

https://github.com/Azure/azure-webjobs-sdk-extensions/blob/v2.x/test/WebJobs.Extensions.Tests/Extensions/DocumentDB/DocumentDBTestUtility.cs#L26

Issue #342, opened 2017-12-01
Re-add ErrorTrigger


brettsam:
ErrorTrigger was removed when we removed internal usage of `TraceWriter` as it was built on the `TraceWriter`. This issue tracks the re-adding of this to the .NET Core branch.

mathewc:
The feature is currently built on the TraceWriter event stream, so should be directly portable over to the ILogger event stream, since the whole infrastructure is simply sliding window monitoring over an event stream.

I still think it is is a logical design for this to be based on the log stream. While we might consider invention of a new message bus to base this on, I don't think we need to do that.

rholsman:
Is there any way we can get examples of this new pattern? I happened to start a new webjob project within a week of the release of SDK 3.0 and had lots of trouble finding documentation on the changes. I waited for updates to the wiki or microsoft documentation but still can only find the old documentation which no longer works. I have went through the sample code and unit tests on the git hub pages and I find it difficult to find tests or examples for many things such as logging, di in a webjob, activators, webjob hosting (this may be better now), and error handling. With large breaking changes to patterns, libraries, and application design with each major release it makes it very hard to keep on top of best practices. It doesn't feel like there is functional parity between versions; I am sure there is but its very difficult to find especially as functionalities are moved into different objects and namespaces.

simpolainen:
Do you have any ETA on when the ErrorTriggerAttribute will be added to .NET Core?

mathewc:
No plans or ETA for this currently.

OzBob:
Still in wiki: https://github.com/Azure/azure-webjobs-sdk-extensions/wiki/Error-Monitoring 
and Readme references a code sample not in the repository.

jonathanantoine:
Any update ? Error management is pretty useful in real world :)

jimpaton-ufx:
Again, 18 months from the last comment, is there any progress on this? I just inherited a .NET 4.5.2 project that uses v1 of the WebJobs SDK and it's impossible to update in any meaningful way because it uses this feature and it's been removed with no replacement. If it's as simple as replacing every usage of `TraceWriter` with `ILogger`, why is the feature still missing after nearly 4 years and another major version release of the SDK?

remde:
Hi -- 18 months after the last comment, is there any progress or alternative for this?

PR #343, opened 2017-12-01
removing internal usage of TraceWriter; everything now uses ILogger

brettsam:
_Note: This won't build until https://github.com/Azure/azure-webjobs-sdk/pull/1438 is in and deployed._

This absorbs the "no more internal usage of `TraceWriter`" from the SDK branch. 

Notable changes:
- Removal of `ErrorTrigger`. This was built on the `TraceWriter` so no longer works. Issue #342 tracks the re-adding of this trigger.
- Removal of trace-related classes like `TraceFilter`, `TraceMonitor`, etc.


src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -84,7 +84,7 @@ internal System.Timers.Timer Timer
                 // check to see if we've missed an occurrence since we last started.
                 // If we have, invoke it immediately.
                 ScheduleStatus = await ScheduleMonitor.GetStatusAsync(_timerName);
-                _trace.Verbose($"Function '{_timerName}' initial status: Last='{ScheduleStatus?.Last.ToString("o")}', Next='{ScheduleStatus?.Next.ToString("o")}', LastUpdated='{ScheduleStatus?.LastUpdated.ToString("o")}'");
+                _logger.LogDebug($"Function '{_timerName}' initial status: Last='{ScheduleStatus?.Last.ToString("o")}', Next='{ScheduleStatus?.Next.ToString("o")}', LastUpdated='{ScheduleStatus?.LastUpdated.ToString("o")}'");
```
mathewc: Doesn't LogTrace map to Verbose?

src/WebJobs.Extensions/Extensions/Core/Bindings/ErrorTriggerAttributeBindingProvider.cs
```diff
@@ -1,170 +0,0 @@
...
-using Microsoft.Azure.WebJobs.Extensions.Core.Listener;
-using Microsoft.Azure.WebJobs.Host;
-using Microsoft.Azure.WebJobs.Host.Bindings;
-using Microsoft.Azure.WebJobs.Host.Listeners;
-using Microsoft.Azure.WebJobs.Host.Protocols;
-using Microsoft.Azure.WebJobs.Host.Triggers;
-
-namespace Microsoft.Azure.WebJobs.Extensions.Core
-{
-    internal class ErrorTriggerAttributeBindingProvider : ITriggerBindingProvider
```
mathewc: I'm somewhat OK with temporarily removing this as long as we prioritize adding it back soon by addressing https://github.com/Azure/azure-webjobs-sdk-extensions/issues/342. I don't want this feature to become a casualty.

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
brettsam: Verbose could map to either Trace or Debug as LogLevel has more possibilities than TraceLevel does. I believe I've chosen Debug everywhere. I was reserving Trace for even more detailed logging which we may want to add in the future.

PR #337, opened 2017-11-28
Exposing Change Feed options on Trigger

ealsur:
This change exposes a number of Change Feed Host options to the `CosmosDBTrigger`:

- Lease Preffix: This is used when the user wants to share the Lease collection among several Triggers and reduces operational costs.
- Polling intervals: Exposing different interval options that allow the user to customize the frequency on which the Trigger will check for changes and other time-related features.

Additionally, `UserAgentSuffix` was added to the `DocumentClient` initializations in the trigger to identify traffic coming from the Trigger in the account's analytics.

PR #349, opened 2017-12-13
Add "HaltOnError" attribute to CosmosDBTrigger

nkinnan:
This allows consumers to do reliable processing of change notifications without the possibility of dropping/losing a notification due to, for example, a temporary network issue or downstream dependency failure.

Default value preserves existing behavior.


nkinnan:
A couple tests are failing on the continuous integration build, but they are from different components.  I get even more test failures running them locally, so this seems like a maintenance thing unrelated to my change.

paulbatum:
If the local test failures are repeatable then it likely means that you dont have all the dependencies set up correctly (understandable, there are quite a few of them).

I looked at the test failures on the app veyor build for your PR and I can confirm that we have an issue where the infrastructure is running more tests than expected (we don't want to automatically run tests that require connection strings for PRs). Fixing this is tracked by:
https://github.com/Azure/azure-webjobs-sdk/issues/1340

When we are further along in review, we can tag this PR to run with full CI and then these failures should not be present.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -40,7 +42,18 @@ public Task OpenAsync(ChangeFeedObserverContext context)
...
+            // Not sure why these methods aren't declared with the "async" keyword, but wrapping this logic in a task taking an async 
+            // lambda so I can use it properly and avoid blocking with a Task.Wait() or Task.Result() call.
+            return Task.Run(async () =>
+            {
+                var result = await this.executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, CancellationToken.None);
+
+                if (!result.Succeeded && this.haltOnFailure == true)
+                {
+                    // unfortunate that we can't re-throw and preserve stack trace here
+                    throw result.Exception;
```
brettsam: Rather than throwing, can we swap to doing the checkpointing manually here, based on this `haltOnFailure`? Could we provide more control that way?

@ealsur -- you probably know more about what's possible here...

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -89,5 +89,17 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
...
+        /// Optional, defaults to false.
+        /// If false, any exception thrown by the function will be ignored and the checkpoint updated.
+        /// If true, throwing an exception from the function will cause the Feed Processor not to checkpoint. Processing will be halted for the 
+        /// current partition for a brief time and then any notifications since the last checkpoint will be re-delivered.
+        /// If set to true, the function should implement its own any poison document handling to prevent an un-processable notification from 
+        /// permanently halting processing of the partition and potentially running up execution costs; this flag is meant to enable guaranteed
+        /// processing of every notification, even in case of a temporary "downstream dependency failure" preventing the function from completing
+        /// successfully.
+        /// </summary>
+        public bool HaltOnFailure { get; set; } = false;
```
brettsam: This is more of a `RetryOnFailure`, isn't it?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
brettsam: For example, could we do any of these things?:
- Set a max retry for a document (is there any tracking for that?)
- Set some retry delay (is there a way to 'pause' the change feed?)

brettsam: More questions :-)
- This will restart the observer -- but how long will it wait to restart? A few seconds? Minutes? Is it configurable?
- I'm assuming the entire batch is retried? So if some were successfully processed, those will be sent again as well?

ealsur:
I added one comment, but I would really like @paulbatum's point of view regarding this change. The code itself is correct, but is the effect that concerns me.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
nkinnan: The naming was suggested by Matias due to the underlying behavior of the change feed processing library.  Its technically correct, but for a public API, "RetryOnFailure" may be more easily understood.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: Comments on 1)  This is desired behavior for customers who require reliable processing, but I agree it should be documented well enough for them to understand the implications.  I've tried to make the comments on the attribute very descriptive, but MSDN/etc should also be updated.  Dead lettering would be a possible alternative for some but not all situations.  For example, in our production system in which failure to process is an issue that require immediate attention, I set retry count to infinite on our QueueAttributes.  A dead-lettered notification/queue item means dropped customer data and downstream failures.  Additionally, in this case, we require in-order processing as well, so we can't just check the dead letter queue periodically.  I can see it being a valid solution for other use cases though.

Comments on 2)  Suspect this new observer will be created based on lease expiry time.  The "gotcha" about "first run / no checkpoint was ever created" I believe is a general issue with the change feed library and something I've actually hit in real-life using that library directly!  My account has 24 partitions, and for a long time I couldn't figure out why some notifications may be lost if I simply shut down the web job, updated some documents, and restarted the web job.  The notifications that went to a partition that was never checkpointed just got dropped!  Understanding that this is because the checkpoint is in the form of a continuation token from DocDB it finally made sense, but was quite confusing at first.  I'm considering submitting separate feedback on this as its a problem that is not specific to CosmosDBTrigger.

Yes, it does require the function to be idempotent and have the ability to process replays of sequences instead of individual notifications.  This could be fixed by setting max batch size to 1 and checkpoint-every to 1 as well, at the cost of performance.  I don't believe these change feed processor settings are currently exposed by CosmosDBTrigger.

nkinnan: (the "max batch size" and "checkpoint-every" settings I referred to are in ChangeFeedHostOptions.CheckpointFrequency)

paulbatum: I agree with the concerns around infinite loops and would prefer a design along the lines of what Brett eluded to with x retries and y back off. It is already the case that customers have to be aware they are getting an "at least once" guarantee and that they need to write their code to allow for the possibility of retries so I'm not that concerned about a design that makes retries more frequent.

Can we talk more about how straightforward it would be to implement "retry x times with y back off"? I don't know the code well enough to understand whether this is something we can easily do purely within functions or if we need functionality from the change feed processor.

nkinnan: Are you suggesting "before propagating the throw back to the change feed processor library and thus causing it not to checkpoint, do a couple retries under the attribute's control while holding onto the partition lock"?  I almost implemented that as part of the change, it was only a couple more lines, but given that retry will happen anyway, didn't consider it important and didn't want to confuse the pull request.  I'm curious about your thoughts here and what concerns you have that "attribute controlled retries" would resolve?  

In my situation, the most common case would be, for example, a Cosmos outage that could last from a few minutes to a few hours and I'd prefer to halt processing /ordered/ notifications until it's back up.  Or a similar issue surrounding a Splunk dependency.  (For me, in each case the write action is effectively idempotent, and even a series of retried writes would be idempotent.)

ealsur: @nkinnan Max Batch Size is coming as a new attribute too among other things, working on it here https://github.com/Azure/azure-webjobs-sdk-extensions/pull/337

nkinnan:
I have added a local retry strategy as requested.  It uses the Microsoft EnterpriseLibrary and is fully user configurable.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
paulbatum: lets move this conversation out of the inline comments and into the main PR comments because they are hidden by github automatically as the code is updated.

paulbatum:
Haven't looked too closely yet but I should mention that we really want to avoid taking on new dependencies. Its a complicated topic but at a high level, the more dependencies we have, the more chance we have of conflicting with the dependencies our users have.

Moving some inline conversation to the main PR conversation - earlier I wrote:

> I agree with the concerns around infinite loops and would prefer a design along the lines of what Brett eluded to with x retries and y back off. It is already the case that customers have to be aware they are getting an "at least once" guarantee and that they need to write their code to allow for the possibility of retries so I'm not that concerned about a design that makes retries more frequent.
> 
> Can we talk more about how straightforward it would be to implement "retry x times with y back off"? I don't know the code well enough to understand whether this is something we can easily do purely within functions or if we need functionality from the change feed processor.

@nkinnan wrote:

> Are you suggesting "before propagating the throw back to the change feed processor library and thus causing it not to checkpoint, do a couple retries under the attribute's control while holding onto the partition lock"? I almost implemented that as part of the change, it was only a couple more lines, but given that retry will happen anyway, didn't consider it important and didn't want to confuse the pull request. I'm curious about your thoughts here and what concerns you have that "attribute controlled retries" would resolve?
> 
> In my situation, the most common case would be, for example, a Cosmos outage that could last from a few minutes to a few hours and I'd prefer to halt processing /ordered/ notifications until it's back up. Or a similar issue surrounding a Splunk dependency. (For me, in each case the write action is effectively idempotent, and even a series of retried writes would be idempotent.)

Its possible I have misunderstood some fundamentals about how this works, so lets try to clear that up first. I have been relying on my knowledge of how Event Hubs processing works in functions to reason about the change feed processor and that might have lead me astray.

When you write an Event Hub triggered function, a checkpoint will occur regardless of whether the function completed succesfully or not. The reason for this is that we don't want to allow one bad message that consistently results in an exception to cause processing of the event stream to stop (or more accurately: infinitely retry the same batch over and over). 

The comment _before propagating the throw back to the change feed processor library and thus causing it not to checkpoint_ is confusing me. If we allow this to happen, then doesn't that create the above problem?

nkinnan:
Yes, in fact that's the entire point of this PR, to allow that "block" to happen, preventing checkpointing if the function throws, and ensure that if Cosmos or Splunk (or any other dependencies that I'm talking to in the function) are down, I can always get re-delivery until I successfully propagate the notification to all of them.  Having the notifications not be reliable makes this trigger currently useless to me, forcing me to go directly to the change feed processor (where I do propagate the exception up) in order to be able to get that kind of reliability.  

Which is in fact what I have done since I couldn't wait for this PR to go through.  I'm just trying to contribute this so that I can move to a Function instead in the future, instead of having to deal with a WebJob and directly accessing the change feed processor.

The issue isn't a poison message, its transient errors due to downstream dependencies.  The XML comments on the "RetryOnFailure" explicitly warn about poison message handling though.  Its a tradeoff the user has to make for themselves: A) turn RetryOnFailure on to ensure transient failures do not cause data loss while implementing my own poison logic if necessary B) drop notifications due to transient failures.

I thought you were asking for "local retries on top of that".  I suppose if the attribute consumer is asking for infinite retries, there's no need to propagate the exception into the change feed processor (which would never happen anyway with infinite retries...) and if the service goes down or is redeployed the checkpoint still wouldn't be written.  That would basically just be removing the RetryOnFailure attribute but leaving the RetryStrategy attribute.

Understood about the dependencies concern.  I was simply trying to use a standard Microsoft created and published retry library.  If you'd prefer that I duplicate this functionality by basically copying the classes into the WebJobs namespace (like AzureStorage and others have done in order to support the exact same functionality, everyone's implementation "looks identical" but in a different namespace) then I'd be happy to do so.

ealsur:
@nkinnan I tested what might happen when an exception is thrown and I could verify that indeed, the internal Processor would retry the batch again, if the throw happens again, it would do it again and again. Like you said, this makes it so there is a guarantee processing order, but it does have drawbacks.

@paulbatum wrote:

> When you write an Event Hub triggered function, a checkpoint will occur regardless of whether the function completed succesfully or not. The reason for this is that we don't want to allow one bad message that consistently results in an exception to cause processing of the event stream to stop (or more accurately: infinitely retry the same batch over and over).

That is exactly the same behavior for the Cosmos DB trigger right now, just in case there was confusion 😄 

@nkinnan's change aims at allowing the user to actively say "I want the checkpoint not to happen" and so the batch of messages/changes would be retried. It is true and I also stated it, that it can lead to infinite loops / unexpected billing (in consumption mode) and double processing of documents, and this would require a detailed documentation (maybe it's something we can add on the Functions UI docs we already have for each attribute). 

@nkinnan, I think that our concern regarding infinite loops and the other collateral effects is related to customer problems and support that might come from the accidental or misuse of this feature, while it is a good addition, we have to evaluate the side effects and future issues it might have.

Regarding dependencies, adding to what @paulbatum said, we are shipping the Cosmos DB Trigger in a Net Standard 2.0 environment, so any dependency has also to be Net Standard 2.0 compliant.

paulbatum:
OK cool, I think most of my confusion is cleared up. When I suggested earlier in the thread to use a retry strategy with some amount of backoff, that was **instead of** the "halt" behavior (not in addition). So I agree with all of this:

@nkinnan wrote:

>  I suppose if the attribute consumer is asking for infinite retries, there's no need to propagate the exception into the change feed processor (which would never happen anyway with infinite retries...) and if the service goes down or is redeployed the checkpoint still wouldn't be written. That would basically just be removing the RetryOnFailure attribute but leaving the RetryStrategy attribute.

This allows for a user who really knows what they are doing to set the retry count to something huge and get the behavior you were looking for, but avoids the problem of having a feature that can be simply turned on and result in incredibly large bills.

As for what code should be used to achieve the retry behavior, it would be best for @brettsam to weigh in as he's had to do this before.

nkinnan:
Glad we're all on the same page now!  To sum up, in terms of having a RetryStrategy attribute, plus the ability to specify an "infinite retries" strategy:
* This is not something the user could do by accident and thus not a concern for you in terms of accidental billing - they didn't flip a bit, they literally specified "infinite retries" manually
* Does not throw back to the change feed processor (since its infinite retries) - not really a concern, but just for the record
* Checkpoint still would not be written until the change was successfully processed (even in case of machine cycle or function redeployment) meaning absolutely 100% guaranteed I'll never drop a notification
* Any retries, even in case of service restart or function deployment causing checkpoint loss, would still be in-order, even if it causes a sequence to be replayed (if checkpoint-every isn't 1)

This design seems to address all your concerns, and still allows me to get what I want.  In fact I like this design better in hindsight so I'm glad we hashed it out!

I'll go ahead and update the PR to implement it that way instead, and remove the "RetryOnFailure" attribute.

Last concern is simply dependencies.  For the sake of simplicity, I'll just "duplicate" the classes into the WebJobs namespace.  Other Azure libs do the same thing with RetryStrategy and probably for the same reason.  If @brettsam prefers a different design, I can do another iteration.  Its unfortunate to have to duplicate, but I understand the concern.

brettsam:
A couple of general comments:
- There's some great value here and we'd like to get it in. However, we're approaching the holidays (most of the team is already OOF -- and I only have one more day in the office) -- so this won't be merged until the new year. But I'll give you some review comments that you can work on in the meantime.
- As an FYI -- this is going into our `dev` branch, which targets Functions v2, currently out in Preview. The good news is that this makes it much easier to approve as we have some time before GA to reason through fine-tune the behavior. We'd also likely take this logic and generalize it to work with Event Hubs (and maybe Queues). This does mean that it won't show up in the currently GA'd Functions release -- we're tougher on new features there.
- We should avoid the added dependency, if possible. Keeping our dependency count low makes everything nicer. If someone is writing a Function and wants a newer version of this library, we end up seeing conflicts. And auto-upgrading everyone can lead to other kinds of breaks. So, if it's possible to copy this code locally and use it internally, that would be best. It also allows us to tweak it into a more general (and perhaps simpler) `WebJobsRetryPolicy` in the future.
- It'd be great to have some end-to-end tests that target a real CosmosDB service to make sure this works as we expect. Unfortunately, our dev repo is in a state where a lot of the end-to-end tests aren't running at the moment due to the .NET core update. We want to get those back online as one of the first things we do in the new year. But it does mean that I don't have a pointer to where or how you can write some of these tests. Ideally we'd add a few simple CosmosDBTrigger end-to-end tests in this repo that targeted a real service (I don't think mocking the `ChangeFeedProcessor` would give us the confidence we'd need in the behavior, but @ealsur can weigh in on that). Either @ealsur or I should be able to help with that -- but again it may be a couple of weeks.

I'm going to make a couple more comments in the code...

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -89,5 +90,26 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
...
+        /// </summary>
+        public bool RetryOnFailure { get; set; } = false;
+
+        /// <summary>
+        /// Optional, defaults to Incremental.DefaultExponential
+        /// This setting takes effect when RetryOnFailure is set to true (it defaults to false) and controls the local 
+        /// retry strategy used before abandoning the checkpoint.  
+        /// For an example of a custom retry strategy, see https://msdn.microsoft.com/en-us/library/hh680943(v=pandp.50).aspx
+        /// </summary>
+        public RetryStrategy RetryStrategy { get; set; } = Incremental.DefaultExponential;
```
brettsam: I don't think this will work as an Attribute property since C# requires that an attribute property be a compile-time constant value. @ealsur just hit this with another property.

Can we simplify this down to be something like `MaxRetryCount` and handle all the details internally? The default could be 0 to keep current behavior. We'd want this metadata to be easily definable in JSON for Functions -- so keeping the property simpler would benefit us there as well.

nkinnan:
Thanks @brettsam for the comments!  I'm really happy to see that this new design may be propagated to other triggers, as it enables use cases not otherwise supported.

Understood about the timeline and holidays, in fact I'll have to postpone my PR update which conforms to the new design for a few weeks, but look forward to picking it up again once I'm back in town.

main comment:
----------------

In regards to removing the dependency, you said that one can't use a class instance as an attribute value.  Makes sense, but what about statics?  The value I assigned is a static with no instance members or methods.  Even the delegate returned is statically defined in the form of a lambda.  It compiled so I assumed it worked.  You suggested adding a simple "MaxRetryCount" attribute instead.

I have a concern that this may not be flexible enough in the general case where it isn't simply a case of "I want infinite retries".  It seems important to be able to specify the strategy (linear/infinite/constant) as well as starting delay and how that delay changes over time.  If that's not possible, I think its important to come up with a "works for 95% of our customers" strategy with as few required user inputs as possible, but I'm not sure what that looks like.

If the static class doesn't work (are you sure?) can you recommend a robust design, or at minimum a set of standard defaults that are un-changeable except for MaxRetryCount in order for me to implement a generally applicable retry strategy?

My instinct if the flexibility of a custom designed strategy is not possible, is to say "no delay, immediate retry as many times as you specify, and if you want delay in-between them, implement that yourself in your function's own catch() statement before re-throwing.

Last thing: I'll wait for your guidance on testing after the new year once you've had a chance to address any current issues.  Please let me know and I'm happy to do any validation that I can.

Thanks!

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
ealsur: The issue that Brett refers is that the JIT seems to only allow primitives to be passed as parameters. It will compile but fail at runtime. I tried it before 😢 
So go for the suggestion that Brett stated, it will save you headaches 😄

ealsur:
@nkinnan I added the comment after Brett's but regarding the static class, it will compile, but on runtime the JIT cannot handle classes as parameters, only primitives.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -40,7 +45,44 @@ public Task OpenAsync(ChangeFeedObserverContext context)
...
+                        {
+                            await Task.Delay(delay);
+
+                            retryCount++;
+                            continue;
+                        }
+                        else
+                        {
+                            // unfortunate that we can't re-throw and preserve stack trace here
+                            throw result.Exception;
```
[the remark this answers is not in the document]
nkinnan: Per discussion / design change, no need to throw anymore.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -40,7 +45,44 @@ public Task OpenAsync(ChangeFeedObserverContext context)
 
         public Task ProcessChangesAsync(ChangeFeedObserverContext context, IReadOnlyList<Document> docs)
         {
-            return this.executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, CancellationToken.None);
+            int retryCount = 0;
+
+            // Not sure why these methods aren't declared with the "async" keyword, but wrapping this logic in a task taking an async 
+            // lambda so I can use it properly and avoid blocking with a Task.Wait() or Task.Result() call.
+            return Task.Run(async () =>
```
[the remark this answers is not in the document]
nkinnan: Thanks, just wasn't sure about changing the syntax in case there was a specific reason for it, I have added the async keyword.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
nkinnan: Thanks for the heads up!  Per discussion I have changed this to an int value.

nkinnan:
Thanks everyone!  Sorry for the delay over the holidays, but I have implemented the change as discussed and addressed all comments.

nkinnan:
Had two local commits so the description isn't showing properly: "Switch implementation to use a RetryCount attribute per discussion. This enables guaranteed reliable processing in case of transient downstream dependency failures in case of "-1" or "infinite" retries. It also enables the customer to specify their own custom tradeoff by specifying RetryCount and optionally implementing a backoff strategy and/or poison notification handling inside their Function implementation."

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -38,9 +40,30 @@ public Task OpenAsync(ChangeFeedObserverContext context)
...
+            {
+                var result = await this.executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, CancellationToken.None);
+
+                if (result.Succeeded)
+                {
+                    // notification processed successfully
+                    return;
+                }
+
+                if (retryCount != -1 && retries >= retryCount)
```
ealsur: If the default value is 0, why are you comparing against -1? Wouldn't this check be: retries == retryCount ?

If the value is 0, it would return on the first error (no retry). If the value is higher it would increment the local variable until it matches the `retryCount` value and also stop.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -107,7 +107,7 @@ public CosmosDBTriggerAttributeBindingProvider(INameResolver nameResolver, Cosmo
                 throw new InvalidOperationException(string.Format("Cannot create Collection Information for {0} in database {1} with lease {2} in database {3} : {4}", attribute.CollectionName, attribute.DatabaseName, attribute.LeaseCollectionName, attribute.LeaseDatabaseName, ex.Message), ex);
             }
 
-            return new CosmosDBTriggerBinding(parameter, documentCollectionLocation, leaseCollectionLocation, leaseHostOptions);
+            return new CosmosDBTriggerBinding(parameter, documentCollectionLocation, leaseCollectionLocation, leaseHostOptions, attribute.RetryCount);
```
ealsur: Check at some point, either here or in the Binding or Listener, for values lower than -1. Technically since it's an `int`, the user could input negative values, maybe default to `0` if any negative value is set.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: -1 means "infinite retries" as documented on the attribute, so I need to ignore the count of retries in that case

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
nkinnan: The retry code would treat any negative value other than -1 (meaning "infinite") as effectively zero, but I agree it would be nice to notify the customer of a potential typo.  Should I just throw an ArgumentException in this case?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: Also, I could do "[inf check] && retries == retryCount" instead of "[inf check] && retries >= retryCount" as you suggested, but I used the >= just as sort of failsafe even though the value should never increment more than once per loop.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
nkinnan: Will go the ArgumentException route for now and iterate, pending additional guidance.

nkinnan: Correction: InvalidOperationException, following the existing examples in CosmosDBTriggerAttributeBindingProvider

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -107,7 +107,13 @@ public CosmosDBTriggerAttributeBindingProvider(INameResolver nameResolver, Cosmo
                 throw new InvalidOperationException(string.Format("Cannot create Collection Information for {0} in database {1} with lease {2} in database {3} : {4}", attribute.CollectionName, attribute.DatabaseName, attribute.LeaseCollectionName, attribute.LeaseDatabaseName, ex.Message), ex);
             }
 
-            return new CosmosDBTriggerBinding(parameter, documentCollectionLocation, leaseCollectionLocation, leaseHostOptions);
+            // -1 is allowed and indicates "infinite retries", but other negative values are not valid
+            if (attribute.RetryCount < -1)
+            {
+                throw new InvalidOperationException(string.Format("RetryCount value {0} is not valid, value must be either positive, or -1 for 'infinite retries'", attribute.RetryCount));
```
brettsam: minor nit: we try to use string interpolation for new strings we add -- and `nameof()` for property names so you can change this to: `$"{nameof(RetryCount} value {attribute.RetryCount} is not valid. Value must be either positive, or -1 for infinite retries."`

nkinnan: The other strings in the class (for the branch/version of code I am looking at) are all string.Format'ed rather than using $ and I don't see any other nameof() calls so I assume that's in another branch or something?  Anyway, I've made the requested change and pushed.

dnlhll:
Is there a timeline for this feature? Are there still test framework changes that are blocking? It would be great to have this in the v2 preview. Excited to start using this!

paulbatum:
Sorry, yes we are still having test issues relating to V2 and cosmos DB. It is certainly our intent to get those issues resolved and this PR merged but I don't have an ETA for you yet.

dnlhll:
Ok, it looks like this change has been complete for a while. Is the test framework in this repository, or is it somewhere else? Is there a PR out for the test framework fix?

brettsam:
@nkinnan, the tests are here: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/test/WebJobs.Extensions.Tests/Extensions/CosmosDB/CosmosDBEndToEndTests.cs#L30-L64.

If we can get this rebased and an e2e test written, we should be able to get this in.

dnlhll:
@nkinnan, are you still actively working on this feature?

nkinnan:
The feature is complete.  It has just taken 6 months to get the contribution unblocked, since the Azure Functions test framework was broken and so I was unable to validate the changes as is (reasonably) required before commit.

I'm still passionate about this, believe it is the right thing for all Azure customers, and would be happy to contribute the needed test cases (now doing so is possible).

At the moment, I am temporarily consumed by feature work on my own team, and it may be several weeks or months before I'm able to, after all this time, write the needed test cases.

dnlhll:  if you want to take over with the test cases, and finishing this commit, I think that would be great for both of us, as well as all Azure customers.  Otherwise, it may take me a bit longer to close on this now that progress is unblocked.  If you don't have the bandwidth either, I'l follow up, I've just almost gotten used to having to manually deploy our non-AppService executable that does the same thing in a less supportable way...

dnlhll:
@nkinnan, unfortunately, I do not have the bandwidth to get up to speed on this code base and make the test changes required.

nkinnan:
I’ll follow up as soon as my schedule allows, it may be a few weeks or a month though.  I’m currently in the middle of closing on a large project.

paulbatum:
Yes I'm sorry that this change has been blocked in PR for so long due to test issues. We still intend to get this change in but I am still struggling to provide any type of dates.

dnlhll:
@paulbatum @nkinnan, is the feature unblocked now, is the only work remaining the test cases? Is there still work that needs to be done elsewhere?

nkinnan:
I'm currently on vacation and should have time to help improve the test coverage.  

Looks like @brettsam  added a very basic test about 10 days ago.  

I'll need assistance with test setup as I get a lot of failures due to missing test config values.  Modifying the app.config in WebJobs.Extensions.CosmosDB.Tests to add the missing settings does not seem to work.  This is sort of expected since there aren't even "fill in this value" settings there, but I'm not sure where they're supposed to go or what the full list is.  I'm happy to create azure test resources to fill in the values if necessary.

nkinnan:
Tests also seem to be failing in the continuous integration build.  This will be blocking of course.

dnlhll:
@brettsam @nkinnan are these tests failing because of issues with the feature, or are there gaps in the test framework?

brettsam:
We'll need to rebase this and take a look. It may seem odd to create an issue at this point, but that's how we track our work items and I don't want this to get lost (again). So I've created issue #509 to track the work of getting this finalized. We'll review it at our next triage.

dnlhll:
@brettsam, thanks for keeping at this. looking forward to this feature!

nkinnan:
I've completed the rebase that Brett started a month ago and committed such.  Unless I'm missing something, the remaining blockers are writing additional tests (which I am happy to do) and not being able to complete a test pass due to https://github.com/Azure/azure-webjobs-sdk/issues/1340 which is still open.

All I need is the barest support.  My passion wanes while myself and other customers have waited for nearly a year.  This is such a simple and obvious improvement.  Having embraced the idea of open source and contributing to other teams, other customers, and the greater whole, I have to say that I am very let down right now.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -38,9 +40,31 @@ public Task OpenAsync(ChangeFeedObserverContext context)
...
 
-        public Task ProcessChangesAsync(ChangeFeedObserverContext context, IReadOnlyList<Document> docs)
+        public async Task ProcessChangesAsync(ChangeFeedObserverContext context, IReadOnlyList<Document> docs)
         {
-            return this.executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, CancellationToken.None);
+            int retries = 0;
+
+            while (true)
+            {
+                var result = await this.executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, CancellationToken.None);
```
brettsam: It may be useful to pass the retry count here. That'd let the function decide if it wants to write out the document to some "poison" location for later processing if its on the last attempt.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -38,9 +40,31 @@ public Task OpenAsync(ChangeFeedObserverContext context)
...
+                if (retryCount != -1 && retries >= retryCount)
+                {
+                    // Function invocation failed, but no retries are left
+                    return;
+                }
+
+                // Function invocation failed, but retries are left, or infinite retries was specified
+                retries++;
+                continue;
+            }
```
brettsam: It'd be good to get some trace logging in here. I can plumb that through.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
```diff
@@ -38,9 +40,31 @@ public Task OpenAsync(ChangeFeedObserverContext context)
...
+                var result = await this.executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, CancellationToken.None);
+
+                if (result.Succeeded)
+                {
+                    // notification processed successfully
+                    return;
+                }
+
+                // note: -1 means "infinite retries", so ignoring the count of retries in that case
+                if (retryCount != -1 && retries >= retryCount)
```
brettsam: What was the scenario for allowing infinite retries? Anything infinite always scares us, from a supportability standpoint  :-)

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -149,7 +149,13 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
             }
 
-            return new CosmosDBTriggerBinding(parameter, documentCollectionLocation, leaseCollectionLocation, leaseHostOptions, changeFeedOptions, _logger);
+            // -1 is allowed and indicates "infinite retries", but other negative values are not valid
+            if (attribute.RetryCount < -1)
+            {
+                throw new InvalidOperationException($"{nameof(attribute.RetryCount)} value {attribute.RetryCount} is not valid. Value must be zero, positive, or -1 for infinite retries.");
+            }
+
+            return new CosmosDBTriggerBinding(parameter, documentCollectionLocation, leaseCollectionLocation, leaseHostOptions, changeFeedOptions, _logger, attribute.RetryCount);
```
brettsam: We should make `RetryCount` go through the `NameResolver` so it can be adjusted with app settings. I can update that.

brettsam:
Thanks @nkinnan. Sorry that this got punted for so long. Now that v2 has hit GA, I'll see what I can do to get this pushed over the finish line. I can handle the test-writing and the last few cleanup issues.

I've asked @fabiocav and @mathewc to take a look at the approach as well. I want to make sure that whatever we do here looks good in a broader sense. For example, we may want to do the same thing for Event Hubs (since the behavior is similar) so we need to make sure we're not creating conflicting designs for the different triggers.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: I'll take a look at this.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: Thanks!

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: There are I believe many cases where a system must process in-order and reliably.  At Least Once processing.  This can be critical for many use cases.  Since Documents may be updated incrementally, it may compromise downstream data integrity to "skip" something.  Even with a Dead Letter Queue (which is a good idea and an excellent add-on to this change), reverse-engineering the change sequence to "fix up" downstream dependencies, trying to "insert the missed update" may be difficult, requiring a lot of extra design and implementation effort, or even be effectively impossible.  

For our use case and I believe many others, it is better to see an error logged, repeatedly, until someone fixes it and unblocks the system.  Otherwise our data processing becomes corrupt.  One excellent example which you may be indirectly familiar with is using the change feed in order to maintain a cache of "what device ids are in our document store" and not processing NGP delete requests beyond the watermark of that cache.  Doing a costly self-join query on every NGP delete request would be prohibitive given the size of our collection.

There are several email threads (plus inline in this pull request) discussing this problem with your Architect, and further enumerating the reasons and justification.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
nkinnan: Thx, I was not familiar with that or may have missed something.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: > Thanks @nkinnan. Sorry that this got punted for so long. Now that v2 has hit GA, I'll see what I can do to get this pushed over the finish line. I can handle the test-writing and the last few cleanup issues.

Much appreciated!  I'm happy to contribute however I can.

> I've asked @fabiocav and @mathewc to take a look at the approach as well. I want to make sure that whatever we do here looks good in a broader sense. For example, we may want to do the same thing for Event Hubs (since the behavior is similar) so we need to make sure we're not creating conflicting designs for the different triggers.

Absolutely!  All necessary care should be taken to ensure the design is as forward looking and consistent as possible.  You brought up the possibility of a Dead Letter Queue for the change feed which I think would be an excellent addition.  Just bringing that up as part of the consideration if you have existing DLQ functionality for EventHub and how that would fit in with a retry knob on the EH side.

nkinnan:
> Thanks @nkinnan. Sorry that this got punted for so long. Now that v2 has hit GA, I'll see what I can do to get this pushed over the finish line. I can handle the test-writing and the last few cleanup issues.

Much appreciated! I'm happy to contribute however I can.

> I've asked @fabiocav and @mathewc to take a look at the approach as well. I want to make sure that whatever we do here looks good in a broader sense. For example, we may want to do the same thing for Event Hubs (since the behavior is similar) so we need to make sure we're not creating conflicting designs for the different triggers.

Absolutely! All necessary care should be taken to ensure the design is as forward looking and consistent as possible. You brought up the additional possibility of a Dead Letter Queue for the change feed which I think would be an excellent addition. Just bringing that up as part of the consideration if you have existing DLQ functionality for EventHub and how that would fit in with a retry knob on the EH side for consistency with this change.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerObserver.cs
nkinnan: Also please remember in the "At Least Once" context, that the reason for failure may be downstream dependencies.  For example, I am writing information related to the change to an Azure Table, and Storage goes down for a few minutes.  Changes are distributed to the change feed processor At Least Once already, but that loses meaning when the processor function can't actually process the change, and that may be out of the function's control.

brettsam: @ealsur raised a concern (internally) that long retry loops could cause the lease to expire as the ChangeFeedProcessor would never renew the lease on the partition. That means another instance could start processing these same documents, leading to duplicate processing.

We're contacting the ChangeFeedProcessor owners to see if they have a better recommendation or if we can get better control in a future release.

I see the valid scenario -- but we want to make sure that all the pieces we're using can handle an infinite retry like this. Will report back soon.

dnlhll:
@brettsam is there any internal progress on this? I know you mentioned people were discussing some of the design decisions.

dnlhll:
@brettsam @fabiocav @mathewc has there been any internal conversation on this feature? Hoping we can get the PR merged before the 1 year mark.

dnlhll:
@nkinnan @brettsam @fabiocav @mathewc @ealsur can someone chime in on the status of this PR.

brettsam:
I'm going to go ahead and close this as the feature will take a different shape when implemented. That work is being tracked in https://github.com/Azure/azure-webjobs-sdk/issues/2036. We'll be prioritizing it along with our other work items in our upcoming sprints.

PR #354, opened 2017-12-27
Fix Cosmos DB exception when using headers and query binding expressions

anthonychu:
Currently this exception occurs when trying to use binding expressions that access headers and query from the HTTP trigger in Azure Functions:

`"sqlQuery": "select * from c where c.id = {headers.foo}"`

> Microsoft.Azure.WebJobs.Host: Error while accessing 'foo': property doesn't exist.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
+++ b/src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
@@ -25,9 +25,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             {
                 throw new ArgumentNullException(nameof(bindingData));
             }
-
-            CosmosDBAttribute docDbAttribute = resolvedAttribute as CosmosDBAttribute;
-            if (docDbAttribute == null)
+            
+            if (!(resolvedAttribute is CosmosDBAttribute cosmosDBAttribute))
             {
                 throw new NotSupportedException($"This policy is only supported for {nameof(CosmosDBAttribute)}.");
             }
@@ -36,11 +35,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             SqlParameterCollection paramCollection = new SqlParameterCollection();
 
             string bindingTemplatePattern = bindingTemplate.Pattern;
-
+            
             IDictionary<string, string> expandedTokens = GetExpandedTokens(bindingTemplate, bindingData);
-
-            // also build up a dictionary replacing '{token}' with '@token' 
-            IDictionary<string, string> replacements = new Dictionary<string, string>();
             foreach (var token in expandedTokens)
             {
                 string bindingExpression = $"{{{token.Key}}}";
@@ -52,7 +48,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
                 }
             }
 
-            docDbAttribute.SqlQueryParameters = paramCollection;
+            cosmosDBAttribute.SqlQueryParameters = paramCollection;
 
             return bindingTemplatePattern;
         }
--- a/src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
+++ b/src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
@@ -43,9 +43,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             IDictionary<string, string> replacements = new Dictionary<string, string>();
             foreach (var token in expandedTokens)
             {
-                string sqlTokenName = $"@{EscapeSqlParameterName(token.Key)}";
-                paramCollection.Add(new SqlParameter(sqlTokenName, token.Value));
-                bindingTemplatePattern = bindingTemplatePattern.Replace($"{{{token.Key}}}", sqlTokenName);
+                string bindingExpression = $"{{{token.Key}}}";
+                if (bindingTemplatePattern.Contains(bindingExpression))
+                {
+                    string sqlTokenName = $"@{EscapeSqlParameterName(token.Key)}";
+                    paramCollection.Add(new SqlParameter(sqlTokenName, token.Value));
+                    bindingTemplatePattern = bindingTemplatePattern.Replace($"{{{token.Key}}}", sqlTokenName);
+                }
             }
 
             docDbAttribute.SqlQueryParameters = paramCollection;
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
```diff
@@ -36,18 +35,44 @@ public string TemplateBind(PropertyInfo propInfo, Attribute resolvedAttribute, B
...
+                    IDictionary<string, string> tokens = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                    foreach (var item in tokenDictionary)
+                    {
+                        string fullToken = $"{token}.{item.Key}";
+                        bool tokenInUse = bindingTemplate.Pattern.Contains(fullToken);
+                        if (tokenInUse)
+                        {
+                            string sqlToken = GetSqlParameterName(fullToken);
+                            paramCollection.Add(new SqlParameter(sqlToken, item.Value));
+                            tokens.Add(item.Key, sqlToken);
```
ealsur: This logic seems duplicated from the other above, could they be encapsulated in a helper method to avoid some future break by an edit that forgets the update both?

src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
```diff
@@ -36,18 +35,44 @@ public string TemplateBind(PropertyInfo propInfo, Attribute resolvedAttribute, B
...
+            IDictionary<string, object> replacements = new Dictionary<string, object>();
             foreach (var token in bindingTemplate.ParameterNames.Distinct())
             {
-                string sqlToken = $"@{token}";
-                paramCollection.Add(new SqlParameter(sqlToken, bindingData[token]));
-                replacements.Add(token, sqlToken);
+                object tokenObject = bindingData[token];
+                if (tokenObject is string tokenString)
+                {
+                    string sqlToken = GetSqlParameterName(token);
```
ealsur: This case seems like a subset of the other in terms of logic, just thinking out loud, but what if we had a single logic that created the replacements and the difference is that when the param is a `string` we are sending an input Dictionary of 1 item? Probably @brettsam has a more valuable feedback 😄

ealsur:
@brettsam I added you since you have probably more insight on the bindings

brettsam:
Hey @MikeStall -- I'm taking a look at this as well, but do we have some re-usable logic for parsing out nested property values?

src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
```diff
@@ -36,18 +35,44 @@ public string TemplateBind(PropertyInfo propInfo, Attribute resolvedAttribute, B
...
 
-            string replacement = bindingTemplate.Bind(new ReadOnlyDictionary<string, string>(replacements));
+            string replacement = bindingTemplate.Bind(new ReadOnlyDictionary<string, object>(replacements));
             return replacement;
         }
+
+        private string GetSqlParameterName(string name)
+        {
+            const string safeSeparator = "_";
+            return "@" + name.Replace(".", safeSeparator).Replace("-", safeSeparator);
```
brettsam: Are these not allowed in SqlParameter names? Is the list larger than this, or is it just these two characters? When doing replacements like this, I worry that we could stomp over another matching token. Even if it's rare, I wonder if there's another way to handle it.

test/WebJobs.Extensions.Tests/Extensions/CosmosDB/CosmosDBSqlResolutionPolicyTests.cs
```diff
@@ -52,5 +52,30 @@ public void TemplateBind_DuplicateParameters()
             Assert.Single(resolvedAttribute.SqlQueryParameters, p => p.Name == "@foo" && p.Value.ToString() == "1234");
             Assert.Equal("SELECT * FROM c WHERE c.id = @foo AND c.value = @foo", result);
         }
+
+        [Fact]
+        public void TemplateBind_NestedParameter()
```
brettsam: I'd like to see a test that includes both a nested and a non-nested parameter in the same SqlQuery. You should be able to add something here to check that as well.

src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
anthonychu: @ealsur do you know the complete naming rules for the parameters?

Good point. We can generate something like a GUID for the name to avoid clobbering an existing token. If we do that, as long as it fits under the length limit for a parameter name, it should work.

anthonychu:
@brettsam @MikeStall I think this binding illustrates the need for a bit more information from the `BindingTemplate`. Instead of just the parameter names, it would be nice to be able to iterate over param/token objects that contain info such as the original parameter expression.

test/WebJobs.Extensions.Tests/Extensions/CosmosDB/CosmosDBSqlResolutionPolicyTests.cs
anthonychu: Done

src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
anthonychu: Refactored out the duplicate logic. Hope I didn't make it worse. 😄

anthonychu:
@brettsam @ealsur Anything I can do to help move this PR along? Thanks.

src/WebJobs.Extensions.CosmosDB/CosmosDBSqlResolutionPolicy.cs
ealsur: @anthonychu  Sorry, I missed this for some reason, never saw the notification :( Regarding the rules, I'm not really familiar with all of them but [this one](https://stackoverflow.com/questions/751719/sql-server-invalid-characters-in-parameter-names) looks complete.

anthonychu: Thanks. So I guess Cosmos DB follows SQL rules for param names? Perhaps switching to a GUID is a safer way to go. I'll see if I can do that.

anthonychu:
@brettsam @ealsur Updated to use GUIDs for parameter names to avoid conflicts. Only doing this for nested parameters. Existing behavior should stay the same to avoid accidental breaking changes.

brettsam:
Sorry this fell off my radar! I don't think this is going to tackle all the scenarios that it should, due to the lack of a good API for resolving the template tokens. This works great for headers, but from some of our other code I see that we also support things like JObject which this won't support. Let me follow up offline to see what we can do to reduce some of the code that we need here.

brettsam:
I'm going to go ahead and close this until we can come back with a fix that works for all dotted-notation. See https://github.com/Azure/azure-webjobs-sdk/issues/1726 for the issue being tracked.

PR #369, opened 2018-02-13
Update to fix the Azure Cosmos DB branding

ggailey777:
@brettsam please make sure I didn't make anything less clear WRT Cosmos DB.


ggailey777:
Also, @brettsam  or @jeffhollan are BYOB still in preview? If so, we should make this more clear in the readme and the wiki.

brettsam:
Yikes. Looks good to me. I believe that BYOB is still in preview. We use it everywhere for our bindings but there are some pretty rough edges and it's not fully documented.

brettsam (approved):

PR #372, opened 2018-02-22
Updating WebApi Client package to remove nuget warning

fabiocav:


fabiocav:
This will remove the warnings that generate a lot of confusion today when the extension is referenced:

image

image

PR #373, opened 2018-03-01
Add auth information as an optional parameter in HttpTrigger functions

ConnorMcMahon:
A prototype to access the information that can be found at the /.auth/me endpoint easily as a parameter of the function.

NOTE: The current PR is for v2.x (Functions V1) only. Most of the code I have added in the second commit should work in dev (Functions V2) as well, but the folder structure was different, so files may need to be located in different places, and testing still needs to be done to ensure it works on Functions V2.

Current limitations:

- Doesn't grab information from the key-based authentication that Functions already has
- Requires the function to reference the Microsoft.Azure.WebJobs.Extensions.Http.dll to get access to the new type
- Not robust in handling "mixed" authorization methods (i.e. allowing AND/OR of key-based and EasyAuth-based authorization)

### DIFF
--- a/src/WebJobs.Extensions.Http/AuthenticatedUser.cs
+++ b/src/WebJobs.Extensions.Http/AuthenticatedUser.cs
@@ -1,0 +1,87 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// A class containing information about the current authenticated user.
+    /// </summary>
+    public class AuthenticatedUser
+    {
+        /// <summary>
+        /// The JWT access token for the user
+        /// </summary>
+        [JsonProperty("access_token")]
+        public string AccessToken { get; private set; }
+
+        /// <summary>
+        /// The datetime that the user's access token expires
+        /// </summary>
+        [JsonProperty("expires_on")]
+        public DateTime ExpiresOn { get; private set; }
+
+        /// <summary>
+        /// The JWT token that represents the identity of the user.
+        /// </summary>
+        [JsonProperty("id_token")]
+        public string IdToken { get; private set; }
+
+        /// <summary>
+        /// The string representation of the identity provider for the authenticated user.
+        /// </summary>
+        [JsonProperty("provider_name")]
+        public string ProviderName { get; private set; }
+
+        /// <summary>
+        /// The datetime that the user's access token expires
+        /// </summary>
+        [JsonProperty("refresh_token")]
+        public string RefreshToken { get; private set; }
+
+        [JsonProperty("user_claims")]
+        internal List<Claim> _userClaims { get; set; }
+
+        /// <summary>
+        /// An array of claims that apply to the authenticated user.
+        /// </summary>
+        [JsonIgnoreAttribute]
+        public IReadOnlyList<Claim> UserClaims
+        {
+            get
+            {
+                return this._userClaims.AsReadOnly();
+            }
+        } 
+
+        internal static AuthenticatedUser DeserializeJson(string json)
+        {
+            return JsonConvert.DeserializeObject<List<AuthenticatedUser>>(json, new JsonConverter[] { new ClaimConverter() })[0];
+        }
+
+        private class ClaimConverter : JsonConverter
+        {
+            public override bool CanConvert(Type objectType)
+            {
+                return objectType == typeof(Claim);
+            }
+
+            public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+            {
+                JObject jObject = JObject.Load(reader);
+                return new Claim(jObject["typ"].Value<string>(), jObject["val"].Value<string>());
+            }
+
+            public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+            {
+                throw new NotImplementedException();
+            }
+        }
+    }
+}
--- a/src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
@@ -1,0 +1,168 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    internal class AuthenticatedUserBindingProvider : IBindingProvider
+    {
+        public Task<IBinding> TryCreateAsync(BindingProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            if (context.Parameter.ParameterType != typeof(AuthenticatedUser))
+            {
+                return Task.FromResult<IBinding>(null);
+            }
+
+            return Task.FromResult<IBinding>(new AuthenticatedUserBinding(context.Parameter));
+        }
+
+        private class AuthenticatedUserBinding : IBinding
+        {
+            private readonly ParameterInfo _parameter;
+
+            public AuthenticatedUserBinding(ParameterInfo parameter)
+            {
+                _parameter = parameter;
+            }
+
+            public bool FromAttribute => false;
+
+            public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+            {
+                // TODO: Don't know what to do with this without binding data, or if it 
+                // is even needed in our use case.
+                throw new NotImplementedException("This method doesn't have binding data.");
+            }
+
+            public Task<IValueProvider> BindAsync(BindingContext context)
+            {
+                if (context == null)
+                {
+                    throw new ArgumentNullException("context");
+                }
+
+                return BindAsync(context.BindingData);
+            }
+
+            private static Task<IValueProvider> BindAsync(IReadOnlyDictionary<string, object> bindingData)
+            {
+                return Task.FromResult<IValueProvider>(new AuthenticatedUserValueProvider(bindingData));
+            }
+
+            public ParameterDescriptor ToParameterDescriptor()
+            {
+                return new ParameterDescriptor
+                {
+                    Name = _parameter.Name,
+                    DisplayHints = new ParameterDisplayHints
+                    {
+                        Description = "AuthenticatedUser"
+                    }
+                };
+            }
+
+            private class AuthenticatedUserValueProvider : IValueProvider
+            {
+                private static HttpClient _client = new HttpClient();
+                private IReadOnlyDictionary<string, object> _bindingData;
+
+                public AuthenticatedUserValueProvider(IReadOnlyDictionary<string, object> bindingData)
+                {
+                    _bindingData = bindingData;
+                }
+
+                public Type Type => typeof(AuthenticatedUser);
+
+                public async Task<object> GetValueAsync()
+                {
+                    var initialRequest = _bindingData.Values.FirstOrDefault(val => val.GetType() == typeof(HttpRequestMessage)) as HttpRequestMessage;
+                    if (initialRequest == null)
+                    {
+                        throw new InvalidOperationException("Cannot determine authenticated user without HttpTrigger.");
+                    }
+
+                    var authorizationHeader = GetAuthorizationHeaderValue(initialRequest);
+                    if (authorizationHeader != null)
+                    {
+                        return await GetAuthenticatedUserFromEasyAuth(authorizationHeader);
+                    }
+                    else
+                    {
+                        return GetAuthenticatedUserFromFunctionKey();
+                    }   
+                }
+
+                private static string GetAuthorizationHeaderValue(HttpRequestMessage initialRequest)
+                {
+                    var idTokenHeaders = initialRequest.Headers.Where(header => header.Key.EndsWith("id-token", StringComparison.OrdinalIgnoreCase));
+                    if (!idTokenHeaders.Any())
+                    {
+                        return null;
+                    }
+                    else
+                    {
+                        var idTokenHeader = idTokenHeaders.First();
+                        var idTokenValue = idTokenHeader.Value.First();
+                        return "Bearer " + idTokenValue;
+                    }
+                }
+
+                private async Task<AuthenticatedUser> GetAuthenticatedUserFromEasyAuth(string authorizationHeader)
+                {
+                    string hostname = Environment.GetEnvironmentVariable("WEBSITE_HOSTNAME");
+                    var authUri = "https://" + hostname + "/.auth/me";
+                    HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, authUri);
+                    request.Headers.Add("Authorization", authorizationHeader);
+                    HttpResponseMessage response = await _client.SendAsync(request);
+
+                    if (response.StatusCode == HttpStatusCode.Unauthorized)
+                    {
+                        // TODO: Determine if an exception is the most appropriate use case here
+                        // especially once we start dealing with AND/OR of multiple AuthLevel enums
+                        throw new InvalidOperationException("No user was authenticated.");
+                    }
+                    else if (response.StatusCode == HttpStatusCode.NotFound)
+                    {
+                        //Same concern as the unauthorized case
+                        throw new InvalidOperationException("Authentication/Authorization is not enabled.");
+                    }
+                    else if (response.IsSuccessStatusCode)
+                    {
+                        string authenticatedUserJson = await response.Content.ReadAsStringAsync();
+                        return AuthenticatedUser.DeserializeJson(authenticatedUserJson);
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Do not know how to handle response status code {response.StatusCode} from {authUri}");
+                    }
+                }
+
+                private static AuthenticatedUser GetAuthenticatedUserFromFunctionKey()
+                {
+                    throw new NotImplementedException("TODO: Have not worked out what information I can/should populate here.");
+                }
+
+                public string ToInvokeString()
+                {
+                    //TODO: what goes here?
+                    return string.Empty;
+                }
+            }
+        }
+    }
+}
--- a/src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
@@ -1,0 +1,167 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// This provider provides a binding to Type <see cref="ClaimsIdentity"/>.
+    /// </summary>
+    /// <remarks>
+    /// Attempts to bind to an identity in precedence order based on <see cref="AuthorizationLevel"/> levels.
+    /// I.e. if the request is key authenticated (Function/System/Admin) that identity is returned, if the
+    /// request is user authenticated (User) that identity is returned.
+    /// </remarks>
+    internal class ClaimsIdentityBindingProvider : IBindingProvider
+    {
+        public Task<IBinding> TryCreateAsync(BindingProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            if (context.Parameter.ParameterType != typeof(ClaimsIdentity))
+            {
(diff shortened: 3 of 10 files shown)


src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,167 @@
...
+                if (context.BindingData.TryGetValue(HttpTriggerAttributeBindingProvider.HttpHeadersKey, out value))
+                {
+                    IDictionary<string, string> headers = (IDictionary<string, string>)value;
+                    if (headers != null)
+                    {
+                        headers.TryGetValue(HttpConstants.AntaresEasyAuthProviderHeaderName, out identityProvider);
+                    }  
+                }
+
+                var identity = GetPrimaryIdentity(ClaimsPrincipal.Current, identityProvider);
```
cgillum: `ClaimsPrincipal.Current` won't work in Functions v2. I suggest we avoid the dependency on `System.Security.Claims` entirely.

ConnorMcMahon: Yeah, this was mentioned in the previous code review. I haven't removed the Claims based Identity providers yet from Matthews original work.

---
In reply to: [171723801](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171723801) [](ancestors = 171723801)

src/WebJobs.Extensions.Http/AuthenticatedUser.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// A class containing information about the current authenticated user.
+    /// </summary>
+    public class AuthenticatedUser
```
mathewc: For reference, my prototype PRs were here:

- https://github.com/Azure/azure-webjobs-sdk-extensions/pull/262
- https://github.com/Azure/azure-functions-host/pull/1747

I was trying to avoid adding new types to represent users/claims, and instead use existing .NET Types, since this is all representable as core identity Types.

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                    {
+                        var idTokenHeader = idTokenHeaders.First();
+                        var idTokenValue = idTokenHeader.Value.First();
+                        return "Bearer " + idTokenValue;
+                    }
+                }
+
+                private async Task<AuthenticatedUser> GetAuthenticatedUserFromEasyAuth(string authorizationHeader)
+                {
+                    string hostname = Environment.GetEnvironmentVariable("WEBSITE_HOSTNAME");
```
mathewc: @cgillum and I had discussed in my original PR a way to have EA flow the claims as headers to us so we don't have to make an additional loop back request.

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                if (context == null)
+                {
+                    throw new ArgumentNullException("context");
+                }
+
+                // if the request is EasyAuth authenticated, get the identity provider type
+                object value = null;
+                string identityProvider = null;
+                if (context.BindingData.TryGetValue(HttpTriggerAttributeBindingProvider.HttpHeadersKey, out value))
```
cgillum: It looks like you only use `value` inside the `if` block. In that case, it's better to use one of the new C# features and declare it as part of the out parameter, like this:

```csharp
if (context.BindingData.TryGetValue(HttpTriggerAttributeBindingProvider.HttpHeadersKey, out object value))
```

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,167 @@
...
+                        Description = "ClaimsIdentity"
+                    }
+                };
+            }
+
+            private static ClaimsIdentity GetPrimaryIdentity(ClaimsPrincipal claimsPrincipal, string identityProvider = null)
+            {
+                // if the principal contains a key based identity that will take precedence
+                // over any other identity
+                var identity = claimsPrincipal.Identities.LastOrDefault(p => p.IsAuthenticated && string.Compare(p.AuthenticationType, "key", StringComparison.OrdinalIgnoreCase) == 0);
```
cgillum: Why `string.Compare` and not `string.Equals` (here and in the logic further below)?

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,167 @@
...
+
+                public Task<object> GetValueAsync()
+                {
+                    return Task.FromResult<object>(_identity);
+                }
+
+                public string ToInvokeString()
+                {
+                    // TODO: figure out right value here
+                    return ClaimsPrincipal.Current.ToString();
```
cgillum: Probably `ClaimsPrincipal.Identity.Name` is what you want.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+        public Task<IBinding> TryCreateAsync(BindingProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            if (context.Parameter.ParameterType != typeof(ClaimsPrincipal))
+            {
+                return Task.FromResult<IBinding>(null);
```
cgillum: minor perf nit: There should probably be a cache version of the null task result to avoid unnecessary task object allocations. It seems we do this in many places.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                public Task<object> GetValueAsync()
+                {
+                    return Task.FromResult<object>(ClaimsPrincipal.Current);
+                }
+
+                public string ToInvokeString()
+                {
+                    // TODO: figure out right value here
+                    return ClaimsPrincipal.Current.ToString();
```
cgillum: Probably `ClaimsPrincipal.Identity.Name` is what you want.

test/WebJobs.Extensions.Tests/Extensions/Http/HttpTriggerEndToEndTests.cs
```diff
@@ -104,6 +105,15 @@ public void BindToRequestAndPocoAtSameTime()
             Assert.Equal(request.Properties["result"], "abc"); // Verify resposne was set
         }
 
+        [Fact]
+        public void ClaimsPrincipalBinding()
+        {
+            var request = new HttpRequestMessage(HttpMethod.Get, "http://functions.com/api/TestIdentityBindings");
+
+            var method = typeof(TestFunctions).GetMethod("TestIdentityBindings");
```
cgillum: Instead of hardcoding "TestIdentityBindings", it's more robust to do `nameof(TestIdentityBindings)`.

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Do not know how to handle response status code {response.StatusCode} from {authUri}");
+                    }
+                }
+
+                private static AuthenticatedUser GetAuthenticatedUserFromFunctionKey()
+                {
+                    throw new NotImplementedException("TODO: Have not worked out what information I can/should populate here.");
```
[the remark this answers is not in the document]
ConnorMcMahon: It grabs them off of ClaimsPrincipal.Current I believe, which is null in .NET Core I believe.

---
In reply to: [171932417](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171932417) [](ancestors = 171932417)

src/WebJobs.Extensions.Http/AuthenticatedUser.cs
ConnorMcMahon: So  ClaimsIdentityBindingProvider and ClaimsPrincipalBindingProvider were both cherry-picked in the commit from that first PR you posted. This new object contains a superset of those binding providers, and is non-.NET specific as Chris mentioned below.

However, adding this type does brings about it's own challenges, such as how to import the type itself, as well as maintenance/documentation. I am open to discussing ways around defining a new type, while still allowing cross language support and ideally providing access to more information than just the claims.

---
In reply to: [171758615](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171758615) [](ancestors = 171758615)

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                    if (response.StatusCode == HttpStatusCode.Unauthorized)
+                    {
+                        // TODO: Determine if an exception is the most appropriate use case here
+                        // especially once we start dealing with AND/OR of multiple AuthLevel enums
+                        throw new InvalidOperationException("No user was authenticated.");
+                    }
+                    else if (response.StatusCode == HttpStatusCode.NotFound)
+                    {
+                        //Same concern as the unauthorized case
+                        throw new InvalidOperationException("Authentication/Authorization is not enabled.");
```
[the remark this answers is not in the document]
ConnorMcMahon: Good point about the token store. For now, we have all of the required fields except for the claims (and we want to add those in a header soon anyways). Once we have all of the fields present, then this can become the only method used to populate, and we avoid hitting the token store altogether.

---
In reply to: [171931203](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171931203) [](ancestors = 171931203)

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                    var authUri = "https://" + hostname + "/.auth/me";
+                    HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, authUri);
+                    request.Headers.Add("Authorization", authorizationHeader);
+                    HttpResponseMessage response = await _client.SendAsync(request);
+
+                    if (response.StatusCode == HttpStatusCode.Unauthorized)
+                    {
+                        // TODO: Determine if an exception is the most appropriate use case here
+                        // especially once we start dealing with AND/OR of multiple AuthLevel enums
+                        throw new InvalidOperationException("No user was authenticated.");
```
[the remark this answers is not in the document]
ConnorMcMahon: I'm not quite sure how best to represent an unauthenticated user identity using our current workflow. The object itself as null? An added bool/enum property on the object to check if it is authenticated?

---
In reply to: [171930399](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171930399) [](ancestors = 171930399)

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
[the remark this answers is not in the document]
cgillum: Functions API keys are not integrated into the IIS/ASP.NET identity pipeline, so I'm pretty sure they could not have been pulled off of ClaimsIdentity.Current.

---
In reply to: [171945399](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171945399) [](ancestors = 171945399,171932417)

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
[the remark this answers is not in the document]
cgillum: In ASP.NET today, the "User" object is never null.  Instead, it is initialized to some default values and the IsAuthenticated property is set to false.  I'm suggesting we follow that general model of always providing a non-null user object and use a boolean flag to indicate whether the request was authenticated. We can use that same data structure to carry additional information, such as an enum and maybe a informative message for debugging purposes.

---
In reply to: [171948363](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171948363) [](ancestors = 171948363,171930399)

src/WebJobs.Extensions.Http/AuthenticatedUserBindingProvider.cs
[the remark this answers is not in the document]
ConnorMcMahon: I'm specifically looking at the first few lines of GetPrimaryIdentity() in ClaimsIdentityBindingProvider. I could be confused about what keys that comment was referring to. @mathewc@microsoft.com, thoughts?

---
In reply to: [171972359](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171972359) [](ancestors = 171972359,171945399,171932417)

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,206 @@
...
+{
+    /// <summary>
+    /// This provider provides a binding to Type <see cref="ClaimsIdentity"/>.
+    /// </summary>
+    /// <remarks>
+    /// Attempts to bind to an identity in precedence order based on <see cref="AuthorizationLevel"/> levels.
+    /// I.e. if the request is key authenticated (Function/System/Admin) that identity is returned, if the
+    /// request is user authenticated (User) that identity is returned.
+    /// </remarks>
+    internal class ClaimsIdentityBindingProvider : IBindingProvider
```
ConnorMcMahon: I'm not sold on the utility of this binding. The IEnumerable<ClaimsIdentity> can be gotten straight off of the identity, and us choosing one of them to bind to in the single ClaimsIdentity case just seems like a breaking change waiting to happen.

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
```diff
@@ -0,0 +1,206 @@
...
+                    }
+                }
+                return null;
+            }
+
+            private static ClaimsIdentity GetPrimaryIdentity(ClaimsPrincipal claimsPrincipal, string identityProvider = null)
+            {
+                // if the principal contains a key based identity that will take precedence
+                // over any other identity
+                var identity = claimsPrincipal.Identities.LastOrDefault(p => p.IsAuthenticated && string.Equals(p.AuthenticationType, "key", StringComparison.OrdinalIgnoreCase));
```
ConnorMcMahon: As Chris mentioned earlier, I don't think the "key" identity is present here, even when using key based authentication for an HTTP function.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// This provider provides a binding to Type <see cref="ClaimsPrincipal"/>.
+    /// </summary>
+    internal class ClaimsPrincipalBindingProvider : IBindingProvider
+    {
+        private static Task<IBinding> nullBinding = Task.FromResult<IBinding>(null);
```
cgillum: >static [](start = 16, length = 6)

style nit: Make this readonly and capitalize it

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
cgillum: Do we need to check for null here?

---
In reply to: [171935462](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r171935462) [](ancestors = 171935462)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+                    throw new ArgumentNullException("context");
+                }
+
+                IReadOnlyDictionary<string, object> bindingData = context.BindingData;
+                return Task.FromResult<IValueProvider>(new ClaimsPrincipalValueProvider(ClaimsPrincipalHelper.FromBindingData(bindingData)));
+            }
+
+            public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+            {
+                throw new NotImplementedException("This method does not provide the necessary binding data.");
```
cgillum: >NotImplementedException [](start = 26, length = 23)

Can you add a comment about when this overload would get called, and why it's safe for us to assume that it won't ever get called?

src/WebJobs.Extensions.Http/ClaimsPrincipalHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+    using System.Security.Claims;
+
+    internal static class ClaimsPrincipalHelper
+    {
+        // Shared serializer instance. This is safe for multi-threaded use.
+        private static readonly DataContractJsonSerializer Serializer = GetCustomSerializer();
+
+        public static ClaimsPrincipal FromBindingData(IReadOnlyDictionary<string, object> bindingData)
+        {
+            var req = bindingData.Values.FirstOrDefault(val => val.GetType() == typeof(HttpRequestMessage)) as HttpRequestMessage;
```
cgillum: >HttpRequestMessage [](start = 87, length = 18)

Does functions v2 also use HttpRequestMessage internally (and not the new HttpRequest in .NET Core)?

src/WebJobs.Extensions.Http/ClaimsPrincipalHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (req == null)
+            {
+                return new ClaimsPrincipal();
+            }
+
+            var claimsPrincipalHeaderValue = GetClaimsPrincipalHeaderValue(req);
```
cgillum: >var [](start = 12, length = 3)

Please use variable type names explicitly when saving the result of a method call (in this case string).

src/WebJobs.Extensions.Http/ClaimsPrincipalHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            // Use the ISO 8601 recommended format for internet timestamps (see http://tools.ietf.org/html/rfc3339).
+            settings.DateTimeFormat = new DateTimeFormat("o");
+
+            return new DataContractJsonSerializer(typeof(ClaimsIdentitySlim), settings);
+        }
+
+
+        private static string GetClaimsPrincipalHeaderValue(HttpRequestMessage request)
+        {
+            var claimsPrincipalHeaders = request.Headers.Where(header => string.Equals(header.Key, "x-ms-client-principal", StringComparison.OrdinalIgnoreCase));
```
cgillum: >Where [](start = 57, length = 5)

Would it be simpler to use `FirstOrDefault` instead of `Where` here?

ConnorMcMahon: I was tempted to do that, but since a KeyValuePair  is a struct, you can't compare against default(KeyValuePair) easily. See this stackoverflow thread [https://stackoverflow.com/questions/1335419/how-can-i-make-sure-that-firstordefaultkeyvaluepair-has-returned-a-value/1336077](https://stackoverflow.com/questions/1335419/how-can-i-make-sure-that-firstordefaultkeyvaluepair-has-returned-a-value/1336077). 

However, I think if I can get around that by first performing a Select(header => header.Value.First()) before FirstOrDefault to achieve a cleaner result than what I currently have.

---
In reply to: [174632398](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r174632398) [](ancestors = 174632398)

src/WebJobs.Extensions.Http/ClaimsPrincipalHelper.cs
ConnorMcMahon: Good point. I believe in functions v2, either HttpRequestMessage or HttpRequest can be utilized as this type. However, this current PR is for function v1 (which uses webjobs v2), and targets .NET 4.6, so I don't think I even have access to this type.

---
In reply to: [174631976](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r174631976) [](ancestors = 174631976)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
ConnorMcMahon: Interestingly, until I added readonly, the project wouldn't build because of a style error if I used uppercase with a private error. Good to know adding readonly avoids that. 

---
In reply to: [174630304](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r174630304) [](ancestors = 174630304)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
ConnorMcMahon: After talking with Mike about this, it appears this is used when a parameter of the type ClaimsPrincipal is provided in a programatic call (i.e. via a normal WebJobs). We technically don't need an implementation, but I can right a simple one just to be safe.

---
In reply to: [174631670](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/373#discussion_r174631670) [](ancestors = 174631670)

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
mathewc: I believe my initial idea in having this binding in addition to the principal binding was to facilitate bindings in other languages (e.g. Node), where the binding is JObject based and the ClaimsPrincipal APIs aren't being used. If we verify the programming model for both .NET and Node can naturally access all identities (in cases where there are multiple) and enumerate them, then I agree we should do the minimum required.

src/WebJobs.Extensions.Http/ClaimsIdentityBindingProvider.cs
mathewc: Right - that was additional prototype work that I had done in a separate PR to add that identity, to ensure a consistent experience.

PR #380, opened 2018-03-09
Adding tracing inside CosmosDBTrigger to log StopAsync exceptions

### LINKED ISSUE #379: [CosmosDBTrigger] If Stop() is called before Start(), the trigger can throw an exception
During testing of something unrelated, I saw a stack with:

```
   at Microsoft.Azure.Documents.ChangeFeedProcessor.ChangeFeedEventHost.<StopAsync>d__7a.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.Documents.ChangeFeedProcessor.ChangeFeedEventHost.<UnregisterObserversAsync>d__6.MoveNext()
```

I think it's happening b/c the host is calling Start/Stop very quickly, which means that Stop is actually being called first. It looks like internally the ChangeFeedProcessor may throw in that case. We should either catch that  and ignore it, or check if it's able to be Unregistered (if that's possible).

ealsur:
This change adds tracing inside the CosmosDBTrigger and handles the case where StopAsync might be called before Start.

This PR fixes #379

### DIFF
--- a/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
         private readonly TraceWriter trace;
         private readonly DocumentCollectionInfo monitorCollection;
         private readonly DocumentCollectionInfo leaseCollection;
-        private bool listernerStarted = false;
+        private bool listenerStarted = false;
 
         public CosmosDBTriggerListener(ITriggeredFunctionExecutor executor, DocumentCollectionInfo documentCollectionLocation, DocumentCollectionInfo leaseCollectionLocation, ChangeFeedHostOptions leaseHostOptions, int? maxItemCount, TraceWriter trace)
         {
@@ -58,15 +58,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 
         public async Task StartAsync(CancellationToken cancellationToken)
         {
-            if (this.listernerStarted)
+            if (this.listenerStarted)
             {
-                return;
+                throw new InvalidOperationException("The listener has already been started.");
             }
 
             try
             {
                 await this.host.RegisterObserverFactoryAsync(this);
-                this.listernerStarted = true;
+                this.listenerStarted = true;
             }
             catch (DocumentClientException ex) when (ex.StatusCode == HttpStatusCode.NotFound)
             {
@@ -80,14 +80,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
         {
             try
             {
-                if (this.host != null && this.listernerStarted)
+                if (this.host != null && this.listenerStarted)
                 {
                     await this.host.UnregisterObserversAsync();
+                    this.listenerStarted = false;
                 }
             }
             catch (Exception ex)
             {
-                this.trace.Error("Stopping the observer failed, potentially it was never started.", ex);
+                this.trace.Warning($"Stopping the observer failed, potentially it was never started. Exception: {ex.Message}.");
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -10,23 +10,27 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
...
     using Microsoft.Azure.WebJobs.Host.Listeners;
 
     internal class CosmosDBTriggerListener : IListener, IChangeFeedObserverFactory
     {
         private readonly ITriggeredFunctionExecutor executor;
         private readonly ChangeFeedEventHost host;
+        private readonly TraceWriter trace;
         private readonly DocumentCollectionInfo monitorCollection;
         private readonly DocumentCollectionInfo leaseCollection;
+        private bool listernerStarted = false;
```
brettsam: nit: spelling

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -66,9 +76,19 @@ public void Dispose()
...
+            try
+            {
+                if (this.host != null && this.listernerStarted)
+                {
+                    await this.host.UnregisterObserversAsync();
+                }
+            }
+            catch (Exception ex)
+            {
+                this.trace.Error("Stopping the observer failed, potentially it was never started.", ex);
```
brettsam: Since this doesn't really bring anything down -- maybe we call this a Warning?

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -10,23 +10,27 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
...
     using Microsoft.Azure.Documents.ChangeFeedProcessor;
     using Microsoft.Azure.Documents.Client;
+    using Microsoft.Azure.WebJobs.Host;
     using Microsoft.Azure.WebJobs.Host.Executors;
     using Microsoft.Azure.WebJobs.Host.Listeners;
 
     internal class CosmosDBTriggerListener : IListener, IChangeFeedObserverFactory
     {
         private readonly ITriggeredFunctionExecutor executor;
         private readonly ChangeFeedEventHost host;
```
brettsam: Is `ChangeFeedEventHost` mockable in any way? It'd be nice to have some tests to make sure this works, but that may involve creating a wrapper around this (like I did in the other bindings for `DocumentClient`). Since the logic is fairly simple, it may not be worth the effort...

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -54,9 +58,15 @@ public void Dispose()
 
         public async Task StartAsync(CancellationToken cancellationToken)
         {
+            if (this.listernerStarted)
+            {
+                return;
```
mathewc: A listener that starts successfully should never be started again really. The various triggers aren't all consistent in their handling of this, however some of them (e.g. [ServiceBus](https://github.com/Azure/azure-webjobs-sdk/blob/a3153829592743b07239a9077318c8e2d4e78f62/src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs#L40)) throw in this case. The host won't call Start twice on a listener that successfully started. Your call.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
ealsur: In the new version of the Host we are working on, it is, we have a whole complete set of Unit tests on the library. This version however, is not completely mockable :(

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
ealsur: Sorry, my english sucks sometimes.

src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs
ealsur: Done

PR #381, opened 2018-03-09
Moving to latest SDK version + fixing ErrorTrigger

mathewc:
Moving to the latest SDK version. Also fixing the ErrorTrigger issue caused by the fact that the TraceWriter collection becomes read-only early in startup.

Because the config.Tracing.Tracers collection is snapshotted early in the startup process, the collection can't be modified after indexing. However, ErrorTrigger was previously adding its custom monitors as part of listener startup, which is AFTER the snapshot. With my changes, we add a single TraceWriter early at the right time before the snapshot, and later during indexing we add our monitors to that single instance internally.


test/WebJobs.Extensions.Tests/Extensions/DocumentDB/DocumentDBEndToEndTests.cs
```diff
@@ -287,6 +287,7 @@ private Task RunTestAsync(string testName, IDocumentDBServiceFactory factory, Tr
             JobHost host = new JobHost(config);
 
             await host.StartAsync();
+            testTrace.Events.Clear();
```
mathewc: My approach for dealing with the new Warning error messages was just to modify these test helper methods to clear all startup logs before invoking the function. That's what the tests are expecting anyways - they only want invocation logs, not host init/startup logs.

brettsam: Sounds good!

PR #393, opened 2018-03-26
Fix problem with daylight saving time

ToastHawaii:
Fix a problem with daliy scheduler in a enviroment with a timezone that has daylight saving time. 
Problem description: 
I used a cron pattern like this [TimerTrigger("0 30 23 * * *")] /* 23:30 every day */
On a Azure Web App Service that has the Timezone set to Europe Standard Time. 
The run at 24. March starts as expected  at 24. March 23:30.
The problem occurs on the next run it starts on 26. March at 00:30 instate 25. March at 23:30.


mathewc:
Thanks for taking the time to send a PR for this. I wasn't aware of this PR and actually published my own https://github.com/Azure/azure-webjobs-sdk-extensions/pull/395 which is effectively the same as yours, though a little more explicit in that it doesn't rely on cast conversions and it's explicit that it is using the Local timezone. I also added some tests.

I did run my tests with your one liner fix as well and it worked as well :)

ToastHawaii:
Sounds good to me. As long as it works. :)

mathewc:
Thanks again - funny how several people all discovered this issue on our recent DST transition and were discussing this and working on it in parallel :)

Issue #387, opened 2018-03-21
Cosmos DB Connection Mode configuration setting


ealsur:
Creating this issue so we can track the requirement of configuring the Cosmos DB client creation `ConnectionMode`. 

Right now it uses the default, but the user might need to change it for performance or connectivity reasons.

Possible values are in an enum: `ConnectionMode.Gateway` and  `ConnectionMode.Direct`.

davetheunissen:
+1, we've had to do this to get round our internal corporate proxy.

mathewc:
Can this issue be closed now? Looks like the change has been merged.

PR #400, opened 2018-03-30
Exposing ConnectionMode in DocumentDBConfiguration

ealsur:
This change adds `ConnectionMode` to the `DocumentDBConfiguration` as a string (compatible with JSON files) that lets the user define a customized value for the `ConnectionMode` in the `DocumentClient` instances used in the Bindings and the Trigger.

Related to: https://github.com/Azure/azure-webjobs-sdk-extensions/issues/387


src/WebJobs.Extensions.DocumentDB/Config/IDocumentDBServiceFactory.cs
```diff
@@ -5,6 +5,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 {
     internal interface IDocumentDBServiceFactory
     {
-        IDocumentDBService CreateService(string connectionString);
+        IDocumentDBService CreateService(string connectionString, string desiredConnectionMode);
```
brettsam: Nit: can you make this take the enum? That way services don't have to do the conversion themselves.

ealsur: Sure thing!

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -46,6 +46,11 @@ public DocumentDBConfiguration()
         /// </summary>
         public ChangeFeedHostOptions LeaseOptions { get; set; }
 
+        /// <summary>
+        /// Gets or sets the ConnectionMode used in the DocumentClient instances.
+        /// </summary>
+        public string ConnectionMode { get; set; }
```
brettsam: Chatted with @fabiocav. Let's have this be the `ConnectionMode` enum instead of string. Then we can handle the parsing and conversion here; https://github.com/Azure/azure-functions-host/blob/v1.x/src/WebJobs.Script/Binding/DocumentDBScriptBindingProvider.cs#L47

ealsur: Got it, will update the PR during the day.

ealsur:
PR updated with the refactor using the enum

src/WebJobs.Extensions.DocumentDB/Config/DefaultDocumentDBServiceFactory.cs
```diff
@@ -1,13 +1,15 @@
...
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.Documents.Client;
+
 namespace Microsoft.Azure.WebJobs.Extensions.DocumentDB
 {
     internal class DefaultDocumentDBServiceFactory : IDocumentDBServiceFactory
     {
-        public IDocumentDBService CreateService(string connectionString)
+        public IDocumentDBService CreateService(string connectionString, ConnectionMode? desiredConnectionMode)
```
brettsam: nit -- just use `connectionMode` as the parameter name here (and elsewhere)

ealsur:
Created the second required PR https://github.com/Azure/azure-functions-host/pull/2782

src/WebJobs.Extensions.DocumentDB/Services/DocumentDBService.cs
```diff
@@ -15,10 +15,14 @@ internal sealed class DocumentDBService : IDocumentDBService, IDisposable
...
         private DocumentClient _client;
 
-        public DocumentDBService(string connectionString)
+        public DocumentDBService(string connectionString, ConnectionMode? connectionMode)
         {
             DocumentDBConnectionString connection = new DocumentDBConnectionString(connectionString);
             _client = new DocumentClient(connection.ServiceEndpoint, connection.AuthKey);
+            if (connectionMode.HasValue)
+            {
+                _client.ConnectionPolicy.ConnectionMode = connectionMode.Value;
```
brettsam: Can you add a comment for what this defaults to if we don't set it?

src/WebJobs.Extensions.DocumentDB/Config/DocumentDBConfiguration.cs
```diff
@@ -46,6 +46,11 @@ public DocumentDBConfiguration()
         /// </summary>
         public ChangeFeedHostOptions LeaseOptions { get; set; }
 
+        /// <summary>
+        /// Gets or sets the ConnectionMode used in the DocumentClient instances.
+        /// </summary>
+        public ConnectionMode? ConnectionMode { get; set; }
```
brettsam: Another option is to remove the nullable here and explicitly choose the default (rather than letting the SDK choose it). Do you think that'd be more readable? It'd remove the requirement for checking HasValue, etc.

ealsur: I'd avoid having a default defined by us, the SDK already has a default value, and if that ever changes, we would be overriding the normal behavior?

brettsam: I was actually thinking we wanted to make sure the WebJobs Extension had the same default behavior, no matter what the Cosmos SDK did. But you're probably right, it's okay to do it this way. Updates to new versions will need to be explicit anyway in v2, so they'll need to consult default changes if there are any.

brettsam:
Squash/Rebase (there may be changes that you're missing) and we can get this in once we see AppVeyor pass.

ealsur:
Rebased/Squashed

brettsam:
It looks like nuget added a new switch so using `-o` no longer works. Since we grab the latest build, it's breaking us randomly. I'll fix it, then you can rebase on that.

brettsam:
Okay, try rebasing on v2.x now.

ealsur:
We are green 😄

PR #404, opened 2018-04-16
Add ClaimsPrincipal binding

ConnorMcMahon:
The WebJobs side of the work for closer EasyAuth integration with Azure Functions.


src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        // Shared serializer instance. This is safe for multi-threaded use.
+        private static readonly DataContractJsonSerializer Serializer = GetCustomSerializer();
+
+        /// <summary>
+        /// Retrieves a serialized ClaimsIdentity object from the http request of an HTTP Trigger
+        /// </summary>
+        /// <param name="request">The request message from the HTTP Trigger</param>
+        /// <param name="identityHeaderName">The name of the header the identity is stored on.</param>
+        /// <returns></returns>
```
cgillum: >/// <returns></returns> [](start = 8, length = 23)

Empty `<returns>` comment

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        public static ClaimsIdentity GetIdentityFromHttpRequest(HttpRequest request, string identityHeaderName)
+        {
+            if (request == null)
+            {
+                return null;
```
cgillum: >return null; [](start = 16, length = 12)

Would it make more sense to `throw ArgumentException(nameof(request))` instead of returning `null`?

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <param name="request">The HTTP request to add the header to</param>
+        /// <param name="identity">The ClaimsIdentity to be serialized</param>
+        /// <param name="headerName">The name of the header to add the serialized value to</param>
+        public static void AddIdentityToHttpRequest(HttpRequest request, ClaimsIdentity identity, string headerName)
+        {
+            if (request == null)
+            {
+                return;
+            }
```
cgillum: Same comment about throwing instead of returning null here.

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+            using (var buffer = new MemoryStream(Convert.FromBase64String(payload)))
+            {
+                ClaimsIdentitySlim slim = (ClaimsIdentitySlim)Serializer.ReadObject(buffer);
+                return slim.ToClaimsIdentity();
+            }
+        }
+
+        private static DataContractJsonSerializer GetCustomSerializer()
+        {
+            // These settings make DCJS mostly compatible with Json.NET defaults.
```
cgillum: >// These settings make DCJS mostly compatible with Json.NET defaults. [](start = 12, length = 69)

I think the important thing to call out here is that these are the same DCJS settings used by Easy Auth.

src/WebJobs.Extensions.Http/ClaimsIdentitySlim.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Security.Claims;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// Light-weight representation of a <see cref="ClaimsIdentity"/> object.
+    /// </summary>
+    [DataContract]
+    [KnownType(typeof(ClaimSlim))]
+    internal struct ClaimsIdentitySlim
```
cgillum: >ClaimsIdentitySlim [](start = 20, length = 18)

Add a comment that this needs to be kept in-sync with the Easy Auth version of this same code.

src/WebJobs.Extensions.Http/ClaimSlim.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Runtime.Serialization;
+using System.Security.Claims;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// Light-weight representation of a <see cref="Claim"/> object.
+    /// </summary>
+    [DataContract]
+    internal struct ClaimSlim
```
cgillum: >ClaimSlim [](start = 20, length = 9)

Add a comment that this needs to be kept in-sync with the Easy Auth version of this same code.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+                return Task.FromResult<IValueProvider>(valueProvider);
+            }
+
+            private static ClaimsPrincipal GetClaimsPrincipalFromBindingData(IReadOnlyDictionary<string, object> bindingData)
+            {
+                object headers;
+                if (bindingData.TryGetValue(HttpTriggerAttributeBindingProvider.HttpHeadersKey, out headers))
+                {
+                    List<ClaimsIdentity> identities = (headers as IDictionary<string, string>)
+                        .Where(header => PotentialIdentityHeaders.Any(idHeader => string.Equals(idHeader, header.Key, StringComparison.OrdinalIgnoreCase)))
```
cgillum: >PotentialIdentityHeaders [](start = 41, length = 24)

Why isn't this simply `PotentialIdentityHeaders.Contains(header.Key)? It seems that would be more efficient. You can achieve case-insensitivity by using a case-insensitive HashSet<string> above.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// This provider provides a binding to Type <see cref="ClaimsPrincipal"/>.
+    /// </summary>
+    internal class ClaimsPrincipalBindingProvider : IBindingProvider
+    {
+        private static readonly Task<IBinding> NullBinding = Task.FromResult<IBinding>(null);
+
+        private static readonly IReadOnlyCollection<string> PotentialIdentityHeaders = new HashSet<string>
```
cgillum: >IReadOnlyCollection<string> [](start = 32, length = 27)

This should really be a `HashSet<string>` and not an `IReadOnlyCollection<string>`. Otherwise you lose the benefits of using a HashSet. You should also use the constructor overload which takes a case-insensitive `StringComparer`.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+                else
+                {
+                    return new ClaimsPrincipal();
+                }
+            }
+
+            public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+            {
+                var request = value as ClaimsPrincipal;
```
cgillum: >request [](start = 20, length = 7)

Why is this variable named `request` and not `principal`? `request` sounds like an HTTP request, and that confused me when I was trying to understand what this code was doing.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+                if (request != null)
+                {
+                    var binding = new SimpleValueProvider(typeof(ClaimsPrincipal), request, "request");
+                    return Task.FromResult<IValueProvider>(binding);
+                }
+                throw new InvalidOperationException("Value must be of type " + typeof(ClaimsPrincipal).ToString());
```
cgillum: >Value [](start = 53, length = 5)

What does "Value" represent? Is it the name of a parameter? I suggest we make this exception message more precise to ensure the developer understands what they did wrong.

test/WebJobs.Extensions.Tests/Extensions/Http/ClaimsPrincipalEndToEndTests.cs
```diff
@@ -0,0 +1,206 @@
...
+      ""val"": """ + KeyLevelClaimValue + @"""
+    }
+  ],
+  ""name_typ"": ""http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name"",
+  ""role_typ"": ""http://schemas.microsoft.com/ws/2008/06/identity/claims/role""
+}";
+
+        private const string OutputTestPath = @"webjobs_extensionstests\filebindinge2e_output";
+
+        private static ClaimsPrincipal globalPrincipal;
```
cgillum: >globalPrincipal [](start = 39, length = 15)

Is it guaranteed that these tests only run one-at-a-time?

test/WebJobs.Extensions.Tests/Extensions/Http/ClaimsPrincipalEndToEndTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            var request = HttpTestHelpers.CreateHttpRequest("GET", "http://functions.com/api/TestIdentityBindings");
+            string easyAuthHeaderValue = Convert.ToBase64String(Encoding.UTF8.GetBytes(UserIdentityJson));
+            request.Headers.Add("x-ms-client-principal", easyAuthHeaderValue);
+
+            var method = typeof(TestFunctions).GetMethod(nameof(TestFunctions.HttpRequest));
+            _host.Call(method, new { req = request });
+
+            Assert.Equal(Name, globalPrincipal.Identity.Name);
+            Assert.Equal(UserAuthType, globalPrincipal.Identity.AuthenticationType);
+            Claim testClaim = globalPrincipal.Claims.Where(claim => string.Equals(claim.Type, UserClaimType)).First();
```
cgillum: >Where [](start = 53, length = 5)

I suggest doing `FirstOrDefault(...)` instead of `Where(...).First()`. That will make the cause of a test failure more clear since it will allow you to fail at the `Assert.Equal` instead of having failing with a non-descript `InvalidOperationException`.

test/WebJobs.Extensions.Tests/Extensions/Http/ClaimsPrincipalEndToEndTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            var request = HttpTestHelpers.CreateHttpRequest("GET", "http://functions.com/api/TestIdentityBindings");
+            string easyAuthHeaderValue = Convert.ToBase64String(Encoding.UTF8.GetBytes(UserIdentityJson));
+            request.Headers.Add("x-ms-client-principal", easyAuthHeaderValue);
+            string keyHeaderValue = Convert.ToBase64String(Encoding.UTF8.GetBytes(KeyIdentityJson));
+            request.Headers.Add("x-ms-functions-key-identity", keyHeaderValue);
+
+            var method = typeof(TestFunctions).GetMethod(nameof(TestFunctions.HttpRequest));
+            _host.Call(method, new { req = request });
+
+            Assert.Equal(Name, globalPrincipal.Identity.Name);
```
cgillum: >Assert.Equal(Name, globalPrincipal.Identity.Name); [](start = 12, length = 50)

I suggest calling out in a comment that the user identity is expected to take precedence over the key identity.

test/WebJobs.Extensions.Tests/Extensions/Http/ClaimsPrincipalEndToEndTests.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        [Fact]
+        public void ClaimsPrincipalBindingWithNoIdentities()
+        {
+            var request = HttpTestHelpers.CreateHttpRequest("GET", "http://functions.com/api/TestIdentityBindings");
+            var method = typeof(TestFunctions).GetMethod(nameof(TestFunctions.HttpRequest));
+            _host.Call(method, new { req = request });
+
+            Assert.Null(globalPrincipal.Identity);
```
cgillum: >Assert.Null(globalPrincipal.Identity); [](start = 12, length = 38)

This seems inconsistent with how ASP.NET works. In ASP.NET, failing to authenticate results in a principal with `Identity.IsAuthenticated == false`. I suggest we do the same thing with the claims principal binding for consistency.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
ConnorMcMahon: I had no idea that existed. That was exactly what I was going for!

---
In reply to: [181944504](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r181944504) [](ancestors = 181944504)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
ConnorMcMahon: Not sure what happened here. Likely was taking inspiration from HttpTriggerAttributeBindingProvider and forgot to change variable name.

---
In reply to: [181944940](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r181944940) [](ancestors = 181944940)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
ConnorMcMahon: I'm gonna take some additional inspriation from HttpTriggerAttributeBindingProvider and make this a NotSupportedException, and make the error message more clear.

---
In reply to: [181945049](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r181945049) [](ancestors = 181945049)

test/WebJobs.Extensions.Tests/Extensions/Http/ClaimsPrincipalEndToEndTests.cs
ConnorMcMahon: I'm not positive, but after talking with Mike it sounds like the E2E tests run sequentially to avoid some bugs that come up. 

This is the method I've seen used in our "EndtoEnd" in BYOB extensions (mainly the method I used in Graph tests and seen used in the EventGrid extension). It allows you to keep more of the actual test logic in the test itself, since you have the actual bound object to test against. Assuming the tests are run in parallel, I would prefer to keep the tests this way.

---
In reply to: [181945285](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r181945285) [](ancestors = 181945285)

test/WebJobs.Extensions.Tests/Extensions/Http/ClaimsPrincipalEndToEndTests.cs
ConnorMcMahon: Looking at [https://msdn.microsoft.com/en-us/library/system.security.claims.claimsprincipal.identity(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/system.security.claims.claimsprincipal.identity(v=vs.110).aspx) under Remarks, it looks like if no identities are provided, Identity is null. I don't know what identity we would set if it fails to find both an EasyAuth and a Functions Key identity.

---
In reply to: [181945812](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r181945812) [](ancestors = 181945812)

cgillum: My recollection is that they create add a `GenericIdentity` and set IsAuthenticated to `false`.

---
In reply to: [182183202](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r182183202) [](ancestors = 182183202,181945812)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+            {
+                var request = value as ClaimsPrincipal;
+                if (request != null)
+                {
+                    var binding = new SimpleValueProvider(typeof(ClaimsPrincipal), request, "request");
+                    return Task.FromResult<IValueProvider>(binding);
+                }
+                throw new NotSupportedException($"A ClaimsPrincipal is required to bind using {nameof(ClaimsPrincipalBindingProvider)}.");
```
cgillum: >ClaimsPrincipalBindingProvider [](start = 102, length = 30)

Will the user now what a `ClaimsPrincipalBindingProvider` is? I assume not, in which case we'll want to describe the problem differently. Could we not say "A ClaimsPrincipal parameter type is required when binding to the remote user's identity" or something similar?

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            ClaimsIdentity[] identities = eligibleHeaders
+                .Where(potentialHeader => requestHeaders.ContainsKey(potentialHeader))
+                .Select(headerKey => ClaimsIdentityHelper.FromBase64EncodedJson(requestHeaders[headerKey]))
+                .ToArray();
+            return identities;
+        }
+
+        internal static bool IsEasyAuthEnabled(INameResolver nameResolver)
```
mathewc: I think this should use IEnvironment to access environment, not INameResolver

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+{
+    /// <summary>
+    /// This provider provides a binding to Type <see cref="ClaimsPrincipal"/>.
+    /// </summary>
+    internal class ClaimsPrincipalBindingProvider : IBindingProvider
+    {
+        private static readonly Task<IBinding> NullBinding = Task.FromResult<IBinding>(null);
+        private bool _isEasyAuthEnabled;
+
+        public ClaimsPrincipalBindingProvider(INameResolver nameResolver)
```
mathewc: You'd take IEnvironment here

src/WebJobs.Extensions.Http/HttpConstants.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    internal static class HttpConstants
```
mathewc: Add to the already existing constants class https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.Http/HttpExtensionConstants.cs

src/WebJobs.Extensions.Http/HttpConstants.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        public const string AntaresEasyAuthProviderHeaderName = "X-MS-CLIENT-PRINCIPAL-IDP";
```
mathewc: Nit - use "AppService" prefix rather than "Antares". We do that elsewhere.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -37,11 +38,13 @@ internal class HttpTriggerAttributeBindingProvider : ITriggerBindingProvider
     {
         internal const string HttpQueryKey = "Query";
         internal const string HttpHeadersKey = "Headers";
+        internal const string IdentitiesKey = "IdentitiesArray";
```
mathewc: Just call this this "Identities" to match other keys. E.g. for headers, we don't indicate collection type.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -54,9 +57,10 @@ internal class HttpTriggerAttributeBindingProvider : ITriggerBindingProvider
             typeof(HttpRequestMessage)
         };
 
-        public HttpTriggerAttributeBindingProvider(Action<HttpRequest, object> responseHook)
+        public HttpTriggerAttributeBindingProvider(Action<HttpRequest, object> responseHook, INameResolver nameResolver)
```
mathewc: IEnvironment

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Runtime.Serialization.Json;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// Provides methods to grab the ClaimsPrincipal from an HTTP Request trigger
+    /// </summary>
+    public static class ClaimsIdentityHelper
```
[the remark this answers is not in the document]
mathewc: Also, add unit tests for your new helper methods

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            _isEasyAuthEnabled = ClaimsIdentityHelper.IsEasyAuthEnabled(nameResolver);
+        }
+
+        public Task<IBinding> TryCreateAsync(BindingProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
```
cgillum: >"context" [](start = 48, length = 9)

nit: `nameof(context)`

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+                if (bindingData.TryGetValue(HttpTriggerAttributeBindingProvider.HttpHeadersKey, out headers))
+                {
+                    var identities = new List<ClaimsIdentity>(ClaimsIdentityHelper.GetClaimsIdentitiesFromHeaders(headers as IDictionary<string, string>, _isEasyAuthEnabled)); 
+
+                    if (identities.Count == 0)
+                    {
+                        // According to the source https://github.com/Microsoft/referencesource/blob/master/mscorlib/system/security/principal/genericidentity.cs
+                        // GenericIdentity is unauthenticated if and only if the name is set to the empty string
+                        var unauthentiatedIdentity = new GenericIdentity(name: string.Empty);
+                        identities.Add(unauthentiatedIdentity);
```
cgillum: If there are no identities, why do we create an identity at all? Why not have a ClaimsPrincipal with no identities?

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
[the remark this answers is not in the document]
ConnorMcMahon: Most of the helper methods in this class aren't a good fit for HttpRequestExtensions. However, the one public method is, so I will make this class internal after moving that into HttpRequestExtensions.

---
In reply to: [216385435](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r216385435) [](ancestors = 216385435)

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
ConnorMcMahon: It looks like IEnvironment is defined in the host repo, not in WebJobs, so there is no way for me to pull it in. Is there some other interface you were thinking of?

---
In reply to: [216387019](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r216387019) [](ancestors = 216387019)

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
ConnorMcMahon: so if we don't create an identity, then the user can't do `claimsPrincipal.Identity.IsAuthenticated`, because `claimsPrincipal.Identity` is null. You could argue whether that is required, but either way, I should probably populate it the same here as I do at line 93.

---
In reply to: [216399689](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/404#discussion_r216399689) [](ancestors = 216399689)

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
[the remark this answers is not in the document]
mathewc: Yes that was the main point - not to have the new public type

src/WebJobs.Extensions.Http/ClaimsIdentityHelper.cs
mathewc: Yeah, I guess I meant IConfiguration. Basically you're just reading an environment variable, so you don't need "name resolver" behavior. Behind the scenes, the default INameResolver is just using IConfiguration anyways.

However, I see other extensions are using INameResolver as you are, so you can leave as is if you want.

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -48,20 +49,32 @@ public static async Task<string> ReadAsStringAsync(this HttpRequest request)
             return request.Query.ToDictionary(p => p.Key, p => p.Value.Last());
         }
 
-        public static void AddIdentity(this HttpRequest request, ClaimsIdentity identity, string headerName)
+        public static ClaimsIdentity GetEasyAuthIdentity(this HttpRequest request)
```
cgillum: I would prefer we don't use "Easy Auth" in public APIs. How about using "AppService" instead?

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+            public bool FromAttribute
+            {
+                get { return false; }
+            }
+
+            public Task<IValueProvider> BindAsync(BindingContext context)
+            {
+                if (context == null)
+                {
+                    throw new ArgumentNullException("context");
```
fabiocav: nit: `nameof(context)`

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                var request = context.BindingData[HttpTriggerAttributeBindingProvider.RequestBindingName] as HttpRequest;
+                if (request == null)
```
fabiocav: nit: can we inline this declaration here?

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+                ClaimsPrincipal principal = request.HttpContext.User;
+
+                var valueProvider = new SimpleValueProvider(typeof(ClaimsPrincipal), principal, principal?.Identity?.Name);
+                return Task.FromResult<IValueProvider>(valueProvider);
+            }
+
+            public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+            {
+                var principal = value as ClaimsPrincipal;
+                if (principal != null)
```
fabiocav: nit: inline declaration here (i.e. `value is ClaimsPrincipal principal`)

src/WebJobs.Extensions.Http/Config/HttpExtensionConfigProvider.cs
```diff
@@ -14,13 +14,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
...
+        private readonly INameResolver _nameResolver;
 
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
-        public HttpExtensionConfigProvider(IOptions<HttpOptions> options)
+        public HttpExtensionConfigProvider(IOptions<HttpOptions> options, INameResolver nameResolver)
         {
             _options = options;
+            _nameResolver = nameResolver;
```
fabiocav: Revert changes to this file as it doesn't look like this field is needed anymore.

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -5,17 +5,23 @@
...
 
 namespace Microsoft.Azure.WebJobs.Extensions.Http
 {
     public static class HttpRequestExtensions
     {
         private const int DefaultBufferSize = 1024;
+        private const string EasyAuthIdentityHeader = "x-ms-client-principal";
+
+        // Shared serializer instance. This is safe for multi-threaded use.
+        private static readonly DataContractJsonSerializer Serializer = GetCustomSerializer();
```
fabiocav: Use lazy initialization here: `Lazy<DataContractJsonSerializer>`...

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -42,5 +48,33 @@ public static async Task<string> ReadAsStringAsync(this HttpRequest request)
...
+            using (var buffer = new MemoryStream(Convert.FromBase64String(payload)))
+            {
+                ClaimsIdentitySlim slim = (ClaimsIdentitySlim)Serializer.ReadObject(buffer);
+                return slim.ToClaimsIdentity();
+            }
+        }
+
+        private static DataContractJsonSerializer GetCustomSerializer()
+        {
+            // This serializer has the exact same settings as used by EasyAuth to ensure compatability
```
fabiocav: nit: comment spelling

src/WebJobs.Extensions.Http/ClaimSlim.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public Claim ToClaim()
+        {
+            return new Claim(this.Type, this.Value);
+        }
+
+        internal JObject ToJObject()
+        {
+            var jObj = new JObject();
```
fabiocav: Was this added after our chat about how this would be surfaced to end-user functions? Now we get the more descriptive property names, correct? 
:+1:

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -339,6 +348,14 @@ internal static void ApplyBindingData(object target, IDictionary<string, object>
                     bindingData[HttpHeadersKey] = request.Headers.ToDictionary(kvp => kvp.Key, kvp => kvp.Value.ToString(), StringComparer.OrdinalIgnoreCase);
                 }
 
+                if (!bindingData.ContainsKey(IdentitiesKey))
```
fabiocav: Did you intend to check `isEasyAuthEnabled` here? I don't see that being used anywhere.

src/WebJobs.Extensions.Http/ClaimsPrincipalBindingProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Parameter.ParameterType != typeof(ClaimsPrincipal))
+            {
+                return NullBinding;
+            }
+
+            return Task.FromResult<IBinding>(new ClaimsIdentityBinding(context.Parameter));
```
mathewc: Since this is named ClaimsPrincipalBindingProvider, shouldn't it be providing a ClaimsPrincipalBinding rather than a ClaimsIdentityBinding? I.e. probably need to rename your private class?

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -42,5 +48,33 @@ public static async Task<string> ReadAsStringAsync(this HttpRequest request)
...
+        private static ClaimsIdentity FromBase64EncodedJson(string payload)
+        {
+            using (var buffer = new MemoryStream(Convert.FromBase64String(payload)))
+            {
+                ClaimsIdentitySlim slim = (ClaimsIdentitySlim)Serializer.ReadObject(buffer);
+                return slim.ToClaimsIdentity();
+            }
+        }
+
+        private static DataContractJsonSerializer GetCustomSerializer()
```
mathewc: Since this serializer is specific to ClaimsIdentitySlim, recommend renaming this method GetClaimsIdentitySerializer and also renaming your static property from Serializer to ClaimsIdentitySerializer.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
ConnorMcMahon: This is now checked in the functions host.

src/WebJobs.Extensions.Http/ClaimSlim.cs
ConnorMcMahon: Yes!

PR #405, opened 2018-04-18
[SmtpMailTrigger] Add SmtpMail extension.

DarkCompiled:
Adds the capability to send e-mails using the SMTP Client provided by the framework.

### DIFF
--- a/WebJobs.Extensions.sln
+++ b/WebJobs.Extensions.sln
@@ -36,6 +36,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions.Twilio",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions.CosmosDB", "src\WebJobs.Extensions.CosmosDB\WebJobs.Extensions.CosmosDB.csproj", "{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}"
 EndProject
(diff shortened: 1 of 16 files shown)


paulbatum:
Hey there @DarkCompiled, thanks for sharing this work with us. 

We are in the gradual process of moving each WebJobs extension into its own repository so they can be versioned and released separately. Here are some examples:
- https://github.com/Azure/azure-functions-eventgrid-extension
- https://github.com/Azure/azure-functions-durable-extension

I'd like to encourage you to follow this same pattern and host this extension in its own repo (don't get too hung up on the naming pattern, functions extensions and webjobs extensions are all the same thing). 

Since you sent this as a PR, I'm guessing your intention was for this extension to officially become part of the WebJobs project (i.e. to be maintained by us going forward). Obviously that is not achieved if you host this extension in your own repo outside of the Azure github organization. If this is important to you I suggest you file a separate issue (in this repo is fine) where this can be discussed.

DarkCompiled:
Hi @paulbatum, thank you for you quick message.
I'm new to all this fork and PR topic, I would appreciate your support.

I think I don't have contributor access, that's why I forked and then made a PR.
It would be great if the extension becomes official and maintained by you, my repository is really not important.

I don't mind creating a separate repository... but how can I do that inside Azure GitHub namespace?

fabiocav:
@DarkCompiled following up on this as this PR has been open for a while. 

Looks like you'd like to have this maintained as an official extension, so could you please open an issue on the repo so we can start that discussion?

/cc @mattchenderson

PR #408, opened 2018-05-08
Fixing HTTP request content length check

fabiocav:


src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -280,7 +280,7 @@ internal static void ApplyBindingData(object target, IDictionary<string, object>
             {
                 // apply binding data from request body if present
                 var bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
-                if (request.Body.Length > 0)
```
fabiocav: This change was driven by an issue caused by a mocked request that wasn't setup as expected. Reverting.

mathewc: Thanks for fixing

Issue #415, opened 2018-05-18
CosmosDBBindings ConnectionPolicy Settings


TsuyoshiUshio:
I'd like to add the Connection Policy Setting for the CosmosDB Bindings. 

In our app, frequently configure these policies. 

```
ConnectionPolicy: PreferredLocations
ConnectionPolicy: ConnectionMode
ConnectionPolicy: ConnectionProtocol
```

These parameter is included in the best practices of CosmosDB. I guess a lot of people might want to configure that. 
 
I suggest to make it  configureble the Connection Policy via `host.json`.

If no one try to do it, I'd happy to contribute it.

k-miyake:
CosmosDB Connection Policy Setting is recommended in [Performance tips for Azure Cosmos DB and .NET](https://docs.microsoft.com/en-us/azure/cosmos-db/performance-tips#networking), so I also want it.

paulbatum:
@brettsam Did we recently add connection mode?

brettsam:
Yes, `ConnectionMode` is now added (but not yet released). https://github.com/Azure/azure-webjobs-sdk-extensions/commit/b2c30326ab6b84c005d9f18a774dce5133132ea7, but the other values are not exposed.

cc: @ealsur

ealsur:
There is an issue tracking PreferredLocations: https://github.com/Azure/azure-webjobs-sdk-extensions/issues/363 

The difference is that ConnectionMode & Protocol can be set at host.json level, but PreferredLocations should be set at the Binding / Trigger level. This is because a Trigger or Binding might be pointing to a different ConnectionString / Account and each Account could have different regions.

The problem is that if a user configures a Binding with a PreferredLocation in Function A and another Binding (same ConnectionString) on Function B but without (or different PreferredLocation) we would need to have 2 Clients. This might imply changing the cache of Clients from one per ConnectionString to one per ConnectionString/PreferredLocations.

TsuyoshiUshio:
Thank you, @ealsur 

IMO, we can compromise that. For example, we could have to have different configuration for Function A and Function B. 
However, generally speaking, people might want to configure Preferred Location for Function App, right?  The location is depend on the Region which CosmosDB and FunctionApp deployed. 
Then just add to host.json (or AppSettings) might be good option. 

What do you think @k-miyake ? 

Also, how about ConnectionProtocol? It might be good for `host.json`

ealsur:
[...]

PR #417, opened 2018-06-01
Adding CosmosDB Connection Protocol configuration option

ealsur:
As requested in https://github.com/Azure/azure-webjobs-sdk-extensions/issues/415, this PR adds the missing `Protocol` setting to `DocumentDBConfiguration` to be passed down to the `DocumentClient` instances.


ealsur:
@brettsam I thought the change for V1 in that repo already covered it for V2? Which branch should I use?

PR #431, opened 2018-07-13
Isolating extension tests

fabiocav:


mathewc:
Thanks for taking the time to split these out!

PR #434, opened 2018-07-14
Migrating HTTP extension to new DI model

fabiocav:


src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -43,6 +42,17 @@ internal class HttpTriggerAttributeBindingProvider : ITriggerBindingProvider
         internal const string RequestBindingName = "$request";
 
         private readonly Action<HttpRequest, object> _responseHook;
+        private static readonly Type[] _supportedTypes = new Type[]
```
fabiocav: This is not required, but moved this here as it will be built ahead of time. The goal was to remove the dependency on the `StreamValueBinder`, which would have made this required, but I'm not doing that in this PR.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -63,11 +73,7 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
                 return Task.FromResult<ITriggerBinding>(null);
             }
 
-            // Can bind to user types, HttpRequestMessage, object (for dynamic binding support) and all the Read
-            // Types supported by StreamValueBinder
-            IEnumerable<Type> supportedTypes = StreamValueBinder.GetSupportedTypes(FileAccess.Read)
-                .Union(new Type[] { typeof(HttpRequest), typeof(object), typeof(HttpRequestMessage) });
-            bool isSupportedTypeBinding = ValueBinder.MatchParameterType(parameter, supportedTypes);
+            bool isSupportedTypeBinding = ValueBinder.MatchParameterType(parameter, _supportedTypes);
```
fabiocav: The supported types above also help simplify this a bit.

src/WebJobs.Extensions.Http/WebJobs.Extensions.Http.csproj
```diff
@@ -7,18 +7,23 @@
     <PackageId>Microsoft.Azure.WebJobs.Extensions.Http</PackageId>
     <Description>This package adds binding extensions for Http.</Description>
   </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\WebJobs.Extensions\Framework\Bindings\StreamValueBinder.cs" Link="Binding\StreamValueBinder.cs" />
```
fabiocav: Linking to the extension project files we needed. Will create an issue to revisit this and move completely to the new binding model.

test/WebJobs.Extensions.Http.Tests/Helpers/Fakes/FakeActivator.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
fabiocav: I can see us needing a lot of these helpers in different extensions and repos. Will work on publishing the test helpers to mock/test against WebJobs and have that published as a NuGet package we can consume

fabiocav:
@mathewc have you had a chance to look at this? Any concerns with the changes?

src/WebJobs.Extensions.Http/Config/HttpExtensionOptions.cs
```diff
@@ -13,12 +13,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
...
     /// Defines the configuration options for the Http binding.
     /// </summary>
-    public class HttpExtensionConfiguration : IExtensionConfigProvider
+    public class HttpExtensionOptions
     {
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
-        public HttpExtensionConfiguration()
+        public HttpExtensionOptions()
```
mathewc: Rename to HttpOptions? The pattern for the other extensions I see is <extension name>Options. E.g. ServiceBusOptions, which is what I'd expect. However, I also see JobHostQueuesOptions, JobHostBlobsOptions which I think should be changed. We need a clear consistent pattern here.

mathewc: Kind of like you have HttpHostBuilderExtensions, StorageHostBuilderExtensions, etc.

src/WebJobs.Extensions.Http/Config/HttpHostBuilderExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds the HTTP services and extension to the provided <see cref="IHostBuilder"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHostBuilder"/> to configure.</param>
+        /// <param name="configure">An <see cref="Action{HttpExtensionOptions}"/> to configure the provided <see cref="HttpExtensionOptions"/>.</param>
+        public static IHostBuilder AddHttp(this IHostBuilder builder, Action<HttpExtensionOptions> configure)
```
mathewc: Glad to have this options overload :)

src/WebJobs.Extensions.Http/Config/HttpExtensionConfigProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// Defines the configuration options for the Http binding.
+    /// </summary>
+    public class HttpExtensionConfigProvider : IExtensionConfigProvider
```
mathewc: Does this have to be public anymore? I see other IExtensionConfigProvider implementations are non-public, e.g. TimersExtensionConfig, then others public like CosmosDBConfiguration. Hoping we can take this opportunity to get naming and visibility consistent so we have a recommended pattern.

src/WebJobs.Extensions.Http/HttpRequestManager.cs
```diff
@@ -21,24 +22,24 @@ public class HttpRequestManager
...
                 InitializeRequestQueue();
             }
         }
 
         /// <summary>
-        /// Gets the <see cref="HttpExtensionConfiguration"/>.
+        /// Gets the <see cref="HttpExtensionConfigProvider"/>.
         /// </summary>
-        protected HttpExtensionConfiguration Config { get; }
+        protected HttpExtensionOptions HttpOptions { get; }
```
mathewc: Wonder if this property shouldn't just be "Options"?

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
mathewc: I thought your plan was to cross compile these types rather than copy code? However since StreamValueBinder/ValueBinder are public, that will result in conflicts if people pull in multiple extensions?

src/WebJobs.Extensions.Http/HttpWebJobsStartup.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Extensions.Hosting;
+
+[assembly:WebJobsStartup(typeof(HttpWebJobsStartup))]
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http
+{
+    /// <summary>
+    /// Enable dynamic HTTP registration against WebJobs 
+    /// </summary>
+    public class HttpWebJobsStartup : IWebJobsStartup
```
mathewc: So this will auto-register the http extension with default options if http extension is registered? Is the storage extension and ServiceBus extension also already doing this? I looked and didn't see them doing so.

src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -17,7 +17,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Sample.Extension\Sample.Extension.csproj" />
     <ProjectReference Include="..\WebJobs.Extensions.CosmosDB\WebJobs.Extensions.CosmosDB.csproj" />
-    <ProjectReference Include="..\WebJobs.Extensions.Http\WebJobs.Extensions.Http.csproj" />
```
mathewc: The sample has some http functions - how are they working if you've removed this?

src/WebJobs.Extensions.Http/WebJobs.Extensions.Http.csproj
mathewc: Will this be an issue when other extensions also link these public types, resulting in duplicate type exports if a customer app pulls in Http and SendGrid extensions?

src/WebJobs.Extensions/Extensions/Core/ExecutionContext.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.WebJobs
     public class ExecutionContext
     {
         /// <summary>
-        /// The job function invocation ID
+        /// Gets or sets the job function invocation ID
```
mathewc: We'll probably want to move this binding into core SDK for v2 - this is functionality that should be in core.

src/WebJobs.Extensions.Http/WebJobs.Extensions.Http.csproj
fabiocav: Looking at this again, they'll become internal as we migrate the extensions. Once all the migrations are done, we can keep them in a shared project (or get rid of them)

src/WebJobs.Extensions.Http/Config/HttpExtensionConfigProvider.cs
fabiocav: No, it doesn't. Changing to internal

src/ExtensionsSample/ExtensionsSample.csproj
fabiocav: Just checked. No HTTP samples there, which does make sense as we don't really have a listener supporting HTTP.

PR #433, opened 2018-07-13
[Cosmos DB] Adding tracing and retry fix for Trigger

ealsur:
This change adds tracing inside the CosmosDBTrigger and handles the case where StopAsync might be called before Start.

It also fixes the null-ref that comes from the automatic retry when a a Function fails to load.

This PR fixes #379
This PR fixes #432

PR #437, opened 2018-07-17
[Cosmos DB] [V2] Adding tracing and retry fix for Trigger

ealsur:
This is a port of https://github.com/Azure/azure-webjobs-sdk-extensions/pull/433 to the `dev` branch.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
@@ -10,6 +10,8 @@ using Microsoft.Azure.Documents.Client;
 using Microsoft.Azure.WebJobs.Extensions.CosmosDB.Config;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Triggers;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 {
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -10,31 +10,39 @@
...
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 {
     internal class CosmosDBTriggerListener : IListener, IChangeFeedObserverFactory
     {
         private readonly ITriggeredFunctionExecutor executor;
-        private readonly ChangeFeedEventHost host;
+        private readonly ILogger logger;
```
mathewc: nit: might consider making the field naming consistent with other files (_ prefix)

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -54,21 +62,49 @@ public void Dispose()
...
+            {
+                if (this.host != null && this.listenerStarted)
+                {
+                    await this.host.UnregisterObserversAsync();
+                    this.listenerStarted = false;
+                }
+            }
+            catch (Exception ex)
+            {
+                this.logger.LogWarning($"Stopping the observer failed, potentially it was never started. Exception: {ex.Message}.");
```
mathewc: It looks the logger instance passed into this class may be null (based on `_logger = loggerFactory?.CreateLogger`). Since you require a non-null logger instance here, recommend ensuring it can't be null upstream.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
ealsur: Done

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
ealsur: Good catch!

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -20,13 +22,16 @@ internal class CosmosDBTriggerAttributeBindingProvider : ITriggerBindingProvider
...
         private CosmosDBConfiguration _config;
+        private readonly ILogger _logger;
 
-        public CosmosDBTriggerAttributeBindingProvider(INameResolver nameResolver, CosmosDBConfiguration config, ChangeFeedHostOptions leasesOptions = null)
+        public CosmosDBTriggerAttributeBindingProvider(INameResolver nameResolver, CosmosDBConfiguration config, ILoggerFactory loggerFactory, ChangeFeedHostOptions leasesOptions = null)
         {
             _nameResolver = nameResolver;
             _config = config;
             _leasesOptions = leasesOptions ?? new ChangeFeedHostOptions();
+            _logger = loggerFactory?.CreateLogger(LogCategories.CreateTriggerCategory("CosmosDB"));
```
mathewc: Why would this ever be null? I was thinking you'd just require non-null here then all the way down logger would be non-null. Seems easier than having to check for a null logger each time you use it.

PR #439, opened 2018-07-18
Removing execution context

fabiocav:
Removing the `ExecutionContext` binding from extensions. This is moving to the core SDK in https://github.com/Azure/azure-webjobs-sdk/pull/1808


src/ExtensionsSample/Samples/MiscellaneousSamples.cs
```diff
@@ -13,14 +13,15 @@ public static class MiscellaneousSamples
...
-        [NoAutomaticTrigger]
-        public static void ExecutionContext(
-            ExecutionContext context,
-            TextWriter log)
-        {
-            string msg = string.Format("Invocation ID: {0}", context.InvocationId);
-            Console.WriteLine(msg);
-            log.WriteLine(msg);
-        }
+        // Re-add this when we reference the latest SDK version.
```
fabiocav: The sample project will, unfortunately, be in a bad state until we're done migrating all of the extensions.

src/ExtensionsSample/Program.cs
```diff
@@ -30,7 +30,6 @@ public static void Main(string[] args)
             config.UseSample();
             config.UseMobileApps();
             config.UseTwilioSms();
-            config.UseCore();
```
mathewc: There are Timer/Files samples that rely on this. I think you should leave this. I want to keep the samples running.

src/ExtensionsSample/Samples/MiscellaneousSamples.cs
mathewc: Can't you wait until the core SDK change is made then pull it in? That seems like the correct staging.

fabiocav: The core SDK changes are in. If we consume them here, the other extensions (that have not been migrated yet) will need to be disabled until they're migrated. So we'd have to comment all but the migrated ones, what I have now is the opposite.

Is that what you mean?

src/ExtensionsSample/Program.cs
fabiocav: The only thing core was doing was the registration of the execution context binding provider, right? 
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/7e00bdbcefc1cfcae682a2a089af0dcef1bd992b/src/WebJobs.Extensions/Extensions/Core/Config/CoreJobHostConfigurationExtensions.cs#L36-L45

fabiocav:
@mathewc replied to your comments, not sure if that addresses your concerns.

PR #443, opened 2018-08-02
Exposing router API to retrieve function route template

fabiocav:


fabiocav:
For now, this is to expose the full route in the functions metadata response returned by the API. This should eventually be what the CLI uses as well.

PR #447, opened 2018-08-03
[CosmosDB] Moving to new DI patterns

brettsam:
Moving CosmosDB up to latest DI changes.

Note that in order for this to build, I had to move the CosmosDB sample into its own project (due to the original sample referencing a pre-DI version of WebJobs).


samples/CosmosDBSample/CosmosDBSample.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
mathewc: In https://github.com/Azure/azure-webjobs-sdk-extensions/pull/448 I just migrated everything, so I didn't have to split any samples out.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBHostBuilderExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            return hostBuilder
+                .AddExtension<CosmosDBExtensionConfigProvider>()
+                .ConfigureServices(services =>
+                {
+                    services.AddSingleton<ICosmosDBServiceFactory, DefaultCosmosDBServiceFactory>();
+                    services.AddOptions<CosmosDBOptions>()
+                        .Configure<IConfiguration>((options, config) =>
+                        {
+                            config.GetSection(ConfigSectionName)?.Bind(options);
```
ealsur: If this is using `?.Bind` wouldn't it mean that in the other places where `_options` is used, it should be `_options?` since it might be null?

samples/CosmosDBSample/CosmosDBSample.csproj
brettsam: I actually liked the split-out samples -- it meant no more commenting out stuff before testing my extension. I was hoping to do it with all of them. But I don't care enough to fight for it -- so I'll get these non-sample changes merged with yours (so you get all the tests completed) and add it to your branch.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBHostBuilderExtensions.cs
brettsam: Options won't be null as long as you ask for an `IOptions<CosmosDBOptions>`. By default in both the `HostBuilder` and `WebHostBuilder`, there's a call to `AddOptions()`, which registers the `OptionsManager<>` for any type of `IOptions<>`. This means anytime you request anything via `IOptions<>`, it will be instantiated (or pulled from the cache) and given to you. The call to `Bind()` will set matching properties on that object with values from `IConfiguration`, so if there's no config section found, you'll just end up with the default values.

Good question -- I had to look this up so I made sure I understood why it worked even if there was no config section :-)
- Here's where `HostBuilder` calls `AddOptions()` as part of it's call to `.Build()`: https://github.com/aspnet/Hosting/blob/master/src/Microsoft.Extensions.Hosting/HostBuilder.cs#L183
- Here's where `AddOptions()` registers the `OptionsManager<>`: https://github.com/aspnet/Options/blob/master/src/Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs#L30
- Here's the manager: https://github.com/aspnet/Options/blob/master/src/Microsoft.Extensions.Options/OptionsManager.cs
- And the manager uses this factory for instantiating types:  https://github.com/aspnet/Options/blob/master/src/Microsoft.Extensions.Options/OptionsFactory.cs#L41

brettsam:
I've merged these changes with #448. Closing in favor of that one.

PR #448, opened 2018-08-03
Migrating Extensions to new DI model

mathewc:
All extensions updated and all tests passing


WebJobs.Extensions.sln
```diff
@@ -21,8 +21,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ExtensionsSample", "src\Ext
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebJobs.Extensions.Tests", "test\WebJobs.Extensions.Tests\WebJobs.Extensions.Tests.csproj", "{1729561F-D598-4208-9154-3A5497523D54}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sample.Extension", "src\Sample.Extension\Sample.Extension.csproj", "{55EDABA0-8D31-43C9-AB3C-2A40D0EFF28A}"
```
mathewc: Just decided to remove this sample. We have enough extensions for people to look at as samples.

src/ExtensionsSample/Samples/MiscellaneousSamples.cs
```diff
@@ -1,27 +0,0 @@
...
-namespace ExtensionsSample
-{
-    public static class MiscellaneousSamples
-    {
-        /// <summary>
-        /// Demonstrates use of the ExecutionContext binding
-        /// </summary>
-        // Re-add this when we reference the latest SDK version.
-        //[NoAutomaticTrigger]
-        //public static void ExecutionContext(
```
mathewc: ExecutionContext was moved to core, so we don't need to test here.

src/ExtensionsSample/Samples/TableSamples.cs
```diff
@@ -1,23 +0,0 @@
...
-using Microsoft.Azure.WebJobs;
-using Sample.Extension;
-
-namespace ExtensionsSample
-{
-    public static class TableSamples
-    {
-        // Demonstrates use of a custom table binding extension to bind Table
-        // to a custom type (Table<T>).
-        public static void CustomBinding([Table("sampletable")] Table<Person> table)
```
mathewc: Old outdated sample.

src/WebJobs.Extensions.SendGrid/Config/SendGridOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+            private readonly IConfiguration _config;
+
+            public Setup(IConfiguration config)
+            {
+                _config = config;
+            }
+
+            public void Configure(SendGridOptions options)
+            {
+                // TODO: Need to access raw to/from strings from config
```
mathewc: We need to handle the mapping of host.json to/from raw string values to EmailAddress instances here. Need to determine the right way to do this.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitor.cs
```diff
@@ -17,19 +18,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
     public class StorageScheduleMonitor : ScheduleMonitor
     {
         private const string HostContainerName = "azure-webjobs-hosts";
-        private readonly JobHostConfiguration _hostConfig;
+        private readonly JobHostOptions _hostOptions;
+        private readonly DistributedLockManagerContainerProvider _lockContainerProvider;
```
mathewc: We need to review all the changes made during the storage binding split. There are a lot of new public types and surface area that looks questionable to me.

brettsam:
I've got Cosmos just about ready here: #447. I wasn't aware you were trying to handle all of these at once.

mathewc:
@brettsam Sorry I was unaware you were working on any of these. My problem was that in trying to migrate Timers/Files (core extensions) I had to bump the SDK version. In doing so, it caused all the other extensions to no longer compile due to their dependency on the core Extensions project. So I just kept pulling the thread ...

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -10,11 +10,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.DocumentDB.ChangeFeedProcessor" Version="1.3.3" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.0-beta7-11385" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
   </ItemGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\WebJobs.Extensions\WebJobs.Extensions.csproj" />
```
mathewc: I removed the dependency on the Extensions project from all of these. It used to exist because many of the extensions were using shared binding provider base classes, etc. but that is no longer the case.

brettsam:
Let's figure out how to combine these. Cosmos is effectively done (tests and all) so ideally we could merge these two together without too much effort.

src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -5,17 +5,21 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <Version>2.0.0-beta1$(VersionSuffix)</Version>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
fabiocav: I realize this was added by the tool, but we should move the LangVersion up into the previous group and out of the conditional

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -20,42 +21,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
     /// <summary>
     /// Defines the configuration options for the CosmosDB binding.
     /// </summary>
-    public class CosmosDBConfiguration : IExtensionConfigProvider
```
fabiocav: We still need these types to be public, at least for now, as the build tools still probe for them. I know we talked about making all of these consistent and internal, but unfortunately, we can’t do that yet

src/WebJobs.Extensions.SendGrid/Config/SendGridOptions.cs
fabiocav: The goal is to do this in a generic way for extensions so they don’t all need to have setup classes. I’ll have something in place tomorrow.

src/WebJobs.Extensions/Extensions/ExtensionsWebJobsStartup.cs
```diff
@@ -0,0 +1,20 @@
...
+[assembly: WebJobsStartup(typeof(ExtensionsWebJobsStartup))]
+
+namespace Microsoft.Azure.WebJobs.Extensions
+{
+    public class ExtensionsWebJobsStartup : IWebJobsStartup
+    {
+        public void Configure(IHostBuilder builder)
+        {
+            builder.AddTimers();
+            builder.AddFiles();
```
fabiocav: Do we always want to have those bundled? In the case of script, do we want files?

src/WebJobs.Extensions/Extensions/Files/Bindings/FileTriggerAttributeBindingProvider.cs
```diff
@@ -17,12 +17,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Files.Bindings
...
     internal class FileTriggerAttributeBindingProvider : ITriggerBindingProvider
     {
-        private readonly FilesConfiguration _config;
+        private readonly FilesOptions _options;
         private readonly ILogger _logger;
 
-        public FileTriggerAttributeBindingProvider(FilesConfiguration config, ILoggerFactory loggerFactory)
+        public FileTriggerAttributeBindingProvider(FilesOptions options, ILoggerFactory loggerFactory)
         {
-            _config = config ?? throw new ArgumentNullException("config");
```
fabiocav: nit: mismatched parameter name. Update to use nameof

src/WebJobs.Extensions/Extensions/Files/Bindings/FileTriggerAttributeBindingProvider.cs
```diff
@@ -17,12 +17,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Files.Bindings
 {
     internal class FileTriggerAttributeBindingProvider : ITriggerBindingProvider
     {
-        private readonly FilesConfiguration _config;
+        private readonly FilesOptions _options;
         private readonly ILogger _logger;
 
-        public FileTriggerAttributeBindingProvider(FilesConfiguration config, ILoggerFactory loggerFactory)
```
fabiocav: This should be using `IOptions<T>` here

fabiocav (commented):
Took a quick look at some of the changes and left a couple of comments/questions.

So nice to see of of those being moved over!

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
brettsam: What probes for it? I made it internal in my changes as well. Otherwise we need to make several other internal service interfaces public.

src/WebJobs.Extensions.SendGrid/Config/SendGridOptions.cs
brettsam: I did something in the Cosmos changes that I wanted to get comments on today -- binding directly to a section of `IConfiguration` (and note that you don't need separate setup classes anymore):

    services.AddOptions<CosmosDBOptions>()
        .Configure<IConfiguration>((options, config) =>
        {
            config.GetSection("CosmosDB")?.Bind(options);
        });

mathewc: Updated to follow this pattern. After this PR gets in, if Fabio has a new model we can move to that. These changes work.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
mathewc: I think we should make them internal like they should be and update the tooling to work properly.

src/ExtensionsSample/ExtensionsSample.csproj
mathewc: Fixed

src/WebJobs.Extensions/Extensions/ExtensionsWebJobsStartup.cs
mathewc: These setup classes are designed to auto register all extensions for the extension assembly. While we're not using Files currently in Functions, it won't hurt anything.

src/WebJobs.Extensions/Extensions/Files/Bindings/FileTriggerAttributeBindingProvider.cs
mathewc: Fixed

src/WebJobs.Extensions/Extensions/Files/Bindings/FileTriggerAttributeBindingProvider.cs
mathewc: Fixed

src/ExtensionsSample/Program.cs
```diff
@@ -2,65 +2,74 @@
...
+               .UseConsoleLifetime();
+
+            EnsureSampleDirectoriesExist(filesTestPath);
 
-            host.RunAndBlock();
+            var host = builder.Build();
+            using (host)
+            {
+                // Some direct invocations to demonstrate various binding scenarios
+                var jobHost = (JobHost)host.Services.GetService<IJobHost>();
```
fabiocav: Do we want this in the sample?

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
fabiocav: I agree that it would be nice to change this, but if we do that now, we won't be able to use/test any of the extensions using the existing tooling until that is resolved.

src/WebJobs.Extensions.MobileApps/Config/MobileAppsExtensionConfigProvider.cs
```diff
@@ -4,51 +4,40 @@
...
-using System.Collections.Generic;
-using System.Reflection;
 
 namespace Microsoft.Azure.WebJobs.Extensions.MobileApps
 {
     /// <summary>
     /// Defines the configuration options for the motile table binding.
     /// </summary>
-    public class MobileAppsConfiguration : IExtensionConfigProvider
+    internal class MobileAppsExtensionConfigProvider : IExtensionConfigProvider
```
fabiocav: Same feedback as with the other internal providers.

src/WebJobs.Extensions/Extensions/ExtensionsWebJobsStartup.cs
fabiocav: Hence my question. We'll have the Files extension registered in Functions, which will also make it available (and supported) for use there. Do we want that?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitor.cs
fabiocav: Agreed, that is definitely one of them.

PR #451, opened 2018-08-08
ExtensionAttribute pattern refinements (#3186)

mathewc:
We didn't really want IExtensionConfigProviders to have to be public just for tooling discovery. This actually was problematic in a couple cases (Blobs/Queues) where these couldn't be made public due to the fact that the ctors for the config providers referenced non-public types injected via DI.

So instead I'd like to move to the pattern where it's the WebJobsStartup class for an extension that is annotated. The runtime discovers these and runs them to register extensions, so these really do define the extensions that are installed at runtime. So for the tooling issue https://github.com/Azure/azure-functions-host/issues/3186 I'll update the tooling to instead search for WebJobsStartup types annotated with Extension and those will be added to extensions.json. The runtime already treats these as startup types.


src/Sample.Extension/Sample.Extension.csproj
```diff
@@ -1,12 +0,0 @@
-﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
-  <Import Project="..\..\build\common.props" />
-  <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <AssemblyName>Sample.Extension</AssemblyName>
```
mathewc: Removed this from the solution in the last PR but didn't delete the files :)

PR #456, opened 2018-08-11
Move Options services to DI (#1833)

mathewc:
Applying similar changes to those in core SDK for https://github.com/Azure/azure-webjobs-sdk/issues/1833.


mathewc:
@fabiocav FYI - I made the same DI changes for the Files/Timers extensions as well :)

PR #458, opened 2018-08-13
Pulling in latest SDK

mathewc:
To take connection string changes


fabiocav:
Merging this in as we need to pull these through into the host.

Issue #455, opened 2018-08-10
[CosmosDBTrigger] Race in start/stop


brettsam:
PR #433 introduced some graceful start/stop functionality. But running the tests in the functions-host branch show that we've got a race now between this start/stop where it's possible to leave a ChangeFeedProcessor running after the host has stopped.

It looks like `RegisterObserverFactoryAsync()` can take a little while to run. If we call `Stop()` while it's still running, we'll never call `UnregisterObserversAsync()`, and that processor will steal away change notifications from tests later in the run.

Is calling `Unregister()` before `Register()` is complete okay? It looks like that may be what we used to do. This may just need another check for "Registering" so it knows if its safe to call `Unregister()`. Either that or we need a lock/semaphoreslim that waits for any outstanding register calls to complete before exiting the Stop (which may take a little while).

ealsur:
We could add a lock mechanism, I wonder though, what should be the Stop behavior if the lock is still taken (ie it's Registering)? Should it wait?

brettsam:
It seems like it could be a potentially long lock period, which isn't great. But looking at the previous code it looks like we were okay to call StartAsync() and then StopAsync() pretty quickly and it didn't throw an exception. 

Do you know the behavior of Unregister() being called while there's a currently-running Register()? We weren't having any problems with that pattern before. Maybe we don't need to check for `listenerStarted`?

https://github.com/Azure/azure-webjobs-sdk-extensions/blob/4e792e72faa9c94783cd828c4a1e8fe0d9e3b193/src/WebJobs.Extensions.DocumentDB/Trigger/CosmosDBTriggerListener.cs#L55-L72

ealsur:
Maybe we can change it to:

    try
    {
        if (this.host != null)
        {
            await this.host.UnregisterObserversAsync();
            this.listenerStarted = false;
        }
    }
    catch (Exception ex)
    {
        this.trace.Warning($"Stopping the observer failed, potentially it was never started. Exception: {ex.Message}.");
    }

brettsam:
That call to set `listenerStarted` worries me. What if you call 

`StartAsync()` -> takes 5 seconds
`StopAsync()` -> takes 1 second

You'd end up with `listenerStarted = true`, I think. Yet I believe it would be stopped. Is there any way to get the true state of the host directly without tracking it ourselves?

ealsur:
Does this mean that:
a. The listener can be accessed from multiple threads
[...]

Issue #453, opened 2018-08-09
Cosmos DB Trigger - Add StartFromBeginning setting


ealsur:
Adding this issue to track the need to expose and map the `StartFromBeginning` setting for the Cosmos DB Trigger, which enables a Trigger to start sending all the changes that ever happened in a collection instead of the default behavior of sending changes since the first time the Function runs.

PR #467, opened 2018-08-27
[CosmosDB] Start/Stop and StartFromBeginning

ealsur:
This PR addresses #455 and #453 by adding the listener state as an enum supporting multiple states and adding a new Attribute that maps to the internal `StartFromBeginning`.


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -62,17 +62,26 @@ public void Dispose()
...
             this.InitializeHost();
 
             try
             {
                 await this._host.RegisterObserverFactoryAsync(this);
-                this._listenerStarted = true;
+                if (this._listenerStatus == RegistrationStatus.Registering)
+                {
+                    this._listenerStatus = RegistrationStatus.Registered;
+                }
```
brettsam: There's a potential race condition here. Can you use `Interlocked.CompareExchange()` instead? Note that you'll have to get rid of the enum, I believe, and do something like we've done elsewhere with `int`s. Example here: https://github.com/Azure/azure-webjobs-sdk/blob/bb0b0e305ebbcb192cd57a3e6154c321a682e06a/src/Microsoft.Azure.WebJobs.Host/JobHost.cs#L94-L97.

brettsam: Note that you may want to do something similar above where you set the state to Registering (which would ensure that the state is `NotRegistered` before you begin).

ealsur: Done

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -62,17 +59,20 @@ public void Dispose()
...
 
             this.InitializeHost();
 
             try
             {
                 await this._host.RegisterObserverFactoryAsync(this);
-                this._listenerStarted = true;
+                if (Interlocked.CompareExchange(ref this._listenerStatus, ListenerRegistered, ListenerRegistering) != ListenerRegistering)
+                {
+                    throw new InvalidOperationException("The listener has already been started or was stopped during initialization.");
```
brettsam: You may not want to throw here -- just let it silently stay at "NotRegistered"? That would typically mean that Cancel() was called during StartAsync, in which case we probably just want to exit silently without touching the status. I think removing the throw would accomplish that. Does that seem right?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -87,10 +87,10 @@ public async Task StopAsync(CancellationToken cancellationToken)
         {
             try
             {
-                if (this._host != null && this._listenerStarted)
+                if (this._host != null)
                 {
                     await this._host.UnregisterObserversAsync();
-                    this._listenerStarted = false;
+                    Interlocked.CompareExchange(ref this._listenerStatus, ListenerNotRegistered, ListenerRegistered);
```
brettsam: There's a time when you actually will set it to NotRegistered while it is Registering (during the call to Cancel). But b/c this is a single operation, you should remove the Interlocked.CompareExchange here. Just set it straight to NotRegistered, no matter what.

brettsam:
I can't remember if you still have permissions to merge (we adjusted those recently). If not, can you squash this down to a single commit and I'll merge it for you?

PR #473, opened 2018-08-29
[CosmosDb] fixing trigger retry behavior

brettsam:
Doing some unrelated testing and realized that we weren't re-setting the value here to `NotRegistered` if there was an exception. That meant the retry never succeeded. Also added a test and improved the status error to be more precise.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
@@ -59,25 +59,46 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 
         public async Task StartAsync(CancellationToken cancellationToken)
         {
-            if (Interlocked.CompareExchange(ref this._listenerStatus, ListenerRegistering, ListenerNotRegistered) != ListenerNotRegistered)
+            int previousStatus = Interlocked.CompareExchange(ref this._listenerStatus, ListenerRegistering, ListenerNotRegistered);
+
+            if (previousStatus == ListenerRegistering)
             {
-                throw new InvalidOperationException("The listener has already been started or is starting.");
+                throw new InvalidOperationException("The listener is already starting.");
+            }
+            else if (previousStatus == ListenerRegistered)
+            {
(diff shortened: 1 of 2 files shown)


test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBListenerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB.Tests.Trigger
+{
+    public class CosmosDBListenerTests
+    {
+        [Fact]
+        public async Task StartAsync_Retries()
```
brettsam: I made sure this test failed before the change.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -59,27 +59,48 @@ public void Dispose()
...
-                throw new InvalidOperationException(message, ex);
+                // Reset to NotRegistered
+                this._listenerStatus = ListenerNotRegistered;
+
+                // Throw a custom error if NotFound.
+                if (ex is DocumentClientException docEx && docEx.StatusCode == HttpStatusCode.NotFound)
+                {
+                    // Throw a custom error so that it's easier to decipher.
+                    string message = $"Either the source collection '{_monitorCollection.CollectionName}' (in database '{_monitorCollection.DatabaseName}')  or the lease collection '{_leaseCollection.CollectionName}' (in database '{_leaseCollection.DatabaseName}') does not exist. Both collections must exist before the listener starts. To automatically create the lease collection, set '{nameof(CosmosDBTriggerAttribute.CreateLeaseCollectionIfNotExists)}' to 'true'.";
+                    this._host = null;
```
ealsur: Add `this._listenerStatus = ListenerNotRegistered;` here too?

brettsam: This is nested now (rather than being an exception filter) -- so the call above will execute for this as well.

PR #479, opened 2018-08-30
incrementing cosmos version

brettsam:


brettsam:
@fabiocav -- as soon as this passes appveryor and looks right, i'll merge and then tag for an official build.

PR #488, opened 2018-09-13
[CosmosDB] updating connection string behavior

brettsam:


brettsam:
@ealsur FYI -- we're changing how we pull the default values for all our extensions. See https://github.com/Azure/azure-webjobs-sdk/issues/1901 for context.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -169,52 +164,67 @@ internal static TimeSpan ResolveTimeSpanFromMilliseconds(string nameOfProperty,
 
         private string ResolveAttributeConnectionString(CosmosDBTriggerAttribute attribute)
         {
-            string connectionString = this._monitorConnectionString;
-            if (!TryReadFromSettings(attribute.ConnectionStringSetting, connectionString, out connectionString))
```
ealsur: I think `TryReadFromSettings` is no longer used but still in the code

brettsam: Oops yep you're right. I'll pull it out.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -169,52 +164,67 @@ internal static TimeSpan ResolveTimeSpanFromMilliseconds(string nameOfProperty,
...
+
+            if (string.IsNullOrEmpty(connectionString))
             {
+                string attributeProperty = $"{nameof(CosmosDBTriggerAttribute)}.{nameof(CosmosDBTriggerAttribute.ConnectionStringSetting)}";
+                string optionsProperty = $"{nameof(CosmosDBOptions)}.{nameof(CosmosDBOptions.ConnectionString)}";
                 throw new InvalidOperationException(
-                    string.Format(CultureInfo.CurrentCulture,
-                    "The CosmosDBTriggerAttribute.ConnectionStringSetting '{0}' property specified does not exist in the Application Settings.",
-                    attribute.ConnectionStringSetting));
+                    $"The CosmosDBTrigger connection string must be set either via a '{Constants.DefaultConnectionStringName}' IConfiguration connection string, via the {attributeProperty} property or via {optionsProperty}.");
```
mathewc: It looks like this complicated error message formatting is duplicated? Would be good to centralize it on one place.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -169,52 +164,67 @@ internal static TimeSpan ResolveTimeSpanFromMilliseconds(string nameOfProperty,
...
-            string connectionString = string.IsNullOrEmpty(this._leasesConnectionString) ? triggerConnectionString : this._leasesConnectionString;
-            if (!TryReadFromSettings(attribute.LeaseConnectionStringSetting, connectionString, out connectionString))
+            // First, resolve the string.
+            if (!string.IsNullOrEmpty(unresolvedConnectionString))
             {
-                throw new InvalidOperationException(
-                    string.Format(CultureInfo.CurrentCulture,
-                    "The CosmosDBTriggerAttribute.LeaseConnectionStringSetting '{0}' property specified does not exist in the Application Settings.",
-                    attribute.ConnectionStringSetting));
+                string resolvedString = _nameResolver.Resolve(unresolvedConnectionString);
```
fabiocav: Need to discuss this. Not sure the resolution should go through the name resolver, but at the very least, it should still fallback to a connection string lookup. Ideally, this would be common logic reused across all extensions.

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -10,7 +10,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.DocumentDB.ChangeFeedProcessor" Version="2.0.6" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.0-rc1" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.0-rc1-11444" />
```
fabiocav: Same comment on the suffixed version.

test/WebJobs.Extensions.CosmosDB.Tests/CosmosDBMockEndToEndTests.cs
```diff
@@ -302,24 +303,35 @@ private async Task RunTestAsync(Type testType, string testName, ICosmosDBService
             resolver.Values.Add("Collection", "ResolvedCollection");
             resolver.Values.Add("MyConnectionString", AttributeConnStr);
             resolver.Values.Add("Query", "ResolvedQuery");
-            if (includeDefaultConnectionString)
-            {
-                resolver.Values.Add(CosmosDBExtensionConfigProvider.AzureWebJobsCosmosDBConnectionStringName, DefaultConnStr);
-            }
 
-            IHost host = new HostBuilder()
+            IHost host = new HostBuilder()
```
fabiocav: nit: space

src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="3.0.0-rc1" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="3.0.0-rc1-11444" />
```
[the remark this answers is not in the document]
brettsam: No, the callback that passes in the root config with the extensions path is new.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
brettsam: Yup, will do.

src/ExtensionsSample/Program.cs
```diff
@@ -27,11 +28,10 @@ public static async Task Main(string[] args)
                 typeof(TimerSamples));
 
             var builder = new HostBuilder()
-               .UseEnvironment("Development")
+               .UseEnvironment("Development")
```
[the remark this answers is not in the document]
brettsam: This whitespace should be gone. But the stuff below makes the sample work. It didn't before (type locator didn't get overridden).

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -83,13 +78,12 @@ public void Initialize(ExtensionConfigContext context)
...
-                string.IsNullOrEmpty(_defaultConnectionString))
+                string.IsNullOrEmpty(attribute.ConnectionStringSetting))
             {
+                string attributeProperty = $"{nameof(CosmosDBAttribute)}.{nameof(CosmosDBAttribute.ConnectionStringSetting)}";
+                string optionsProperty = $"{nameof(CosmosDBOptions)}.{nameof(CosmosDBOptions.ConnectionString)}";
                 throw new InvalidOperationException(
-                    string.Format(CultureInfo.CurrentCulture,
-                    "The CosmosDB connection string must be set either via a '{0}' app setting, via the CosmosDBAttribute.ConnectionStringSetting property or via CosmosDBConfiguration.ConnectionString.",
-                    AzureWebJobsCosmosDBConnectionStringName));
+                    $"The CosmosDB connection string must be set either via the '{Constants.DefaultConnectionStringName}' IConfiguration connection string, via the {attributeProperty} property or via {optionsProperty}.");
```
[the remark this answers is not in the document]
brettsam: Yeah, it's confusing b/c we can't guarantee where it's even coming from (b/c config can have different sources). I guess I'll just make it generic and refer to 'configuration'.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
brettsam: We could go directly against `IConfiguration` -- but all other bindings use `INameResolver` and this isn't a change here, it's just been relocated. It *does* fallback to the connection string. That's because it'll return the _options.ConnectionString if this is missing. And that is set up with the ConnectionString.

brettsam: Actually, now I get what you're saying. I'm going to swap this to go directly against IConfiguration and check ConnectionStrings first. I really want to [obsolete] INameResolver so we can eventually remove it. We'll need to port the ResolveWholeString() behavior for resolving %% in strings, but that shouldn't be tough.

PR #492, opened 2018-09-14
[Timer] conditionally adjust for DST

### LINKED ISSUE #489: TimerTrigger DST issues
It appears that we have more issues with the DST fix here: https://github.com/Azure/azure-functions-host/issues/2546. The `StartAsync_ExtendedScheduleInterval_TimerContinuesUntilTotalIntervalComplete` now fails when the end-time falls over the DST line. It's asking explicitly for a 52-hour timer, yet it gets a 53-hour timer when we do our calculation.

brettsam:
Fixes #489

### DIFF
--- a/test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
+++ b/test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
@@ -2,13 +2,13 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Extensions.Tests.Common;
 using Microsoft.Azure.WebJobs.Extensions.Timers;
 using Microsoft.Azure.WebJobs.Extensions.Timers.Listeners;
-using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Extensions.Logging;
 using Moq;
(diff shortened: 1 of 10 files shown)


test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
```diff
@@ -358,49 +358,60 @@ public async Task Listener_LogsInitialNullStatus_WhenUsingMonitor()
...
         /// next occurrence AFTER the DST transition.
         /// </summary>
-        [Fact]
-        public void GetNextInterval_NextAfterDST_ReturnsExpectedValue()
+        [Theory]
+        [MemberData(nameof(TimerSchedulesAfterDST))]
+        public void GetNextInterval_NextAfterDST_ReturnsExpectedValue(TimerSchedule schedule, TimeSpan expectedInterval)
         {
+            Console.WriteLine("Current TimeZone is: " + TimeZoneInfo.Local.ToString());
+            Console.WriteLine("Supports DST: " + TimeZoneInfo.Local.SupportsDaylightSavingTime);
```
brettsam: I added this to see what Appveyor uses -- it uses Pacific time, surprisingly. But that also explains why these tests pass. I'm going to make them throw an exception if the local timezone does not support DST. It would fail anyway, but at least we can give a good explanation then.

fabiocav: This fixes the test, but I think we still have a product issue that needs to be addressed.

brettsam:
Well, it looks like AppVeyor is now setting itself to UTC. I'm not sure how it determines the time zone. So I've forced it to Pacific in the `run-tests.ps1` file (and reset it back to the original). That should guarantee our timezone tests pass reliably.

pjt33:
I'm not convinced that this is a complete fix. In particular, I note that the tests relating to DST transition only test the transition *into* DST and not *out of* it. Bearing that in mind, it's easy to demonstrate that a schedule of `"0 30 * * * *"` skips an hour when DST ends because there will be a time (e.g. 2018-11-04T01:30:00 in PST) where the next execution should be at the *same* local time, which is an hour later.

PR #499, opened 2018-09-19
[CosmosDB] fixing connection setting fallback for leases

brettsam:
Fixes https://github.com/Azure/azure-functions-host/issues/3484

We had some test gaps where we didn't validate the correct `LeaseConnectionStringSetting` -> `ConnectionStringSetting` -> `ConnectionStrings:CosmosDB` fallback. 

I've beefed up those tests, which failed before the fix.

### DIFF
--- a/test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
+++ b/test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDBTrigger.Tests
 
         public static IEnumerable<object[]> ValidCosmosDBTriggerBindingsWithAppSettingsParameters
         {
-            get { return ValidCosmosDBTriggerBindigsWithAppSettings.GetParameters(); }
+            get { return ValidCosmosDBTriggerBindingsWithAppSettings.GetParameters(); }
         }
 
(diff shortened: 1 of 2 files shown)


test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -105,7 +111,8 @@ public async Task ValidParametersWithAppSettings_Succeed(ParameterInfo parameter
             CosmosDBTriggerBinding binding = (CosmosDBTriggerBinding)await provider.TryCreateAsync(new TriggerBindingProviderContext(parameter, CancellationToken.None));
 
             Assert.Equal(typeof(IReadOnlyList<Document>), binding.TriggerValueType);
-            Assert.Equal(binding.DocumentCollectionLocation.Uri, new Uri("https://fromSettings"));
+            Assert.Equal(new Uri("https://fromSettings"), binding.DocumentCollectionLocation.Uri);
+            Assert.Equal(new Uri("https://fromSettings"), binding.LeaseCollectionLocation.Uri);
```
brettsam: I made sure that we validate the lease URI in each of these tests no. Some of these were falling back to use "https://fromEnvironment" when they shouldn't be.

ealsur: Good catch, I didn't notice it in the refactor in the previous PR

PR #508, opened 2018-10-02
[Cosmos DB] Implementing Multi Master and PreferredLocations

### LINKED ISSUE #363: Allow for CosmosDB to set preferred read location for better perf
Per [this uservoice item](https://feedback.azure.com/forums/355860-azure-functions/suggestions/32759485-support-setting-preferred-location-preferences-on) which references [this document](https://docs.microsoft.com/en-us/azure/cosmos-db/tutorial-global-distribution-sql-api#net-sdk) sounds like we may get better read perf if allowed configuration of the location to read from in CosmosDB.  Not sure what impact is or if feasible but wanted to bubble up for tracking and triage.

/cc @ealsur

### LINKED ISSUE #415: CosmosDBBindings ConnectionPolicy Settings

ealsur:
This PR aims at allowing users to interact with [Multi Master accounts](https://aka.ms/multimasterdocs) by exposing the `UseMultipleWriteLocations` flag.

To be able to use Multi Master effectively, users should be able to set the `PreferredLocations` list and define region priorities, so we expose that setting too.

`PreferredLocations` is not only beneficial for Multi Master accounts, but Non-Multi-Master accounts can benefit from it as explained in #363 

For this PR we are enabling  `UseMultipleWriteLocations` and  `PreferredLocations` for the Input/Output bindings and  `PreferredLocations` for the Trigger (the trigger reads the Change Feed, enabling Multi Master there has no real effect).

New tests were added to cover the new attributes.

The `Microsoft.Azure.DocumentDB.Core` package is bumped to version `2.1.1` to support Multi Master.

This PR fixes #363, and fixes #415

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
         internal DocumentClient BindForClient(CosmosDBAttribute attribute)
         {
             string resolvedConnectionString = ResolveConnectionString(attribute.ConnectionStringSetting);
-            ICosmosDBService service = GetService(resolvedConnectionString);
+            ICosmosDBService service = GetService(resolvedConnectionString, attribute.PreferredLocations, attribute.UseMultipleWriteLocations);
 
             return service.GetClient();
         }
@@ -124,16 +124,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             return _options.ConnectionString;
         }
 
-        internal ICosmosDBService GetService(string connectionString)
+        internal ICosmosDBService GetService(string connectionString, string preferredLocations = "", bool useMultipleWriteLocations = false)
         {
-            return ClientCache.GetOrAdd(connectionString, (c) => _cosmosDBServiceFactory.CreateService(c, _options.ConnectionMode, _options.Protocol));
+            string cacheKey = BuildCacheKey(connectionString, preferredLocations, useMultipleWriteLocations);
+            ConnectionPolicy connectionPolicy = CosmosDBUtility.BuildConnectionPolicy(_options.ConnectionMode, _options.Protocol, preferredLocations, useMultipleWriteLocations);
+            return ClientCache.GetOrAdd(cacheKey, (c) => _cosmosDBServiceFactory.CreateService(connectionString, connectionPolicy));
         }
 
         internal CosmosDBContext CreateContext(CosmosDBAttribute attribute)
         {
             string resolvedConnectionString = ResolveConnectionString(attribute.ConnectionStringSetting);
 
-            ICosmosDBService service = GetService(resolvedConnectionString);
+            ICosmosDBService service = GetService(resolvedConnectionString, attribute.PreferredLocations, attribute.UseMultipleWriteLocations);
 
             return new CosmosDBContext
             {
@@ -152,6 +154,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 
             return false;
         }
+
+        internal static string BuildCacheKey(string connectionString, string preferredLocations, bool useMultipleWriteLocations) => $"{connectionString}|{preferredLocations}|{useMultipleWriteLocations}";
 
         private class DocumentOpenType : OpenType.Poco
         {
--- a/src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs
+++ b/src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs
@@ -102,6 +102,23 @@ namespace Microsoft.Azure.WebJobs
         [AutoResolve(ResolutionPolicyType = typeof(CosmosDBSqlResolutionPolicy))]
         public string SqlQuery { get; set; }
 
+        /// <summary>
+        /// Optional.
+        /// Enable to use with Multi Master accounts.
+        /// </summary>
+        public bool UseMultipleWriteLocations { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Defines preferred locations (regions) for geo-replicated database accounts in the Azure Cosmos DB service.
+        /// Values should be comma-separated.
+        /// </summary>
+        /// <example>
+        /// PreferredLocations = "East US,South Central US,North Europe"
+        /// </example>
+        [AutoResolve]
+        public string PreferredLocations { get; set; }
+
         internal SqlParameterCollection SqlQueryParameters { get; set; }
     }
 }
--- a/src/WebJobs.Extensions.CosmosDB/CosmosDBUtility.cs
+++ b/src/WebJobs.Extensions.CosmosDB/CosmosDBUtility.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Azure.Documents;
 using Microsoft.Azure.Documents.Client;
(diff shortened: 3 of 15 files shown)


src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -124,16 +124,17 @@ internal string ResolveConnectionString(string attributeConnectionString)
             return _options.ConnectionString;
         }
 
-        internal ICosmosDBService GetService(string connectionString)
+        internal ICosmosDBService GetService(string connectionString, string preferredLocations = "", bool useMultipleWriteLocations = false)
```
brettsam: nit: use `string.Empty` here.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -153,6 +154,8 @@ internal static bool IsSupportedEnumerable(Type type)
             return false;
         }
 
+        internal static string BuildCacheKey(string connectionString, string preferredLocations, bool useMultipleWriteLocations) => $"{useMultipleWriteLocations}{preferredLocations}{connectionString}";
```
brettsam: nit -- can we use a `|` (or something else) as a delimiter between the values? It'd be easier to parse if we need to debug this.

brettsam: Is there a reason that they are written in reverse-order? I'd assume `connectionString` would be first to sort more logically... although it probably doesn't really matter.

src/WebJobs.Extensions.CosmosDB/Services/CosmosDBService.cs
```diff
@@ -15,21 +15,33 @@ internal sealed class CosmosDBService : ICosmosDBService, IDisposable
         private bool _isDisposed;
         private DocumentClient _client;
 
-        public CosmosDBService(string connectionString, ConnectionMode? connectionMode, Protocol? protocol)
+        public CosmosDBService(string connectionString, ConnectionMode? connectionMode, Protocol? protocol, string preferredLocations, bool useMultipleWriteLocations)
```
brettsam: Can we change this constructor to take a connectionString and a ConnectionPolicy? Then the factory's job is to build that ConnectionPolicy? I have a feeling this is just going to keep growing over time...

src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs
```diff
@@ -102,6 +102,23 @@ public CosmosDBAttribute(string databaseName, string collectionName)
...
+        /// Optional.
+        /// Defines preferred locations (regions) for geo-replicated database accounts in the Azure Cosmos DB service.
+        /// Values should be comma-separated.
+        /// </summary>
+        /// <example>
+        /// PreferredLocations = "East US,South Central US,North Europe"
+        /// </example>
+        [AutoResolve]
+        public string PreferredLocations { get; set; }
+
```
brettsam: So now we have a mix of ConnectionPolicy values that we can set on either the global config (`ConnectionMode`, `ConnectionProtocol`) or the individual attributes (`UseMultipleWriteLocations`, `PreferredLocations`). Is that confusing? Should we support them all in both places, with attribute overriding config? I know the original issue was building up the cache with unique keys, but you have a way to do that now...

brettsam:
This came up at Ignite from several people -- you jumped on it before I could even bring it up with you :-)

src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs
ealsur: The logic I thought was that ConnectionMode and ConnectionProtocol are environment configurations, they depend on where the Function is running since they are affected by corporate firewalls for example.
While the new attributes are particular to each binding, you can have 2 Output bindings with different PreferredLocations for example.

src/WebJobs.Extensions.CosmosDB/Services/CosmosDBService.cs
ealsur: Good idea.

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
ealsur: Addressed 😄

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
ealsur: This cannot be done, trying to use string.Empty has a compile-time error since only compile-time values can be used.

src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs
brettsam: Yeah, that's right. Thanks for reminding me.

brettsam: Yeah, that's right. Thanks for reminding me.

src/WebJobs.Extensions.CosmosDB/CosmosDBUtility.cs
```diff
@@ -43,6 +45,47 @@ internal static async Task CreateDatabaseAndCollectionIfNotExistAsync(ICosmosDBS
...
+        internal static IEnumerable<string> ParsePreferredLocations(string preferredRegions)
+        {
+            if (string.IsNullOrEmpty(preferredRegions))
+            {
+                return Enumerable.Empty<string>();
+            }
+
+            return preferredRegions
+                .Split(',')
+                .Select((region) => region.Trim())
```
brettsam: `region` can't be null here, right? Can't remember if `Split()` will ever give you a null string...

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -137,7 +143,7 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
                 if (attribute.CreateLeaseCollectionIfNotExists)
                 {
                     // Not disposing this because it might be reused on other Trigger since Triggers could share lease collection
-                    ICosmosDBService service = _configProvider.GetService(leasesConnectionString);
+                    ICosmosDBService service = _configProvider.GetService(leasesConnectionString, attribute.PreferredLocations);
```
brettsam: This should probably used the resolved `PreferredLocations`, which will be used for a lookup, right? Would allow sharing of the service if it's configured the same for an output/input binding.

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -191,6 +198,21 @@ public async Task ValidParametersWithEnvironment_ConnectionMode_Succeed(Paramete
...
+        {
+            var nameResolver = new TestNameResolver();
+            nameResolver.Values["regions"] = "East US, North Europe,";
+
+            CosmosDBTriggerAttributeBindingProvider provider = new CosmosDBTriggerAttributeBindingProvider(_emptyConfig, nameResolver, _options, CreateExtensionConfigProvider(_options), _loggerFactory);
+
+            CosmosDBTriggerBinding binding = (CosmosDBTriggerBinding)await provider.TryCreateAsync(new TriggerBindingProviderContext(parameter, CancellationToken.None));
+
+            Assert.Equal(2, binding.DocumentCollectionLocation.ConnectionPolicy.PreferredLocations.Count);
+            Assert.Equal(2, binding.LeaseCollectionLocation.ConnectionPolicy.PreferredLocations.Count);
```
brettsam: Can you validate the actual PreferredLocation values here, just to be safe?

src/WebJobs.Extensions.CosmosDB/CosmosDBUtility.cs
ealsur: Not null but empty or with spaces, that's why I have the filter afterwards.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: Good catch!

PR #522, opened 2018-10-23
updating versions

brettsam:
Updating Extensions and CosmosDB versions as they have changed since last release. Since we no longer need to release each extension with every revision, I'm starting to separate out their versioning.


brettsam:
@paulbatum / @mathewc / @fabiocav  -- Just a heads up that I'll be getting the Extensions and CosmosDB packages ready. And that I've split off the versioning of Extensions so editing common.props doesn't affect it. Since we don't release everything at once (nothing else has changed), I didn't want to be revving all of the versions.

PR #526, opened 2018-10-28
Feature Proposal TimerSchedule w/TimeZone

davidrevoledo:
This is a feature proposal for the sdk extensions.

**Please feel free to ignore if it doesn't make any sense or if there is another way to resolve this I didn't realize** 

Currently I've facing a problem with Functions and `TimeTrigger`.

As I understand and  I've analyze you have 2 options to resolve when a trigger should be executed.

- By a ncron expression
- By a custom `TimerSchedule.cs` there is  a few in the sdk that are handy (Daily, Weekly and so on).

But I'm running distributed functions around the globe and they've to run in a specific time (Not in the region they're deployed but a time for the Company's HQ timezone).

I've realized the server return UTC when is running in Azure So for solve my problem I've created a custom `TimerSchedule` that takes timing and convert to a configured TimeZone and it is working.

However I think it would be handy for others scenarios, let's say that I've configured my function to 8:00 Pacific TimeZone and for any reason I move the function to other region with different Zone but This need to run the same time so, a code change is necessary.

The proposition is add The TimeZone to the `TimeTrigger` so any can add any custom logic agnostic for the zone and just extend that logic configuring a TimeZone. 

 I was running the code and realized that to get the difference between next ocurrence and local now this  method does the job `GetNextTimerInterval`. 

With a property `TimeZone` the package can detect if the next occurrence match no matter the local time.

So this

``` C#
 public class MyDailySchedule : DailySchedule
        {
            public MyDailySchedule()
                : base("8:00:00", "12:00:00", "22:00:00")
            {
            }
        }
```
become this

``` C#
   public class MyDailySchedule : DailySchedule
        {
            public MyDailySchedule()
                : base("8:00:00", "12:00:00", "22:00:00")
            {
                TimeZone = "Central America Standard Time";
            }
        }
```
and all should work smoothly.

If you are agree I can complete the unit tests for all the cases including the configured time zone is the same as local.

Thanks.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Bindings/TimerTriggerBinding.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Bindings/TimerTriggerBinding.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Bindings
         {
             if (context == null)
             {
-                throw new ArgumentNullException("context");
+                throw new ArgumentNullException(nameof(context));
             }
             return Task.FromResult<IListener>(new TimerListener(_attribute, _schedule, _timerName, _options, context.Executor, _logger, _scheduleMonitor));
         }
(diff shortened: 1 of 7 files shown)


brettsam:
Have you tried using the WEBSITE_TIME_ZONE environment variable? Does that let you do what you're looking for? https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer#cron-time-zones

davidrevoledo:
@brettsam I'm aware of that configuration however I was looking for something to re distribute w/o changing any configuration or code, I've functions running in a non-consumption plan mode changing the machine timezone can potentially break others things and other app services so that's not a solution that suit for me.

Issue #536, opened 2018-12-03
HttpTrigger can not bind from JSON array or throw 500 error with C# Enum


shibayan:
If JSON contains an array, null is returned as a result.

Cause JsonInputFormatter is not used for model binding. The implementation seems to expect that JsonInputFormatter will be used.

#### Repro steps

1. Create Azure Functions v2 project with HttpTrigger
2. Use Custom class (include C# array) binding in HttpTrigger
3. Post to Function with JSON payload include array

#### Expected behavior

Use ASP.NET Core InputFormatters for Json model binding.

#### Actual behavior

InputFormatters collection is empty.

![image](https://user-images.githubusercontent.com/1356444/49359241-ed75db80-f718-11e8-8016-bfc751f61282.png)

#### Known workarounds

Create IWebJobsStartup and add JsonInputFormatter manually.

```csharp
public class JsonWebJobsStartup : IWebJobsStartup
{
    public void Configure(IWebJobsBuilder builder)
    {
        builder.Services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<MvcOptions>, MvcJsonMvcOptionsSetup>());
    }
}
```

Maybe related to these issues. #485 #486 #511 #523 https://github.com/Azure/azure-functions-host/issues/3370

#### Related information 

* Function Runtime Version: 2.0.12175.0
* Microsoft.Azure.WebJobs.Extensions.Http v3.0.0

shibayan:
I have confirmed that it has been fixed with Azure Function v3 (Runtime Version `3.0.12930.0`).

Issue #523, opened 2018-10-24
Azure functions and HttpTrigger binding to POCO with Enums fails to deserialize


mannafhaque:
After upgrading Azure Functions App from .Net Framework 4.7 to .Net Standard 2.0, I am 
having issue deserializing POCO object with enum:

Having the same issue as [HttpTrigger binding to POCO fails on Enums](https://github.com/Azure/azure-webjobs-sdk-extensions/issues/486) 

I have inclued a proof of concept for this bug in this respository: https://github.com/mannafhaque/FunctionsEnumPOC.  

Help would be much appreciated. 

```public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] Customer customer, HttpRequest req, ILogger log)```

````Executing 'FunctionBindingToPocoObjectWithEnum' (Reason='This function was programmatically called via the host APIs.', Id=436d04c9-9bc0-41b5-8291-164f58b00edc)
Executed 'FunctionBindingToPocoObjectWithEnum' (Failed, Id=436d04c9-9bc0-41b5-8291-164f58b00edc)
System.Private.CoreLib: Exception while executing function: FunctionBindingToPocoObjectWithEnum. Microsoft.Azure.WebJobs.Host: One or more errors occurred. 
(Exception binding parameter 'customer') (Exception binding parameter 'req'). 
Exception binding parameter 'customer'. System.Private.CoreLib: Invalid cast from 'System.String' to 'Common.Models.CustomerType'```

Issue #511, opened 2018-10-03
Model binding for Arrays is broken in V2


ealsur:
Binding any trigger to a POCO (C#) model that contains arrays is broken in V2, while it works in V1.

#### Repro steps

Repro repo available here: https://github.com/ealsur/arraybinding 

1. Running the [V1 project](https://github.com/ealsur/arraybinding/tree/master/BindingSampleV1) exposes an HTTP Endpoint for an HTTPTrigger.

2. Send this sample payload using Postman on a POST body:
    [{
        "id":"test",
        "name":"test",
        "city": "test"
        
    },{
        "id":"tes2t",
        "name":"test2",
        "city": "test2"
        
    }]

3. Runtime executes correctly:
    [3/10/2018 13:02:19] Executing 'Function1' (Reason='This function was programmatically called via the host APIs.', Id=568da040-e708-451a-a133-574778c27bf7)
    [3/10/2018 13:02:19] Received 2 documents
    [3/10/2018 13:02:19] Function completed (Success, Id=568da040-e708-451a-a133-574778c27bf7, Duration=140ms)
    [3/10/2018 13:02:19] Executed 'Function1' (Succeeded, Id=568da040-e708-451a-a133-574778c27bf7)
    [3/10/2018 13:02:19] Executed HTTP request: {
    [3/10/2018 13:02:19]   "requestId": "23200401-20b9-495e-b942-f052422a5dc9",
    [3/10/2018 13:02:19]   "method": "POST",
    [3/10/2018 13:02:19]   "uri": "/api/Function1",
    [3/10/2018 13:02:19]   "authorizationLevel": "Anonymous",
    [3/10/2018 13:02:19]   "status": "OK"
    [3/10/2018 13:02:19] }

4. Running the [V2 project](https://github.com/ealsur/arraybinding/tree/master/BindingsSampleV2) that also exposes the same HTTPTrigger with the same binding (`MyClass[]`) and sending the same payload, fails with this error:

    Function Runtime Version: 2.0.12115.0
    [3/10/2018 13:03:01] Executing 'Function1' (Reason='This function was programmatically called via the host APIs.', Id=1eac5217-8d76-4f11-a726-9732a57135cb)
    [3/10/2018 13:03:01] Executed 'Function1' (Failed, Id=1eac5217-8d76-4f11-a726-9732a57135cb)
[...]

Issue #486, opened 2018-09-11
HttpTrigger binding to POCO fails on Enums


dlalex83:
I've recently updated my Azure Functions app to v2.0.12050.0 which upgrade the webjobs-sdk-extensions from 3.0.0-beta5 to 3.0.0-beta8.

I have several functions using HttpTrigger in POST binding to a POCO object.
The POCO has an enum property and before the upgrade I was able to post content into it and the sdk was correctly binding an int value to the enum. 
See below for more details.

POCO:

```
public class Message
{
     public Region RegionId { get; set; }
     public string Content { get; set; }
}
```

Request payload:
```
{
   RegionId: 1,
   Content: "test"
}
```

Function declaration:
`public static async Task ProcessMessage([HttpTrigger(AuthorizationLevel.Anonymous, "POST")]Message message);`

#### Expected behavior

The message property should be populated from the payload and the enum property Region correctly set.

#### Actual behavior

Invocation breaks with the error: 

> System.Private.CoreLib: Exception while executing function: AddRequestTweakingWebsiteRule. Microsoft.Azure.WebJobs.Host: Exception binding parameter 'req'. System.Private.CoreLib: Invalid cast from 'System.String' to 'LoveTheSales.Shared.Enums.Region'.

Stacktrace: 
```
 System.InvalidOperationException: Exception binding parameter 'req' ---> System.InvalidCastException: Invalid cast from 'System.String' to 'Shared.Enums.Region'.
   at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   at System.String.System.IConvertible.ToType(Type type, IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType)
   at Microsoft.Azure.WebJobs.Extensions.Http.HttpTriggerAttributeBindingProvider.HttpTriggerBinding.ConvertValueIfNecessary(Object value, Type targetType) in C:\azure-webjobs-sdk-extensions\src\WebJobs.Extensions.Http\HttpTriggerAttributeBindingProvider.cs:line 413
[...]

PR #539, opened 2018-12-10
Adding JsonInputFormatter configuration

shibayan:
In order to use JsonInputFormatter correctly, register necessary services in DI. #536 

Related issues #485 #486 #511 #523 Azure/azure-functions-host#3370

### DIFF
--- a/src/WebJobs.Extensions.Http/Config/HttpWebJobsBuilderExtensions.cs
+++ b/src/WebJobs.Extensions.Http/Config/HttpWebJobsBuilderExtensions.cs
@@ -4,12 +4,14 @@
 using System;
 using System.Net.Http.Formatting;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Formatters.Json.Internal;
 using Microsoft.AspNetCore.Mvc.WebApiCompatShim;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Extensions.Http;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Hosting
 {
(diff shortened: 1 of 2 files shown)


shibayan:
Please review or triage or etc... I think we can fixed many issues and happy.
@fabiocav @jeffhollan 

Thank you.

jeffhollan:
Thanks @shibayan - we'll do some investigation in an upcoming sprint to see if this fix may provide what we need without any unintended consequences.  Appreciate you opening the PR.

shibayan:
@jeffhollan Thank you for the reply! I will wait for it.

shibayan:
Due to the difference configuration in Dependency Injection for unit test, `JsonInputFormatter` is already used in unit test, but it seems that it is not used in products. I think the test cases is broken.

fabiocav:
@shibayan sorry about the delay on this. Though this does address the problem on the surface, the actual issue is caused by a bug with the MVC options configuration done at the host level and not being properly surfaced to the JobHost scope. That would be the right bug to fix. I can open an issue with more details on this.

shibayan:
@fabiocav Thank you for the reply. As you say, this is caused by the fact that the MvcOptions that should have been set are not applied to the current scope.
(I was not able to investigate the cause to there.)

I also hope that this problem will be solved fundamentally.

shibayan:
I have confirmed that it has been fixed with Azure Function v3 (Runtime Version `3.0.12930.0`).

Works unless you use v2 compatibility mode. I have already moved to v3, so this PR is no longer necessary. 
Thank you.

ionescudan12:
Hi @shibayan, I have tried to use the formatter to return validation errors like it does on web api with ModelState.IsValid on objects when a function endpoint deserializes the object, but I can't seem to find a proper way to do this besides forcing the validation of the objects again after they are deserialized into a concrete object. I have not found a way in the code nor the examples for this

I'll exemplify

What I wanted:
```csharp
        [FunctionName(nameof(Test))]
        public async Task<IActionResult> Test(
            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "test")] Payload payload,
            HttpRequest req)
        {
            // this sort of logic made more sense, in a way using how the payload is being serialized
            if (!ModelState.IsValid) return new BadRequestObjectResult(ModelState);

            return new OkObjectResult();
        }
```

What I did but it's clearly re-evaluating the payload:
```csharp
        [FunctionName(nameof(Test))]
        public async Task<IActionResult> Test(
            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "test")] Payload payload,
            HttpRequest req)
        {
            var validationErrors = payload.Validate(new ValidationContext(payload)).ToList();
            if (validationErrors.Any())
                return new BadRequestObjectResult($"Found {validationErrors.Count} configuration error(s): {string.Join(',', validationErrors)}");

            return new OkObjectResult();
        }
```

Payload implements IValidatableObject, for clearing that out

How can I perform this since the MvcOptions were added to v3 functions?

PR #542, opened 2019-01-02
Catch all routes in proxies should not override /admin routes.

safihamid:
https://github.com/Azure/azure-functions-host/issues/2590

### DIFF
--- a/src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
+++ b/src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
@@ -51,6 +51,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
 
         public Task RouteAsync(RouteContext context)
         {
+            // /admin/* routes should not be allowed to be overriden by proxies.
+            if (context.HttpContext.Request.Path.StartsWithSegments(new PathString("/admin"), System.StringComparison.OrdinalIgnoreCase))
+            {
+                return Task.CompletedTask;
+            }
+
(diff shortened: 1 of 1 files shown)


paulbatum:
If I understand this change correctly does it really resolve Azure/azure-functions-host#2590 ? A catchall route would not start with the string "admin". Is this fix just handling the case where the user attempts to override admin through proxies, not using a catch-all?

safihamid:
@paulbatum other than /admin endpoints we are fine and the user can do a catch-all for all other routes. This is the behavior we had in V1 as well, Am i Missing something?

paulbatum:
So putting aside proxies for a second, here is an example of the case that I think @fabiocav was worried about when he filed that issue:

```csharp        
public static async Task<IActionResult> Run(
            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = "{*test}")] HttpRequest req,
            ILogger log)
```

If you do this, then you'll override admin endpoints such as `/admin/host/status`. You can try this yourself locally very easily.

Extending this scenario to proxies, I suspect you should be worried about this case:

```javascript
{
    "proxies": {
        "proxy_test": {
            "debug": true,
            "matchCondition": {
                "route": "/{*test}"
            },
            "backendUri": "https://somehost.azurewebsites.net/admin/{test}"
        }
    }
}
```

safihamid:
Thanks @paulbatum for the examples. Verified the function example works with this fix. and /admin endpoints can still be reached even with the Route specified above in function code and removing routePrefix.

For the proxy example, the example above was never supported even in V1. in V1 we were returning infinite loop detected and now with this fix we return 404 which is even better I think.

PR #553, opened 2019-02-13
[TimerTrigger] improving logs

### LINKED ISSUE #550: Improve timer trigger logging
If the timer trigger fails to fire for whatever reason, more logs would help us diagnose the issues. Most of these can be verbose so that they're only logged to our system logs and won't clog up users with normal verbosity:
- What CRON expression did we actually get? Right now we log the next schedule, but not what the CRON was before we parsed it.
- Log when we start a timer, and exactly what length that timer was for. If a timer fails to fire, we know to look at thread starvation/ CPU/ etc. But at least we know we started the timer as we expected.
- Include the actual function name when writing logs (i.e. the name from metadata rather than `Function 'FunctionApp3.Function1.Run' is past due on startup. Executing now.`). Ideally, include these logs in the FunctionName column in system logs.
- If a timer is running because it's past due, put it into the `Reason` for the function execution. Right now you have to search up for a message (without a FunctionName or FunctionInvocationId) to figure out why a timer fired at an odd time. Same thing for `RunOnStartup`.
- Include the timezone offset (or maybe include UTC in parens) in the `The next 5 occurrences...` message. We log the time zone higher up, but it'd be good to include this all in one message.

brettsam:
Fixes #550

### Using function's name in startup logs when [FunctionName] is used
```
namespace WebJobsSandbox.One.Two.Three
{
    public static class TimerSamples
    {    
        [FunctionName("MyTimer")]
        public static void CronJob([TimerTrigger("*/15 * * * * *")] TimerInfo timerInfo)
        {
        }
    }
}
```
Before: `The next 5 occurrences of the 'WebJobsSandbox.One.Two.Three.TimerSamples.CronJob' schedule`...
After: `The next 5 occurrences of the 'MyTimer' schedule`...

### Using function's short name in startup logs if [FunctionName] is *not* used
```
namespace WebJobsSandbox.One.Two.Three
{
    public static class TimerSamples
    {
        public static void CronJob([TimerTrigger("*/15 * * * * *")] TimerInfo timerInfo)
        {
        }
    }
}
```
Before: `The next 5 occurrences of the 'WebJobsSandbox.One.Two.Three.TimerSamples.CronJob' schedule`...
After: `The next 5 occurrences of the 'TimerSamples.CronJob' schedule`...

### Logging "schedule" changes
The schedule has been changed to be more easily parse-able and include more information:
- Including Schedule information in string. Note that the Cron library we use automatically evaluates the string so "*/15" -> "0,15,30,45".
- Using 2-digit months, days, and hours
- Using 24-hour clock
- Using UTC offset (if necessary)
- Including UTC time for schedule (if necessary)

Before:
```
The next 5 occurrences of the 'WebJobsSandbox.One.Two.Three.TimerSamples.CronJob' schedule will be:
2/13/2019 9:17:15 AM
2/13/2019 9:17:30 AM
2/13/2019 9:17:45 AM
2/13/2019 9:18:00 AM
2/13/2019 9:18:15 AM
```

After, not UTC:
```
The next 5 occurrences of the 'MyTimer' schedule (Cron: '0,15,30,45 * * * * *') will be:
02/13/2019 08:57:45-08:00 (02/13/2019 16:57:45Z)
02/13/2019 08:58:00-08:00 (02/13/2019 16:58:00Z)
02/13/2019 08:58:15-08:00 (02/13/2019 16:58:15Z)
02/13/2019 08:58:30-08:00 (02/13/2019 16:58:30Z)
02/13/2019 08:58:45-08:00 (02/13/2019 16:58:45Z)
```

After, UTC:
```
The next 5 occurrences of the 'MyTimer' schedule (Cron: '0,15,30,45 * * * * *') will be:
02/13/2019 17:07:00Z
02/13/2019 17:07:15Z
02/13/2019 17:07:30Z
02/13/2019 17:07:45Z
02/13/2019 17:08:00Z
```

~~### Logging warning if RunOnStartup (`Warning` level)
New: `Function 'MyTimer' has 'RunOnStartup' set to 'true'. This property is intended for testing purposes only and may result in unintended behavior if used in production.`~~

### Logging details at *execution time* if RunOnStartup or IsPastDue (`Information` level)
These TriggerDetail log have the proper `FunctionName` and `FunctionInvocationId` set so they will be returned via AppInsights queries looking for traces on this invocation. This functionality is already built into the host, we're just going to start using it now to provide more context.

New: `Trigger Details: UnscheduledInvocationReason: RunOnStartup`
New: `Trigger Details: UnscheduledInvocationReason: IsPastDue, OriginalSchedule: 2019-02-13T08:55:56.0531482-08:00`

### Logging .NET timer creation (`Debug` level)
New: `Timer for 'MyTimer' started with interval '00:00:14.9854675'.`

### Adding schedule to "time zone" logging at startup (`Debug` level)
Admittedly this isn't as important now that it's also logged with the future schedule, but...

Before: `The 'WebJobsSandbox.One.Two.Three.TimerSamples.CronJob' timer is using the local time zone: '(UTC-08:00) Pacific Time (US & Canada)'`
After:  `The 'MyTimer' timer is using the schedule 'Cron: '0,15,30,45 * * * * *'' and the local time zone: '(UTC-08:00) Pacific Time (US & Canada)'`

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Timers;
@@ -14,12 +15,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
     [Singleton(Mode = SingletonMode.Listener)]
     internal sealed class TimerListener : IListener
     {
+        public const string UnscheduledInvocationReasonKey = "UnscheduledInvocationReason";
+        public const string OriginalScheduleKey = "OriginalSchedule";
+
         private readonly TimerTriggerAttribute _attribute;
         private readonly TimersOptions _options;
         private readonly ITriggeredFunctionExecutor _executor;
         private readonly ILogger _logger;
         private readonly CancellationTokenSource _cancellationTokenSource;
-        private readonly string _timerName;
+
+        // _functionShortName is often the [FunctionName] value and used for logging, 
+        // while _timerLookupName is the fully-qualified method name and used for lookups
+        private readonly string _functionShortName;
+        private readonly string _timerLookupName;
 
         // Since Timer uses an integer internally for it's interval,
         // it has a maximum interval of 24.8 days.
@@ -31,16 +39,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
         private bool _disposed;
         private TimeSpan _remainingInterval;
 
-        public TimerListener(TimerTriggerAttribute attribute, TimerSchedule schedule, string timerName, TimersOptions options, ITriggeredFunctionExecutor executor, ILogger logger, ScheduleMonitor scheduleMonitor)
+        public TimerListener(TimerTriggerAttribute attribute, TimerSchedule schedule, string timerName, TimersOptions options, ITriggeredFunctionExecutor executor,
+            ILogger logger, ScheduleMonitor scheduleMonitor, string functionShortName)
         {
             _attribute = attribute;
-            _timerName = timerName;
+            _timerLookupName = timerName;
             _options = options;
             _executor = executor;
             _logger = logger;
             _cancellationTokenSource = new CancellationTokenSource();
             _schedule = schedule;
             ScheduleMonitor = _attribute.UseMonitor ? scheduleMonitor : null;
+            _functionShortName = functionShortName;
         }
 
         internal static TimeSpan MaxTimerInterval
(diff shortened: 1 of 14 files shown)


src/WebJobs.Extensions/Extensions/Timers/Scheduling/DailySchedule.cs
```diff
@@ -61,13 +61,13 @@ public override DateTime GetNextOccurrence(DateTime now)
                 // no more occurrences for today, so start back at the beginning of the
                 // the schedule tomorrow
                 TimeSpan nextTime = schedule[0];
-                DateTime nextOccurrence = new DateTime(now.Year, now.Month, now.Day, nextTime.Hours, nextTime.Minutes, nextTime.Seconds);
+                DateTime nextOccurrence = new DateTime(now.Year, now.Month, now.Day, nextTime.Hours, nextTime.Minutes, nextTime.Seconds, now.Kind);
```
brettsam: Now that I'm writing out the timezone offset, it exposed a little bug in how we created these new DateTimes. They'd end up with a Kind of `Unspecified`, which meant no timezone offset would appear when formatting the string. This changes it to copy the input value's `Kind` (which, for `DateTime.Now` is `Local`).

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -104,18 +114,19 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
-                await InvokeJobFunction(now, isPastDue: true);
+                _logger.LogDebug($"Function '{_functionShortName}' is past due on startup. Executing now.");
+                await InvokeJobFunction(now, isPastDue: true, originalSchedule: ScheduleStatus.Next);
             }
             else if (_attribute.RunOnStartup)
             {
                 // The job is configured to run immediately on startup
-                _logger.LogDebug($"Function '{_timerName}' is configured to run on startup. Executing now.");
+                _logger.LogDebug($"Function '{_functionShortName}' is configured to run on startup. Executing now.");
+                _logger.LogWarning($"Function '{_functionShortName}' has '{nameof(TimerTriggerAttribute.RunOnStartup)}' set to 'true'. This property is intended for testing purposes only and may result in unintended behavior if used in production.");
```
mathewc: I don't think we should add this one. For usages outside of Functions (long running continuous WebJobs) this isn't true and using RunOnStartup is perfectly reasonable. It's only in the context of Azure Functions and short lived host instances that people confuse themselves with this.

brettsam: Fair enough -- I figured there was a reason we hadn't added it already. I had a CRI on this last week (and @mhoeger suggested it as well) but I wasn't thinking of WebJobs. The TriggerDetails noting that the invocation happened b/c of RunOnStartup should help people identify this themselves (I hope).

mhoeger: @brettsam - we'd have to do some hard-coding, but we could also check for that log in a detector. Any chance we could configure that message to be a little more query-friendly?

brettsam: This warning is now gone -- but we still log the details if it's been run b/c of RunOnStartup.

The log looks like: `Trigger Details: UnscheduledInvocationReason: RunOnStartup`. So we should look for `UnscheduledInvocationReason` -- that should also have the FunctionName and FunctionInvocationId associated with it.

PR #566, opened 2019-04-29
Enabling Multi Master for leases in Cosmos DB Trigger

ealsur:
This PR adds support for the Cosmos DB Trigger to use Multi Master for the leases container. Previously, while the users could use PreferredLocations to specify reads from a particular region, writes for the leases state was being sent to the main hub region.

For users with Multi Master accounts, they can leverage local writes on any region, so this new attribute exposed in the Trigger can help them take advantage of that.

Additionally, bumping the version of Cosmos DB SDK and Change Feed SDK to get the latest fixes/improvements.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
@@ -144,8 +144,17 @@ namespace Microsoft.Azure.WebJobs
         /// Values should be comma-separated.
         /// </summary>
         /// <example>
-        /// PreferredLocations = "East US,South Central US,North Europe"
+        /// PreferredLocations = "East US,South Central US,North Europe".
         /// </example>
         public string PreferredLocations { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Enable to use with Multi Master accounts.
+        /// </summary>
+        /// <remarks>
+        /// This setting only applies to the Leases collection, as there are no write operations done to the monitored collection.
+        /// </remarks>
+        public bool UseMultipleWriteLocations { get; set; }
     }
 }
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -144,8 +144,14 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
...
         /// </summary>
         /// <example>
-        /// PreferredLocations = "East US,South Central US,North Europe"
+        /// PreferredLocations = "East US,South Central US,North Europe".
         /// </example>
         public string PreferredLocations { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Enable to use with Multi Master accounts.
```
brettsam: Maybe add something that this only applies to the lease collection?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -130,6 +130,8 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
                     leaseCollectionLocation.ConnectionPolicy.PreferredLocations.Add(location);
                 }
 
+                leaseCollectionLocation.ConnectionPolicy.UseMultipleWriteLocations = attribute.UseMultipleWriteLocations;
```
brettsam: Why do you have to set this here and with the client? What if one was true and one was false? 

I'm mostly just curious :-) -- since it doesn't seem like that could happen here...

ealsur: Honestly, just in case. The CFP Library has both settings available and in case there is some internal logic checking one and not the other, I rather set it on both as it doesn't really hurt.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
ealsur: Done

PR #585, opened 2019-08-07
[Http] replacing EnableRewind() with EnableBuffering()

brettsam:
`EnableRewind()` went internal in ASP.NET Core 3.0 preview 7. But there's a direct replacement available in 2.x as well -- https://github.com/aspnet/AspNetCore/issues/12508#issuecomment-515601441.

Getting this in unblocks a bunch of .NET Core 3.0 tests in Functions.


brettsam:
Internally, it just calls `EnableRewind()` -- https://github.com/aspnet/AspNetCore/blob/4ef204e13b88c0734e0e94a1cc4c0ef05f40849e/src/Http/Http/src/Extensions/HttpRequestRewindExtensions.cs#L23

brettsam:
I ran the Functions v2 tests locally with this change and all looks good.

sc1994:
`context.Request.EnableBuffering();` Doesn't seem to save my request body.
at `3.0.100-preview8-013656`  version.

PR #586, opened 2019-08-21
Added special conversion code for guids in Http triggers and supporting tests

### LINKED ISSUE #2960: Unable to use "guid" as an HttpTrigger route parameter (Invalid cast from 'System.String' to 'System.Guid')
When trying to pass a guid as a parameter binding within a route definition the following exception is thrown

`Exception while executing function: TestParams -> One or more errors occurred. -> Exception binding parameter 'req' -> Invalid cast from 'System.String' to 'System.Guid'.`

#### Investigative information

Please provide the following:

- Timestamp: 2018-06-07T11:27:47.578
- Function App version (1.0 or 2.0-beta): 1.0.12.1
- Function App name: vos6functiondev
- Function name(s) (as appropriate): GetPendingDamages
- Invocation ID:  (requestid): 65ead198-8c37-438d-b9a5-ac17165f1e41
- Region: UK West

#### Repro steps

Provide the steps required to reproduce the problem:

- For an HttpTrigger, configure a route parameter binding for a Guid

```
[FunctionName("GetPendingDamages")]
        public static async Task<HttpResponseMessage> Run(
            [HttpTrigger(AuthorizationLevel.Function, "get", Route = "GetPendingDamages/{stayId:guid}")]
            HttpRequestMessage req,
            Guid stayId,
            TraceWriter log)
        {...}
```
- Invoke the fuction as a GET request
- The following exception will be thrown
` "message": "Exception while executing function: TestParams -> One or more errors occurred. -> Exception binding parameter 'req' -> Invalid cast from 'System.String' to 'System.Guid'.",`

#### Expected behavior

Provide a description of the expected behavior.

I would expect to either
- have a valid Guid value available as an input parameter
- receive a 400 Bad Request, in case the value could not be parsed as a Guid

#### Known workarounds

Provide a description of any known workarounds.

You can do a `Guid.TryParse()` in code, but it would be nice to be able to have the Guid bindings available like in Web API 2

as it has been reported here 
http://dontcodetired.com/blog/post/Dynamic-Binding-in-Azure-Functions-with-Imperative-Runtime-Bindings
and here https://stackoverflow.com/questions/46743180/is-attribute-routing-possible-in-azure-functions

Thanks :)

gzuber:
Resolves https://github.com/Azure/azure-functions-host/issues/2960

This addresses GUIDs not being handled properly when being bound in the http extension, resulting in a 500 error for any function that expects a Guid type. I've also added tests for all data type route constraints so we know if they break in the future.

This should fix both Guids in route constraints as well as Guids in POCOs

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -403,17 +403,24 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
             {
                 if (value != null && !targetType.IsAssignableFrom(value.GetType()))
                 {
+                    var underlyingTargetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
+
                     var jObject = value as JObject;
                     if (jObject != null)
                     {
                         value = jObject.ToObject(targetType);
                     }
+                    else if (underlyingTargetType == typeof(Guid) && value.GetType() == typeof(string))
+                    {
+                        // Guids need to be converted by their own logic
+                        // Intentionally throw here on error to standardize behavior
+                        value = Guid.Parse((string)value);
+                    }
                     else
                     {
                         // if the type is nullable, we only need to convert to the
                         // correct underlying type
-                        targetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
-                        value = Convert.ChangeType(value, targetType);
+                        value = Convert.ChangeType(value, underlyingTargetType);
                     }
                 }
 
--- a/test/WebJobs.Extensions.Http.Tests/HttpTriggerBindingTests.cs
+++ b/test/WebJobs.Extensions.Http.Tests/HttpTriggerBindingTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Http;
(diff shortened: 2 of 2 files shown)


src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -408,6 +408,12 @@ private static object ConvertValueIfNecessary(object value, Type targetType)
                     {
                         value = jObject.ToObject(targetType);
                     }
+                    else if (targetType == typeof(Guid) && value.GetType() == typeof(string))
```
mathewc: Below we're doing Nullable.GetUnderlyingType - do we need this here as well? Does your new code work if the method binds to Guid?

mathewc (approved):
Approved, but recommend adding a few nullables to your test case as I mentioned.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
gzuber: That's a really good point -- I swear I had tried out that scenario but evidently not. I fixed that and added test cases to cover nullables like you suggested.

mathewc (approved):

andyblack19:
Hi folks, when is this likely to make it into production?

mathewc:
@gzuber - what release will this be a part of?

gzuber:
@andyblack19 -- sorry for the delay! We're trying to get this in for the next host release. https://github.com/Azure/azure-functions-host/pull/5100

andyblack19:
Awesome, thanks @gzuber

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -403,17 +403,24 @@ private static object ConvertValueIfNecessary(object value, Type targetType)
...
+                        // Intentionally throw here on error to standardize behavior
+                        value = Guid.Parse((string)value);
+                    }
                     else
                     {
                         // if the type is nullable, we only need to convert to the
                         // correct underlying type
-                        targetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
-                        value = Convert.ChangeType(value, targetType);
+                        value = Convert.ChangeType(value, underlyingTargetType);
```
arkiaconsulting: Would it be possible to use TypeDescriptor.GetConverter and use the resulted converter instead ? It would cover the case when we use ValueObjects in DDD.

lukasvosyka:
Hi, 

I stumbled across this, since it seems to be also not working with Runtime 3. Tested with Function Runtime Version: 3.0.13353.0

jnhaffey:
There does seem to be an issue still with Function Runtime Version: 3.0.14492.0
```
        [FunctionName("XXXXX")]
        public async Task<IActionResult> XXXXX(
            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "/{orgId:guid}")]
            Guid orgId, HttpRequest req)
        {
            throw new NotImplementedException();
        }
```
Results in the following error when launching locally:
_The 'XXXXX' function is in error: Microsoft.Azure.WebJobs.Host: Error indexing method 'XXXXX'. Microsoft.Azure.WebJobs.Extensions.Http: Can't bind HttpTriggerAttribute to type 'System.Guid'._

LeeIsMe77:
Bump... because despite the team acknowledging this is fixed... it is not.  Guid's absolutely do not work, as demonstrated by [#jnhaf](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/586#issuecomment-702793415).... 3 years ago.

megaromb:
Let me confirm that today the issue has not been resolved yet (4+ years)

```
Azure Functions Core Tools
Core Tools Version:       4.0.5390 Commit hash: N/A  (64-bit)
Function Runtime Version: 4.25.3.21264
```

```
 Microsoft.Azure.WebJobs.Host: Error indexing method 'SomeMethod'. Microsoft.Azure.WebJobs.Host: Cannot bind parameter 'parameterName' to type Guid. Make sure the parameter Type is supported by the binding. If you're using binding extensions (e.g. Azure Storage, ServiceBus, Timers, etc.) make sure you've called the registration method for the extension(s) in your startup code (e.g. builder.AddAzureStorage(), builder.AddServiceBus(), builder.AddTimers(), etc.).
```

So this merged PR doesn't resolve the issue.

@gzuber are you aware of this issue? is it possible to reopen it? just something to pay your team's attention. 

Thanks

PR #588, opened 2019-09-10
Adding CosmosDB support for runtime scaling with unit tests.

### LINKED ISSUE #591: Update CosmosDB extension to support Scale Monitor Interface
Part of Azure/azure-functions-host#4821

The CosmosDB extension needs to implement IScaleMonitor in order to support the new runtime scale monitoring feature of premium functions.

gzuber:
Resolves #591

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
@@ -2,6 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -9,13 +12,15 @@ using Microsoft.Azure.Documents;
 using Microsoft.Azure.Documents.ChangeFeedProcessor;
 using Microsoft.Azure.Documents.ChangeFeedProcessor.Monitoring;
 using Microsoft.Azure.Documents.ChangeFeedProcessor.PartitionManagement;
+using Microsoft.Azure.WebJobs.Extensions.CosmosDB.Trigger;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 {
-    internal class CosmosDBTriggerListener : IListener, Microsoft.Azure.Documents.ChangeFeedProcessor.FeedProcessing.IChangeFeedObserverFactory
+    internal class CosmosDBTriggerListener : IListener, IScaleMonitor<CosmosDBTriggerMetrics>, Microsoft.Azure.Documents.ChangeFeedProcessor.FeedProcessing.IChangeFeedObserverFactory
     {
         private const int ListenerNotRegistered = 0;
         private const int ListenerRegistering = 1;
@@ -32,18 +37,40 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
         private readonly IHealthMonitor _healthMonitor;
         private IChangeFeedProcessor _host;
         private ChangeFeedProcessorBuilder _hostBuilder;
+        private ChangeFeedProcessorBuilder _workEstimatorBuilder;
+        private IRemainingWorkEstimator _workEstimator;
         private int _listenerStatus;
+        private string _functionId;
+        private ScaleMonitorDescriptor _scaleMonitorDescriptor;
+
+        private static readonly Dictionary<string, string> KnownDocumentClientErrors = new Dictionary<string, string>()
+        {
+            { "Resource Not Found", "Please check that the CosmosDB collection and leases collection exist and are listed correctly in Functions config files." },
+            { "The input authorization token can't serve the request", string.Empty },
+            { "The MAC signature found in the HTTP request is not the same", string.Empty },
+            { "Service is currently unavailable.", string.Empty },
+            { "Entity with the specified id does not exist in the system.", string.Empty },
+            { "Subscription owning the database account is disabled.", string.Empty },
+            { "Request rate is large", string.Empty },
+            { "PartitionKey value must be supplied for this operation.", "We do not support lease collections with partitions at this time. Please create a new lease collection without partitions." },
+            { "The remote name could not be resolved:", string.Empty },
+            { "Owner resource does not exist", string.Empty },
+            { "The specified document collection is invalid", string.Empty }
+        };
 
         public CosmosDBTriggerListener(ITriggeredFunctionExecutor executor,
+            string functionId,
             DocumentCollectionInfo documentCollectionLocation,
             DocumentCollectionInfo leaseCollectionLocation,
             ChangeFeedProcessorOptions processorOptions,
             ICosmosDBService monitoredCosmosDBService,
             ICosmosDBService leasesCosmosDBService,
-            ILogger logger)
+            ILogger logger,
+            IRemainingWorkEstimator workEstimator = null)
         {
             this._logger = logger;
             this._executor = executor;
+            this._functionId = functionId;
             this._hostName = Guid.NewGuid().ToString();
 
             this._monitorCollection = documentCollectionLocation;
@@ -52,6 +79,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             this._monitoredCosmosDBService = monitoredCosmosDBService;
             this._leasesCosmosDBService = leasesCosmosDBService;
             this._healthMonitor = new CosmosDBTriggerHealthMonitor(this._logger);
+
+            this._workEstimator = workEstimator;
+
+            this._scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-CosmosDBTrigger-{_monitorCollection.DatabaseName}-{_monitorCollection.CollectionName}".ToLower());
+        }
+
+        public ScaleMonitorDescriptor Descriptor
+        {
+            get
+            {
+                return _scaleMonitorDescriptor;
+            }
         }
 
         public void Cancel()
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -54,6 +75,14 @@ internal class CosmosDBTriggerListener : IListener, Microsoft.Azure.Documents.Ch
             this._healthMonitor = new CosmosDBTriggerHealthMonitor(this._logger);
         }
 
+        public string Id
+        {
+            get
+            {
+                return $"CosmosDBTrigger-{_hostName}-{_monitorCollection.DatabaseName}-{_monitorCollection.CollectionName}".ToLower();
```
mathewc: Please follow the same pattern as in QueueListener for this - e.g. `return $"{_functionId}-QueueTrigger-{_queue.Name}".ToLower();`

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
         }
+
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await GetMetricsAsync();
+        }
+
+        internal async Task<CosmosDBTriggerMetrics> GetMetricsAsync(IRemainingWorkEstimator workEstimator)
+        {
+            _workEstimator = workEstimator;
```
mathewc: Don't do this. Instead just take the param value and pass it into the GetMetricsAsync impl. You should have the 2 interface impls calling an internal GetMetricsAsync method taking the estimator. Your unit tests then call the internal one with their mock

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+
+        public async Task<CosmosDBTriggerMetrics> GetMetricsAsync()
+        {
+            int partitionCount = 0;
+            long remainingWork = 0;
+            IReadOnlyList<RemainingPartitionWork> workPerPartition = null;
+
+            try
+            {
+                InitializeBuilder();
```
mathewc: Recommend encapsulating this builder init and creation of the estimator into a single GetWorkEstimatorAsync method

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+
+                if (_workEstimator == null)
+                {
+                    _workEstimator = await this._hostBuilder.BuildEstimatorAsync();
+                }
+
+                workPerPartition = await this._workEstimator.GetEstimatedRemainingWorkPerPartitionAsync();
```
mathewc: nit: rename local to partitionWorkList

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+
+                if (workPerPartition != null)
```
mathewc: When would this be null?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+                }
+            }
+            catch (Microsoft.Azure.Documents.DocumentClientException e)
+            {
+                if (!TryHandleDocumentClientException(e))
+                {
+                    _logger.LogWarning("Unable to handle Microsoft.Azure.Documents.DocumentClientException: {0}", e.Message);
+                }
+            }
+            catch (NullReferenceException e)
```
mathewc: I don't think we should be swallowing null refs

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+            {
+                // This shouldn't be hit because we're not using ChangeFeedHost, and WorkEstimator throws DocumentClientException
+                // for missing collections.
+                _logger.LogWarning(e.Message);
+            }
+            catch (InvalidOperationException e)
```
mathewc: Not sure about all this error handling. Are you trying to identify transient errors? I don't like parsing the error strings looking for things (KnownDocumentClientErrors). Thinking we should be able to simplify and just log any exceptions we receive as Warnings + failure to get scale status. @ealsur may have an opinion as well

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+
+        ScaleStatus IScaleMonitor.GetScaleStatus(ScaleStatusContext context)
+        {
+            ScaleStatus status = new ScaleStatus
+            {
+                Vote = ScaleVote.None
+            };
+
+            const int NumberOfSamplesToConsider = 5;
+            List<CosmosDBTriggerMetrics> metrics = (context as ScaleStatusContext<CosmosDBTriggerMetrics>)?.Metrics?.ToList();
```
mathewc: Need to update this casting code - see my latest QueueListener for an example: https://github.com/Azure/azure-webjobs-sdk/blob/8a459a6cff394e4f1d0a5418e731024e7674256e/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs#L453

If you update your tests to test both the generic and non-generic interface methods, you'll find this code will fail as is

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -32,8 +37,24 @@ internal class CosmosDBTriggerListener : IListener, Microsoft.Azure.Documents.Ch
...
+        private static readonly Dictionary<string, string> KnownDocumentClientErrors = new Dictionary<string, string>()
+        {
+            { "Resource Not Found", "Please check that the CosmosDB collection and leases collection exist and are listed correctly in Functions config files." },
+            { "The input authorization token can't serve the request", string.Empty },
+            { "The MAC signature found in the HTTP request is not the same", string.Empty },
+            { "Service is currently unavailable.", string.Empty },
+            { "Entity with the specified id does not exist in the system.", string.Empty },
+            { "Subscription owning the database account is disabled.", string.Empty },
+            { "Request rate is large", string.Empty },
+            { "PartitionKey value must be supplied for this operation.", "We do not support lease collections with partitions at this time. Please create a new lease collection without partitions." },
```
ealsur:
```suggestion
            { "PartitionKey value must be supplied for this operation.", "Lease collections when partitioned, need to be partitioned by /id." },
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
ealsur: `DocumentClientException` would mean something happened at runtime trying to access the leases, for example, the user deleted the collections, or, we got throttled. `InvalidOperationException` would mean there is some issue with the input parameters, some parameter is null or missing.

I guess a nullref can happen if the user configuration is empty or it's pointing at a connection string that does not exist?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+            
+            // At least 5 samples are required to make a scale decision for the rest of the checks.
+            if (metrics.Count < NumberOfSamplesToConsider)
+            {
+                return status;
+            }
+
+            // Maintain a minimum ratio of 1 worker per 1,000 items of remaining work.
+            long latestRemainingWork = metrics.Last().RemainingWork;
+            if (latestRemainingWork > context.WorkerCount * 1000)
```
ealsur: Would 1000 be a good number? The Trigger can, by default (can be raised), process 100 items per execution, does Event Hub work similarly?

mathewc: I believe Graham has just ported the current scale controller heuristics here. We can tune this as needed in the future, but for now I think parity is fine

mathewc:
@ealsur The per trigger scale monitoring is only active if the Function App is put into this mode. It is only for Elastic Premium currently with a min reserved instance count of 1. There has to be at least one guaranteed instance running for this - the primary host instance is the one responsible for monitoring event sources, thus replacing the ScaleController monitoring. SC then is just responsible for hitting the new admin/host/scale/status endpoint to get scale votes, which each trigger can contribute to as you see here.

In the future we may open this up for other skus - but the requirement that at least one instance is guaranteed running remains.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +177,235 @@ private void InitializeBuilder()
...
+            long remainingWork = 0;
+            IReadOnlyList<RemainingPartitionWork> workPerPartition = null;
+
+            try
+            {
+                InitializeBuilder();
+
+                if (_workEstimator == null)
+                {
+                    _workEstimator = await this._hostBuilder.BuildEstimatorAsync();
```
ealsur: using the same builder to build both has not been giving great results, we believe some property gets cleaned after the build. Ideally two builders (they can have the same configuration)  are needed.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
gzuber: @mathewc those are valid points. These catch blocks, as well as the code related to KnownDocumentClientErrors, were all pulled from the existing scaling logic in ScaleController. So the code here is primarily my attempt to match the existing logging as closely as possible so we emit similar errors and troubleshooting info. I can definitely clean it up and do it the way you describe if that's what we would prefer.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
gzuber: I should have mentioned I had questions about how to replicate that queue scheme. I'm not sure what the equivalent of queue name would be for Cosmos. I've replaced `_queue.name` with `monitoredDb-monitoredCollection` if that works.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
gzuber: You're right, it shouldn't be. I think this was an artifact of copying from ScaleController

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
mathewc: Yes, that's it

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
mathewc: @ealsur might have been the one who wrote the original ScaleController code? If so and if he's OK with this then I guess it's fine. The whole idea of parsing these error messages is very fragile and if Cosmos starts returning different error messages, we'll silently start missing them.  Ideally the Cosmos errors would be structured and return an error ID we could key off of. Anyhow let's not block on this for the first rev.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +181,253 @@ private void InitializeBuilder()
...
+            IReadOnlyList<RemainingPartitionWork> partitionWorkList = null;
+
+            try
+            {
+                partitionWorkList = await workEstimator.GetEstimatedRemainingWorkPerPartitionAsync();
+
+                partitionCount = partitionWorkList.Count;
+                remainingWork = partitionWorkList.Sum(item => item.RemainingWork);
+            }
+            catch (Microsoft.Azure.Documents.DocumentClientException e)
```
mathewc: Recommend having a single block for the DocumentClientException and InvalidOperationException handling since the logic is repeated. E.g.

```csharp
catch (Exception ex) when (ex is DocumentClientException || ex is InvalidOperationException )
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +181,253 @@ private void InitializeBuilder()
...
+            {
+                string errormsg;
+
+                var webException = e.InnerException as WebException;
+                if (webException != null &&
+                    webException.Status == WebExceptionStatus.ProtocolError)
+                {
+                    string statusCode = ((HttpWebResponse)webException.Response).StatusCode.ToString();
+                    string statusDesc = ((HttpWebResponse)webException.Response).StatusDescription;
+                    errormsg = string.Format("CosmosDBtrigger status {0}: {1}.", statusCode, statusDesc);
```
mathewc: nit: CosmoseDBTrigger casing here and below :)

mathewc:
FYI - SDK has been published to myget, version Microsoft.Azure.WebJobs.3.0.13-11649

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
ealsur: I didn't write the ScaleController code, so I'm not very familiar with this logic, but I agree with your other comment regarding unifying them if they are similar.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -148,5 +181,253 @@ private void InitializeBuilder()
                     .WithObserverFactory(this);
             }
         }
+
+        private async Task<IRemainingWorkEstimator> GetWorkEstimatorAsync()
+        {
+            if (_workEstimatorBuilder == null)
+            {
+                _workEstimatorBuilder = new ChangeFeedProcessorBuilder()
```
ealsur: Just wondering, since I don't know the nature of the code that calls this, there won't be a change of it being called concurrently, right? Otherwise this might need a semaphore or lock

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
gzuber: After talking with @mathewc I think we decided to leave this error handling as it is for now (matching ScaleController), but I've removed the null ref block so we don't just swallow them.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
gzuber: removed

PR #590, opened 2019-09-27
Add Warmup (trigger) extension

ankitkumarr:
Related issue -- https://github.com/Azure/azure-functions-host/issues/4588
See related Host PR -- https://github.com/Azure/azure-functions-host/pull/4983

Adding a new Warmup extension that can be invoked through an admin API in the host. For more details, please look at the Host PR.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupContext.cs
+++ b/src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupContext.cs
@@ -1,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs
+{
+    public class WarmupContext
+    {
+    }
+}
--- a/src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerAttribute.cs
+++ b/src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerAttribute.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Description;
+
+namespace Microsoft.Azure.WebJobs
+{
+    [AttributeUsage(AttributeTargets.Parameter)]
+    [Binding]
+    public sealed class WarmupTriggerAttribute : Attribute
+    {
+    }
+}
--- a/src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerBinding.cs
+++ b/src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerBinding.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.Triggers;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Warmup.Trigger
+{
+    internal class WarmupTriggerBinding : ITriggerBinding
+    {
+        private readonly ParameterInfo _parameter;
+
+        public WarmupTriggerBinding(ParameterInfo parameter)
+        {
+            _parameter = parameter;
+        }
+
+        public Type TriggerValueType
+        {
+            get { return typeof(WarmupContext); }
+        }
+
+        public IReadOnlyDictionary<string, Type> BindingDataContract => null;
+
+        public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
+        {
+            return Task.FromResult<ITriggerData>(new TriggerData(null, null));
+        }
+
+        public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            return Task.FromResult<IListener>(new NullListener());
+        }
+
+        public ParameterDescriptor ToParameterDescriptor()
+        {
+            return new TriggerParameterDescriptor
+            {
+                Name = _parameter.Name
+            };
+        }
+
+        private class NullListener : IListener
+        {
+            /// <summary>
+            /// WarmupTrigger does not need to listen at all.
+            /// </summary>
+            public NullListener()
+            {
+            }
+
+            public Task StartAsync(CancellationToken cancellationToken)
+            {
+                return Task.CompletedTask;
+            }
+
+            public Task StopAsync(CancellationToken cancellationToken)
+            {
+                return Task.CompletedTask;
+            }
+
+            public void Cancel()
+            {
+            }
+
+            public void Dispose()
+            {
+            }
+        }
+    }
+}
--- a/src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
+++ b/src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Description;
+using Microsoft.Azure.WebJobs.Extensions.Warmup.Trigger;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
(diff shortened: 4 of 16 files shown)


test/WebJobs.Extensions.Warmup.Tests/WebJobs.Extensions.Warmup.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+  <ItemGroup>
+    <Folder Include="Properties\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\WebJobs.Extensions.Warmup\WebJobs.Extensions.Warmup.csproj" />
+    <ProjectReference Include="..\WebJobs.Extensions.Tests.Common\WebJobs.Extensions.Tests.Common.csproj" />
+  </ItemGroup>
+
+
```
ankitkumarr: Note to self: remove the extra line

ankitkumarr:
This is ready to review. I had forgotten to update the build scripts to include `warmup`. I just committed that.

src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Description;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Warmup.Trigger
```
mathewc: For intellisense discoverability, all binding attributes should be in the root Microsoft.Azure.WebJobs namespace. See other attributes.

src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Warmup
+{
+    [Extension("Warmup")]
+    public class WarmupConfigProvider : IExtensionConfigProvider
```
ankitkumarr: need to make this internal

src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerListener.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Warmup.Trigger
+{
+    internal class WarmupTriggerListener : IListener
```
mathewc: Rename to NullListener, and move to a private class in WarmupTriggerBinding to bury it

src/WebJobs.Extensions/Extensions/Warmup/WarmupWebJobsBuilderExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Warmup
```
mathewc: Builder extensions should be in Microsoft.Extensions.Hosting namespace for discoverability. See other extensions

src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var logger = _loggerFactory.CreateLogger(LogCategories.CreateTriggerCategory("Warmup"));
+            logger.LogInformation("Initializing Warmup Extension.");
+
+            context
+                .AddBindingRule<WarmupTriggerAttribute>()
+                .AddConverter<string, JObject>((str) => JObject.Parse(str))
+                .AddOpenConverter<string, OpenType.Poco>(typeof(JsonStringToPocoConverter<>))
```
mathewc: Perhaps we should define a Type WarmupContext as the trigger type. This way we define the contract for the data we'll pass through. For non .NET languages, we'll have the required converters (e.g. WarmupContext -> JObject, or string as needed).

Question is what data should WarmupContext contain?

src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerBinding.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            get { return _parameter.ParameterType; }
+        }
+
+        public IReadOnlyDictionary<string, Type> BindingDataContract => null;
+
+        public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
+        {
+            var bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+            Utility.ApplyBindingData(value, bindingData);
```
mathewc: Your BindingDataContract is null above (which makes sense for this trigger I think), so why are you applying binding data here? I think for this trigger we don't need any binding data/contract. 

Manual trigger is different, because users own the POCO being passed in, and may use it to define values that they want to bind to (part of binding data) in other input/output bindings.

src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
mathewc: Also, I don't think we need to support OpenType.Poco

src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerBinding.cs
ankitkumarr: You are right. I think I mixed this up. My thought was initially to not have a contract. Then I thought maybe users can bind to certain elements and not the entire trigger input (if we were to expand the context of this trigger). But that does not make sense in this context. And if we were to do something like that, I should have had a contract.

Good catch! I will update.

src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
ankitkumarr: At the moment, the data sent is a JSON with "hostState" in it.

Maybe the type `WarmupContext` could just have that. And in future if we need to expand it, we can.

src/WebJobs.Extensions/Utility.cs
```diff
@@ -23,5 +26,68 @@ public static bool IsDynamic
                 return string.Compare(value, DynamicSku, StringComparison.OrdinalIgnoreCase) == 0;
             }
         }
+
+        /// <summary>
+        /// Applies any additional binding data from the input value to the specified binding data.
+        /// This binding data then becomes available to the binding process (in the case of late bound bindings)
+        /// </summary>
+        internal static void ApplyBindingData(object value, Dictionary<string, object> bindingData)
```
mathewc: This code already lives in this project, as part of the http extension. However, since manual trigger won't have a binding contract, I think you won't need this.

test/WebJobs.Extensions.Tests/Extensions/Warmup/WarmupTriggerEndToEndTests.cs
```diff
@@ -0,0 +1,129 @@
...
+        private readonly TestLoggerProvider _loggerProvider = new TestLoggerProvider();
+
+        [Theory]
+        [InlineData("WarmupTriggerParams.TestWarmup_String")]
+        [InlineData("WarmupTriggerParams.TestWarmup_JObject")]
+        [InlineData("WarmupTriggerParams.TestWarmup_Poco")]
+        public async Task WarmupTriggerTest_Success(string functionName)
+        {
+            _functionOut = null;
+            var hostStatus = new
```
mathewc: I think we need to define the actual input contract here (e.g. WarmupContext) not based on a loose bag of properties.

src/WebJobs.Extensions/Extensions/Warmup/WarmupConfigProvider.cs
ankitkumarr: @mathewc, why not support Poco? It can't hurt to enable it from the extension, no?

This way if users do want to use a POCO, they are free to.

src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupTriggerBinding.cs
```diff
@@ -0,0 +1,57 @@
...
+        private readonly ParameterInfo _parameter;
+
+        public WarmupTriggerBinding(ParameterInfo parameter)
+        {
+            _parameter = parameter;
+        }
+
+        public Type TriggerValueType
+        {
+            get { return _parameter.ParameterType; }
```
mathewc: With the new Type, this should be return typeof(WarmupContext) - similar to TimerTrigger

src/WebJobs.Extensions/Extensions/Warmup/Trigger/WarmupContext.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs
```
ankitkumarr: guessing this needs to be here for intellisense too @mathewc

PR #594, opened 2019-10-07
changing EnableRewind to EnableBuffering

brettsam:
Missed this occurrence when I did #585.


brettsam:
I'm going to get this in so I can absorb it in v3. Explanation with some other links is in #585.

PR #597, opened 2019-10-11
changing synchronous stream.Write() call to async

brettsam:
This is causing issues due to AllowSynchronousIO now being false by default in ASP.NET Core 3.

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -463,8 +463,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                         {
                             using (var reader = new StreamReader(_request.Body))
                             {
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -463,8 +463,8 @@ public override async Task<object> GetValueAsync()
                         {
                             using (var reader = new StreamReader(_request.Body))
                             {
-                                var serializer = JsonSerializer.CreateDefault();
-                                return serializer.Deserialize(reader, typeof(object));
+                                string jsonString = await reader.ReadToEndAsync();
+                                return JsonConvert.DeserializeObject(jsonString);
```
brettsam: This ended up being the actual problem. If you were using `dynamic`, it'd throw in v3. Is this the best way to do this?

mathewc: Looks ok to me

PR #604, opened 2019-11-10
Support all FileSystemWatcher events

### LINKED ISSUE #250: FileTrigger does not support Renamed, Deleted WatcherChangeTypes
Please provide a succinct description of the issue.

#### Repro steps

Provide the steps required to reproduce the problem

1. Set FileTrigger to use WatcherChangeTypes.Renamed or WatcherChangeTypes.Deleted
2. View WebJob log stated only Changed or Created are supported.

#### Expected behavior

Renamed and Deleted WatcherChangeTypes should be supported, and work fine externally to FileTrigger.

#### Actual behavior

WebJob will not start and log states only Changed and Created are supported.

#### Known workarounds

Write custom FileSystemWatcher and not use SDK.

#### Related information 

Provide any related information 

* Package version
* Links to source:  https://github.com/Azure/azure-webjobs-sdk-extensions/blob/master/src/WebJobs.Extensions/Extensions/Files/Listener/FileListener.cs  Lines 171-178

robertmclaws:
Fixes #250.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Files/Listener/FileListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Files/Listener/FileListener.cs
@@ -157,13 +157,6 @@ namespace Microsoft.Azure.WebJobs.Files.Listeners
 
         private void CreateFileWatcher()
         {
-            if (_attribute.ChangeTypes != WatcherChangeTypes.Created &&
-                _attribute.ChangeTypes != WatcherChangeTypes.Changed &&
-                _attribute.ChangeTypes != (WatcherChangeTypes.Created | WatcherChangeTypes.Changed))
-            {
-                throw new NotSupportedException("Only the 'Created' and 'Changed' change types are supported.");
-            }
-
             if ((_attribute.ChangeTypes & WatcherChangeTypes.Changed) != 0 && _attribute.AutoDelete)
             {
                 throw new NotSupportedException("Use of AutoDelete is not supported when using change type 'Changed'.");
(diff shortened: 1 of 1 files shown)


robertmclaws:
FYI, all of the errors seem to come from the unit tests not having a valid Azure Storage connection string. None of the relevant tests failed.

paulbatum:
Thanks for the PR! I added the required label to get a full CI run for this. I think @mathewc should review this, he's out of the office right now but should be able to take a look next week.

robertmclaws:
Hey folks! Looks like the tests are passing now, thanks for that! Think this PR might be mergeable?

mathewc:
I'll pull the fork today and run the tests. I don't see any problems with getting this merged. My recollection (it was long ago when I created this extension) was that in App Service renames would trigger during file creates as well (e.g. when dragging/dropping a file into Kudu). How are you testing this out.

That said, since the watcher types are opt int, should be OK allowing customers to do what they want.

robertmclaws:
So we ran into a similar issue with Created events, and that's why we had to build support for the Renamed event. On a network transfer, the Created event will fire before the file is fully written, meaning the WebJob would pick up the file before it was finished, and would error out because the file would still be locked for writing.

We worked around that by writing to a different file extension, and then calling File.Move() in the same directory with the correct file extension after the write was finished. .NET simply runs a Rename operation in that case, which triggers a Rename event by the `FileSystemWatcher`. 

If you are able to accept this PR, then I will make a PR to the documentation showing how to make that scenario work properly.

mathewc:
@robertmclaws What version of the WebJobs SDK are you using? The reason I ask is because your PR is against our dev branch which is WebJobs v3 .NET Core, and in trying to port over the tests I mentioned above, it looks to me like this extension isn't in a good working state in the dev branch. I suspect you're using WebJobs v2? Our v2 branch is [here](https://github.com/Azure/azure-webjobs-sdk/tree/v2.x).

robertmclaws:
I'm using the v3 SDK, and confirmed that the problems I am having were with this branch. 

If the extension is not in a good state, or the tests are broken... let's fix them.

mathewc:
Ok - I fixed the extension, added back all the tests and got them passing again in https://github.com/Azure/azure-webjobs-sdk-extensions/pull/608.

I rebased your PR and merged that as well in https://github.com/Azure/azure-webjobs-sdk-extensions/pull/609.

Thanks for your contribution! Sorry it took so long for us to get this in.

robertmclaws:
No worries, thank you so much! :)

PR #616, opened 2020-01-23
Changed OnTimer async void

yojagad:


src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -174,9 +174,9 @@ public void Dispose()
             }
         }
 
-        private void OnTimer(object sender, ElapsedEventArgs e)
+        private async void OnTimer(object sender, ElapsedEventArgs e)
         {
-            HandleTimerEvent().Wait();
+            await HandleTimerEvent();
```
brettsam: @mathewc  / @fabiocav -- do you see any potential side effects of doing this? We had a customer point out the `.Wait()` and @yojagad and I both said "sure, we can fix that". 

Now I'm curious if there's any subtle side effects (like exception handling?)

yojagad:
Closing this for now. Will reopen if/when needed

PR #602, opened 2019-11-01
Pull through Host.Storage and Extensions.Storage updates

paulbatum:
- Updates to storage SDK v11
- Includes a major version rev for WebJobs.Extensions

This would go in after https://github.com/Azure/azure-webjobs-sdk/pull/2356

PR #630, opened 2020-02-20
Update to storage SDK 11.x

brettsam:
- Pull through Host.Storage 4.x and Extensions.Storage 4.x
- Bumped WebJobs.Extensions to 4.x

Replacement for #602

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!-- Extensions can have independent versions and only increment when released -->
     <Version>3.0.0$(VersionSuffix)</Version>
-    <ExtensionsVersion>3.0.6$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
+    <ExtensionsVersion>4.0.0$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
     <CosmosDBVersion>3.0.5$(VersionSuffix)</CosmosDBVersion>
     <HttpVersion>3.0.6$(VersionSuffix)</HttpVersion>
     <MobileAppsVersion>3.0.0$(VersionSuffix)</MobileAppsVersion>
--- a/src/ExtensionsSample/ExtensionsSample.csproj
+++ b/src/ExtensionsSample/ExtensionsSample.csproj
@@ -21,7 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="3.0.7" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="4.0.0-preview1" />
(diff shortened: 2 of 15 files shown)


whitwa:
I was just wondering when an updated nuget package for this will become available? I am having to use the source code at the moment.

brettsam:
I commented elsewhere but will link here as well -- it's been released for several months: https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.Storage/4.0.2.

whitwa:
> I commented elsewhere but will link here as well -- it's been released for several months: https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.Storage/4.0.2.

Thanks. But I already have version 4.0.2 of Microsoft.Azure.WebJobs.Extensions.Storage installed. I need Microsoft.Azure.WebJobs.Extensions (latest available version from NuGet is 3.0.6) for TimerTrigger support in my WebJob?

brettsam:
Oh -- you don't actually need to reference that version directly. We automatically include it for you when the host runs. If you do something like this:

```
[FunctionName("Function2")]
public static void Run([TimerTrigger("0 */5 * * * *", RunOnStartup = true)] TimerInfo myTimer, ILogger log)
{
    log.LogInformation(typeof(TimerTriggerAttribute).Assembly.FullName);
}
```

You'll see it log:
```
Microsoft.Azure.WebJobs.Extensions, Version=4.0.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35
```

whitwa:
Thanks. If I remove the Microsoft.Azure.WebJobs.Extensions reference from the project, it no longer builds however.

error CS0246: The type or namespace name 'TimerTrigger' could not be found (are you missing a using directive or an assembly reference?)

brettsam:
Sorry -- I wasn't clear. You do need a reference to *some version* of WebJobs.Extensions so you can have the attribute. But no matter what, you'll run on the latest build when you hit F5 locally or when you deploy to Azure. This is because the Functions host process includes the latest version by default and your trigger will run against that version.

whitwa:
Thanks and no problem. If I install the latest Microsoft.Azure.WebJobs.Extensions (3.0.6) from NuGet instead of using my own source code then I get the following exception when HostBuilder.Run is executed:

The listener for function 'Functions.ProcessTimerMessage' was unable to start.

Inner exception: Could not load file or assembly 'Microsoft.WindowsAzure.Storage, Version=9.3.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.

I have no reference to Microsoft.WindowsAzure.Storage anywhere in my code. It's like the latest version of Microsoft.Azure.WebJobs.Extensions is not being utilized as you are expecting?

brettsam:
Would you mind creating a new issue here (since this is a PR we're commenting on) and mentioning me when you do it? If you can include your .csproj file and your function .cs file, that'd be very helpful.

PR #635, opened 2020-04-19
Throughput improvements

mathewc:
For Functions work item: https://github.com/Azure/azure-functions-host/issues/5601

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -95,9 +95,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
             private readonly bool _isUserTypeBinding;
             private readonly Dictionary<string, Type> _bindingDataContract;
             private readonly Action<HttpRequest, object> _responseHook;
-            private readonly bool _isEasyAuthEnabled;
 
-            public HttpTriggerBinding(HttpTriggerAttribute attribute, ParameterInfo parameter, bool isUserTypeBinding, Action<HttpRequest, object> responseHook = null, bool isEasyAuthEnabled = false)
+            public HttpTriggerBinding(HttpTriggerAttribute attribute, ParameterInfo parameter, bool isUserTypeBinding, Action<HttpRequest, object> responseHook = null)
             {
                 _responseHook = responseHook;
                 _parameter = parameter;
@@ -113,7 +112,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                 }
 
                 _bindingDataContract = GetBindingDataContract(attribute, parameter);
-                _isEasyAuthEnabled = isEasyAuthEnabled;
             }
 
             public IReadOnlyDictionary<string, Type> BindingDataContract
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -138,26 +138,43 @@ public async Task<ITriggerData> BindAsync(object value, ValueBindingContext cont
...
                 else
                 {
                     valueProvider = new HttpRequestValueBinder(_parameter, request, invokeString);
                 }
 
+                // collection of binding data is expensive, so we define a factory to allow this
+                // to happen lazily only when needed
+                Func<Task<IReadOnlyDictionary<string, object>>> bindingDataFactory = () =>
+                {
+                    return GetBindingDataAsync(request, valueProvider);
```
mathewc, line 155: Making binding data resolution lazy saves a bunch of work, since for http it involves attempted json deserialization of the requet body and a bunch of other operations, that usually aren't needed.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -290,8 +299,11 @@ internal static void ApplyBindingData(object target, IDictionary<string, object>
                 var bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
                 if (request.ContentLength != null && request.ContentLength > 0)
                 {
-                    string body = await request.ReadAsStringAsync();
-                    Utility.ApplyBindingData(body, bindingData);
+                    if (string.IsNullOrEmpty(request.ContentType) || string.Equals(request.ContentType, "application/json", StringComparison.OrdinalIgnoreCase))
```
mathewc, line 302: This is the main change. Avoid the body read + json deserialization when we can. E.g. if the content type is specified as text/html, we shouldn't be trying to json deserialize it.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -284,11 +291,12 @@ internal static void ApplyBindingData(object target, IDictionary<string, object>
                 return aggregateDataContract;
             }
 
-            internal static async Task<IReadOnlyDictionary<string, object>> GetRequestBindingDataAsync(HttpRequest request, Dictionary<string, Type> bindingDataContract = null, bool isEasyAuthEnabled = false)
+            internal static async Task<IReadOnlyDictionary<string, object>> GetRequestBindingDataAsync(HttpRequest request, Dictionary<string, Type> bindingDataContract = null)
             {
                 // apply binding data from request body if present
                 var bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
-                if (request.ContentLength != null && request.ContentLength > 0)
+                if (request.ContentLength != null && request.ContentLength > 0 &&
```
mathewc, line 298: If a content type is specified and it's not json, don't try to json deserialize it. That avoids performance issues, e.g. if large images or other data are sent with content type - we shouldn't be trying to json deserialize them. For back compat, if no content type specified, we'll still try.

pragnagopa (approved):
👍

PR #636, opened 2020-04-21
Desearilizing collections in POCO

alrod:
Fixes: https://github.com/Azure/azure-functions-host/issues/3370

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -403,11 +403,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                 {
                     var underlyingTargetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
 
-                    var jToken = value as JToken;
-                    if (jToken != null)
+                    if (value is JObject jObject)
+                    {
+                        value = jObject.ToObject(targetType);
+                    }
+                    else if (value is JArray jArray)
                     {
                         // value is JObject or JArray
-                        value = jToken.ToObject(targetType);
+                        value = jArray.ToObject(targetType);
                     }
                     else if (underlyingTargetType == typeof(Guid) && value.GetType() == typeof(string))
                     {
--- a/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
@@ -403,10 +403,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                 {
                     var underlyingTargetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
 
-                    var jObject = value as JObject;
-                    if (jObject != null)
+                    var jToken = value as JToken;
+                    if (jToken != null)
                     {
-                        value = jObject.ToObject(targetType);
+                        // value is JObject or JArray
+                        value = jToken.ToObject(targetType);
                     }
                     else if (underlyingTargetType == typeof(Guid) && value.GetType() == typeof(string))
                     {
--- a/src/WebJobs.Extensions.Http/Utility.cs
+++ b/src/WebJobs.Extensions.Http/Utility.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
                     // parse the object adding top level properties
                     JObject parsed = JObject.Parse(json);
                     var additionalBindingData = parsed.Children<JProperty>()
-                        .Where(p => p.Value != null && (p.Value.Type != JTokenType.Array))
+                        .Where(p => p.Value != null)
                         .ToDictionary(p => p.Name, p => ConvertPropertyValue(p));
 
                     if (additionalBindingData != null)
@@ -54,6 +54,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
             if (property.Value != null && property.Value.Type == JTokenType.Object)
             {
                 return (JObject)property.Value;
+            }
+            else if (property.Value != null && property.Value.Type == JTokenType.Array)
+            {
+                return (JArray)property.Value;
             }
             else
             {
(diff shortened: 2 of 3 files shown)


test/WebJobs.Extensions.Http.Tests/HttpTriggerBindingTests.cs
```diff
@@ -169,7 +173,7 @@ public async Task BindAsync_Poco_FromRequestBody()
             JObject requestBody = new JObject
             {
                 { "Name", "Mathew Charles" },
-                { "Location", "Seattle" }
+                { "Location", "Seattle" },
```
fabiocav, line 176: ?

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
```diff
@@ -403,10 +403,11 @@ private static object ConvertValueIfNecessary(object value, Type targetType)
                 {
                     var underlyingTargetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
 
-                    var jObject = value as JObject;
-                    if (jObject != null)
+                    var jToken = value as JToken;
+                    if (jToken != null)
                     {
-                        value = jObject.ToObject(targetType);
+                        // value is JObject or JArray
```
mathewc, line 409: It's a JToken - how do you know its an array or object? It could be a string, or scalar value right?

src/WebJobs.Extensions.Http/Utility.cs
```diff
@@ -26,7 +26,7 @@ internal static void ApplyBindingData(object value, Dictionary<string, object> b
                     // parse the object adding top level properties
                     JObject parsed = JObject.Parse(json);
                     var additionalBindingData = parsed.Children<JProperty>()
-                        .Where(p => p.Value != null && (p.Value.Type != JTokenType.Array))
+                        .Where(p => p.Value != null)
```
mathewc, line 29: The code comment above indicates that we're intentionally only pulling binding data from top level properties. We don't want to add JArray properties like Length, etc. to binding data

mathewc, line 29: Note that this same method also exists as a copy in the Functions host.  If we make changes, we need to keep in sync. This code is pretty explicit that the intent is NOT to add arrrays to binding data, and it's been that way since the beginning I believe (check history).

mathewc:
The issue in the original bug seems to be that on the bound poco instance, the child arrays are null. It doesn't look like these changes address that at all - you're just modifying binding data?

fabiocav:
The comment above was my observation as well, were you able to validate the original issue after these changes?

src/WebJobs.Extensions.Http/Utility.cs
alrod: Do you remember why we do not want to add collections to binding data? We already add objects. 
Commit:
https://github.com/Azure/azure-functions-host/commit/4d112b27512a507028ded133bc9fde5a4cf259d2#diff-a63b998bf2cabc5184041834fd8678e1

alrod:
Binding data is copied to poco properties:
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L175

alrod:
> Changes look good, but I'm worried about changes in behavior impacting existing workloads relying on how this works today. Thoughts? Did you see any impact to end-to-end test runs and other validation?

I tested it e2e in acript host referencing new Http Extension. I would not check in the change until all tests are passed in host using new HTTP extension. I am not sure about impacting workloads.

alrod:
> The issue in the original bug seems to be that on the bound poco instance, the child arrays are null. It doesn't look like these changes address that at all - you're just modifying binding data?

Binding data is copied to poco properties:
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L175

fabiocav:
@alrod I wouldn't expect this code path to be the target of the changes in 3.x (which is what this fix should be targeting). In 3.x, all the .NET response should typically be handled in the `HttpBinding` initial SetResponse evaluation.

I expected changes to the serializer configuration to be sufficient here.

test/WebJobs.Extensions.Http.Tests/HttpTriggerBindingTests.cs
alrod: Fixed

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
alrod: Fixed

mathewc:
While dumping all the values into the BindingData bag might work, I don't think it's the right solution. Instead, we should just figure out why the POCO hydration code skipping these properties. That code is here: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/8f45bd7f0f59a1c29f4d15e25c96fdbfe6ec1b27/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L146. Maybe it's just an issue with serialization settings? Maybe the type needs to have JProperty attributes on the properties or something?

I don't think we've ever added these collection properties to binding data. How did this work in v2 or v1? Likely we didn't there as well.

src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs
mathewc: comment is still wrong

mathewc: This still doesn't seem right. Here you know its not a JObject, but a JArray, right?

fabiocav:
> While dumping all the values into the BindingData bag might work, I don't think it's the right solution. Instead, we should just figure out why the POCO hydration code skipping these properties. That code is here:
> 
> https://github.com/Azure/azure-webjobs-sdk-extensions/blob/8f45bd7f0f59a1c29f4d15e25c96fdbfe6ec1b27/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L146
> 
> . Maybe it's just an issue with serialization settings? Maybe the type needs to have JProperty attributes on the properties or something?
> I don't think we've ever added these collection properties to binding data. How did this work in v2 or v1? Likely we didn't there as well.

This is essentially what I meant above. The right fix for this issue would be an update to the serialization behavior/configuration. Also, this is still targeting `dev`, when the fix should likely only go into 3.0

fabiocav:
Following up as the comment about the target isn't clear. This would likely be a change in the runtime serialization behavior. If done in the extension (which is an option as well), you'd need to make sure this is opt-in

alrod:
Closing this as the issue is fixed in V3.

PR #648, opened 2020-06-05
updating Storage nugets; bumping extensions version

brettsam:


brettsam:
FYI -- @mathewc -- I'm gonna merge this myself since it's so minor

I was going to kick off an official Http extension build for our next release and decided I should update the nugets here as well. It's nothing major, but 4.0.1 of Host.Extensions cleans up some dependencies and updates to the latest storage (which has some fixes for us).

Issue #600, opened 2019-10-21
CosmosDBTrigger use the UseDefaultJsonSerialization parameter not correct


schw3de:
#### Repro steps

Provide the steps required to reproduce the problem

Use the CosmosDBTriggerSamples in the project.

1. Step A

Set the UseDefaultJsonSerialization to true.

```
public static void Listen(
[CosmosDBTrigger("ItemDb", "ItemCollection", LeaseCollectionName = "Leases", UseDefaultJsonSerialization = true)] 
```

2. Step B

Run the sample.

#### Expected behavior

Exception should occur since no JsonConvert.DefaultSettings was set.
If you have declared the default settings those should take into account.

#### Actual behavior

The UseDefaultJsonSerialization set the parameter for useMultipleWriteLocations instead.

#### Known workarounds

Here is a commit for a possible bugfix:
https://github.com/schw3de/azure-webjobs-sdk-extensions/commit/35bb7dde63908017f18e835b51baa98b6551d654

#### Related information 

Known in:
Microsoft.Azure.WebJobs.Extensions.CosmosDB  Version=3.0.4

schw3de:
If you want I can open a Pull Request from the fork. Just let me know.

ealsur:
This was fixed on the linked PR and shipped in 3.0.8

PR #575, opened 2019-07-10
Add Default JsonSerializerSettings to DocumentClient to fix DateTimeOffset parsing

JohnLTaylor:
[See bug 574](https://github.com/Azure/azure-webjobs-sdk-extensions/issues/574)

DateTimeOffset are having their Offset changed from the value stored to server local time

PR #651, opened 2020-06-29
Fix for CosmosDbTrigger to honor UseDefaultJsonSerialization setting

james-world:
As reported in issue #600, the UseDefaultJsonSerialization setting is not correctly passed by CosmosDbTriggerAttributeBindingProvider - it incorrectly passes the setting UseMultipleWriteLocations instead, which is irrelevant on the monitoring collection as this is only read from.

This means that using custom json serialization settings is not currently possible.

Fix is very small, unit test included which checks that the JsonConverter.DefaultSettings is actually accessed. This is a bit sneaky, but does accurately check the desired outcome and avoids having to change the code to be able to inspect the passed setting.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
@@ -147,7 +147,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
                     throw new InvalidOperationException("The monitored collection cannot be the same as the collection storing the leases.");
                 }
 
-                monitoredCosmosDBService = _configProvider.GetService(triggerConnectionString, resolvedPreferredLocations, attribute.UseDefaultJsonSerialization);
+                monitoredCosmosDBService = _configProvider.GetService(triggerConnectionString, resolvedPreferredLocations, false, attribute.UseDefaultJsonSerialization);
                 leaseCosmosDBService = _configProvider.GetService(leasesConnectionString, resolvedPreferredLocations, attribute.UseMultipleWriteLocations);
 
                 if (attribute.CreateLeaseCollectionIfNotExists)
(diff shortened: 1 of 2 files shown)


james-world:
I don't think these build failures are related to this PR. It certainly builds and tests OK locally.

ealsur (approved):

ealsur:
@brettsam Could you kickoff the CI tests again? See if this is transient?

brettsam:
Someone internally had uploaded the wrong version of a dependency to our MyGet feed. And since nuget uses a "first one found wins" approach, the CI was grabbing the wrong package with the wrong dependency. I've hidden the myget packages since the one on nuget is the real one -- and now it looks like it's building.

ealsur:
@brettsam @paulbatum Could this be merged?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -147,7 +147,7 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
                     throw new InvalidOperationException("The monitored collection cannot be the same as the collection storing the leases.");
                 }
 
-                monitoredCosmosDBService = _configProvider.GetService(triggerConnectionString, resolvedPreferredLocations, attribute.UseDefaultJsonSerialization);
+                monitoredCosmosDBService = _configProvider.GetService(triggerConnectionString, resolvedPreferredLocations, false, attribute.UseDefaultJsonSerialization);
```
brettsam, line 150: What does this `false` do?

ealsur: This is a bug introduced on https://github.com/Azure/azure-webjobs-sdk-extensions/pull/575, basically `GetService` had originally 3 parameters, that PR added a 4th, but put it as the 3th.

image

The false is the default value for `UseMultipleWriteLocations`.

brettsam: Ah, so we were inadvertently controlling `UseMultipleWriteLocations` with `UseDefaultJsonSerialization`? Got it, this makes sense now :-)

brettsam (approved):

PR #654, opened 2020-07-15
removing check for application/json before calling EnableBuffering()

### LINKED ISSUE #655: HttpTrigger issues in production 3.0.14111 and 2.0.14048
Exception is -- 
```
System.NotSupportedException : Specified method is not supported.
   at Microsoft.AspNetCore.Server.IIS.Core.ReadOnlyStream.set_Position(Int64 value)
   at Microsoft.AspNetCore.Server.IIS.Core.WrappingStream.set_Position(Int64 value)
```

The issue is with any requests that have content and the content-type header is not explicitly application/json
 
v2:
- Cannot seek the request body in a C# app
- Out-of-proc languages fail.
v3:
- Cannot seek the request body in a C# app

brettsam:
fixes #655

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -4,7 +4,7 @@
     <Version>3.0.0$(VersionSuffix)</Version>
     <ExtensionsVersion>4.0.1$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
     <CosmosDBVersion>3.0.8$(VersionSuffix)</CosmosDBVersion>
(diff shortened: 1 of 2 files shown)


brettsam:
@mathewc as FYI for future -- we likely need to come back and re-evaluate this.

mathewc:
Thanks. Looks like we just need to ensure EnableBuffering gets called in all cases, then we'll be able to add back this optimization. We really shouldn't be json parsing bodies that explicitly declare a non json payload.

PR #663, opened 2020-09-17
Adding support for static web app client principal

mkarmark:


src/WebJobs.Extensions.Http/WebJobs.Extensions.Http.csproj
```diff
@@ -1,12 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
-    <Version>$(HttpVersion)</Version>
+    <Version>3.0.8.1</Version>
```
mkarmark: @ConnorMcMahon I know you mentioned the need to update the assembly version but not sure if this is what you meant.

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -63,25 +64,36 @@ public static ClaimsIdentity GetAppServiceIdentity(this HttpRequest request)
                 return null;
             }
             string headerValue = request.Headers[EasyAuthIdentityHeader].First();
-            return FromBase64EncodedJson(headerValue);
+            return FromBase64EncodedJson<ClaimsIdentitySlim>(headerValue, EasyAuthClaimsIdentitySerializer.Value);
         }
 
-        private static ClaimsIdentity FromBase64EncodedJson(string payload)
+        public static ClaimsIdentity GetStaticWebAppsIdentity(this HttpRequest request)
```
cgillum: I don't see any callers of this method. Is that because it will be called by another project?

src/WebJobs.Extensions.Http/WebJobs.Extensions.Http.csproj
ConnorMcMahon: Ah, you actually don't need to update the assembly version. You just need to update the nuget version everytime you test locally to avoid hitting the cached nuget version.

In general, you will want to update the Version before this releases, but I think you will want to edit the $(HttpVersion) variable. I think that comes from the common.props file?

cgillum (commented):
Adding @fabiocav to review the csproj changes since I can't speak to whether those are okay to make.

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
ConnorMcMahon: In general, these are called by the AuthenticationHandler in the Functions WebHost.

mkarmark: Yup that's where I call this method!

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -63,25 +64,36 @@ public static ClaimsIdentity GetAppServiceIdentity(this HttpRequest request)
...
+            if (!request.Headers.ContainsKey(EasyAuthIdentityHeader))
+            {
+                return null;
+            }
+            string headerValue = request.Headers[EasyAuthIdentityHeader].First();
+            return FromBase64EncodedJson<StaticWebAppsClientPrincipal>(headerValue, StaticWebAppsClaimsIdentitySerializer.Value);
+        }
+
+        private static ClaimsIdentity FromBase64EncodedJson<T>(string payload, DataContractJsonSerializer serializer)
+            where T : IIdentityPrincipal
```
ConnorMcMahon, line 82: The interface isn't technically needed, if this method instead returns the object of type `T` immediately, and then the callers of this method now have a `ClaimsIdentitySlim` or `StaticWebAppsClientPrincipal` that they then just call whatever methods on, instead of needing to code against the interface. I wouldn't be anti-interface necessarily, but I do think it's a bit of unnecessary boilerplate that I don't see us needing the extensibility down the line.

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
ConnorMcMahon: So it looks like both this and the original `GetAppServiceIdentity()` throw an exception if the "x-ms-client-principal" header doesn't match the expected format. Since the format can now feasibly be either, that means that catching the exception will be forced to be part of the normal control flow, which is not ideal. 

My recommendation would be to do one of the following:

1. Make these `TryGet*` method so that it is more apparent that this may not successfully return the expected ClaimsIdentity.
2. Return a `null` value. This generally can lead to more aesthetically pleasing code (i.e. `var headerIdentity = httpContext.GetAppServiceIdentity() ?? httpContext.GetStaticWebAppsIdentity()`), but if you go with this approach, you have to properly document that the methods can return null, and still be aware that some consumers of the method will not read the method docs and won't properly handle the null case.

I would probably recommend solution 1.

ConnorMcMahon (commented):
In general, the logic to generate the claims identity seems sound. It is a bit hard to review in isolation, considering the consumer of these methods is in another repository.

ConnorMcMahon (commented):
Looks good so far.

I'm going to review the other PR before giving final approval though.

ConnorMcMahon (approved):
LGTM!

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -56,32 +57,59 @@ public static bool IsJsonContentType(this HttpRequest request)
...
+                return false;
+            }
+
+            claimsIdentity = claimsIdentitySlim.ToClaimsIdentity();
+            return true;
+        }
+
+        public static bool TryGetStaticWebAppsIdentity(this HttpRequest request, out ClaimsIdentity claimsIdentity)
+        {
+            claimsIdentity = null;
```
fabiocav: This method duplicates the logic above. Can we consolidate the logic in a single method instead? We can expose a new argument (enum) so the consumer can indicate the expected identity type which would drive the logic for which serializer to use and the comparisons.

Your `FromBase64EncodedJson` can also revert to the previous behavior where it returns the claims identity. That would require you to add a constraint and introduce an interface that is implemented by the principal/claims types, defining the method they'd need to implement (`ToClaimsIdentity`). This would allow you to simplify this and avoid duplication.

fabiocav (changes requested):
Implementation suggestion. Let me know what you think.

src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs
```diff
@@ -56,32 +57,59 @@ public static bool IsJsonContentType(this HttpRequest request)
                 string.Equals(headerValue.MediaType, "application/json", StringComparison.OrdinalIgnoreCase);
         }
 
-        public static ClaimsIdentity GetAppServiceIdentity(this HttpRequest request)
+        public static bool TryGetAuthIdentity(this HttpRequest request, AuthIdentityEnum authIdentity, out ClaimsIdentity claimsIdentity)
```
mathewc: This is a breaking change to a public API - is this a necessary change? Who is calling this new API -  I don't see it in your host PR.

fabiocav: Indeed. We can keep the existing API and call into this with the original behavior.

ConnorMcMahon (approved):
LGTM!

Just make sure we validate this end-to-end with the Functions host.

mkarmark:
> LGTM!
> 
> Just make sure we validate this end-to-end with the Functions host.

E2E test added here and it passes w the updated extensions nuget: https://github.com/Azure/azure-functions-host/pull/7256/files

src/WebJobs.Extensions.Http/StaticWebAppsClientPrincipal.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        [DataMember(Name = "userDetails")]
+        internal string UserDetails;
+
+        [DataMember(Name = "userRoles")]
+        internal List<string> UserRoles;
+
+        public ClaimsIdentity ToClaimsIdentity()
+        {
+            var staticWebAppsIdentity = new ClaimsIdentity(this.IdentityProvider);
```
fabiocav: small nit: by convention, in this codebase, we don't use `this.` unless required.

fabiocav (approved):
Small comment, but looks good!

PR #668, opened 2020-10-15
Improve functions startup time when warmup triggers are provided.

safihamid:
This is the WebJobs.Extensions.Http part.
This combined with the PR for WebJobs.Script will move the
/admin/warmup call completely within the functions execution pipeline rather than being a separate controller
and will avoid all the JIT cost that incur during the first real function execution even when warmup triggers are present.

### DIFF
--- a/src/WebJobs.Extensions.Http/Routing/IWebJobsRouter.cs
+++ b/src/WebJobs.Extensions.Http/Routing/IWebJobsRouter.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
     {
         IInlineConstraintResolver ConstraintResolver { get; }
 
-        void AddFunctionRoutes(IRouter functionRoutes, IRouter proxyRoutes);
+        void AddFunctionRoutes(IRouter functionRoutes, IRouter proxyRoutes, IRouter warmupRoute);
 
         void ClearRoutes();
 
--- a/src/WebJobs.Extensions.Http/Routing/IWebJobsRouter.cs
+++ b/src/WebJobs.Extensions.Http/Routing/IWebJobsRouter.cs
@@ -9,7 +9,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
     {
         IInlineConstraintResolver ConstraintResolver { get; }
 
-        void AddFunctionRoutes(IRouter functionRoutes, IRouter proxyRoutes, IRouter warmupRoute);
+        void AddFunctionRoutes(IRouter functionRoutes, IRouter proxyRoutes);
+
+        void AddCustomRoutes(IRouter customRoutes);
 
         void ClearRoutes();
 
--- a/src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
+++ b/src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
@@ -15,7 +15,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
         private readonly IInlineConstraintResolver _constraintResolver;
         private RouteCollection _functionRoutes;
         private RouteCollection _proxyRoutes;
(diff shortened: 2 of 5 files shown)


fabiocav (changes requested):
Can we come up with a design that doesn't drive knowledge of the warmup route into the routing APIs? This is something we can do with a custom `IRouter` that just exposes the right information and/or behavior

src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
```diff
@@ -51,12 +51,6 @@ private void InitializeRouteCollections()
 
         public Task RouteAsync(RouteContext context)
         {
-            // /admin/* routes should not be allowed to be overriden by proxies.
-            if (context.HttpContext.Request.Path.StartsWithSegments(new PathString("/admin"), System.StringComparison.OrdinalIgnoreCase))
```
safihamid, line 55: This is really not needed as it brings the knowledge from ScriptHost into routing APIs and we are already doing a check in ScriptHost for this.
The check is here: [https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/ScriptHost.cs#L765](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/ScriptHost.cs#L765)

safihamid:
sure, updated the PR with this change.

---
In reply to: [509866214](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/668#issuecomment-509866214) [](ancestors = 509866214)

src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
```diff
@@ -79,19 +73,27 @@ public void AddFunctionRoutes(IRouter functionRoutes, IRouter proxyRoutes)
...
+
         public string GetFunctionRouteTemplate(string functionName)
         {
             if (functionName == null)
             {
                 throw new ArgumentNullException(nameof(functionName));
             }
 
-            if (_functionRoutes == null)
+            if (_routeCollection == null)
```
fabiocav, line 91: Is this change just to ensure we're also processing the new/custom routes? The impact is that proxy routes will also be evaluated

fabiocav (approved):
Small comments

src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
```diff
@@ -51,12 +51,6 @@ private void InitializeRouteCollections()
 
         public Task RouteAsync(RouteContext context)
         {
-            // /admin/* routes should not be allowed to be overriden by proxies.
```
safihamid, line 54: Yes it should not get hit. If we left it here then I needed to add more logic for admin/warmup here because that's an exception. we can still add this logic in host repo under InvokeAsync in ScriptRouteHandler

src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
safihamid: This method was just added for tests and I needed to update it for new custom route tests

---
In reply to: [511065642](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/668#discussion_r511065642) [](ancestors = 511065642)

src/WebJobs.Extensions.Http/Routing/IWebJobsRouter.cs
```diff
@@ -11,6 +11,8 @@ public interface IWebJobsRouter : IRouter
 
         void AddFunctionRoutes(IRouter functionRoutes, IRouter proxyRoutes);
 
+        void AddCustomRoutes(IRouter customRoutes);
```
mathewc, line 14: @fabiocav This is a breaking changes to a public interface. Is there a way for us to make this change w/o breaking this interface?

PR #670, opened 2020-10-19
Remove Error Trigger and Mobile Apps from readme

paulbatum:
The error trigger section had a broken link that was reported, and anyone landing on it from functions land is going to be confused anyway. Also since Mobile Apps is pretty much gone I removed that section too.

Same PR as https://github.com/Azure/azure-webjobs-sdk-extensions/pull/669 but for master


paulbatum:
I screwed this one up, closing.

Issue #674, opened 2020-10-22
Specific CRON Schedule Results Race Condition on Schedule Monitor


wenhzha:
For timer trigger, we initialize schedule monitor to resume any missed/interrupted timer triggered execution. The original desing is to enable this when the timer is less frequent than every minute. [The code we implemented this logic](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/a2b2eb64e21225dac9d6ec0f284d953600a0cdba/src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs#L70) is too simple, which does not work for specific CRON schedule and cause race condition when the host decides to enable schedule monitor or not. 

#### Repro steps
1. Have a timer trigger using CRON: * */20 * * * *, the function will execute every second for a minute when the minute value is divisible by 20. For example, it will execute every second from 00:00 to 00:01, 00:20 to 00:21, 00:40 to 00:41, 01:00 to 01:01, etc..
2. If function host wakes up around 11:55, in the next a minute, there is no expected occurrence, the host would create schedule monitor and maintain the status blob in host container. 
3. If function host wakes up around 11:59:30, in the next a minute, there will be 30 occurrences, the host would not create schedule monitor.

This cron pattern might not be common and the function execution would not really impacted by having a scheudle monitor or not. But as we start to let scale controller to depend on status blob for site activation and keep workers ping, it becomes critical for scale controller and host agree if a schedule monitor is created or not.

PR #679, opened 2020-10-30
Fix for schedule monitor enablement (#674)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk-extensions/issues/674. Will also make this fix to v2.x.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
@@ -16,9 +16,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
         private readonly CrontabSchedule _cronSchedule;
 
         /// <summary>
-        /// Constructs a new instance based on the specified crontab schedule
+        /// Constructs a new instance based on the specified crontab schedule.
         /// </summary>
-        /// <param name="schedule">The crontab schedule to use</param>
+        /// <param name="schedule">The crontab schedule to use.</param>
         public CronSchedule(CrontabSchedule schedule)
         {
             _cronSchedule = schedule;
--- a/src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Azure.WebJobs.Host;
+using Microsoft.Extensions.Logging;
+using NCrontab;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Timers
 {
@@ -56,7 +58,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
             return occurrences;
         }
 
-        internal static TimerSchedule Create(TimerTriggerAttribute attribute, INameResolver nameResolver)
+        internal static TimerSchedule Create(TimerTriggerAttribute attribute, INameResolver nameResolver, ILogger logger)
         {
             TimerSchedule schedule = null;
 
@@ -66,13 +68,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
                 if (CronSchedule.TryCreate(resolvedExpression, out CronSchedule cronSchedule))
                 {
                     schedule = cronSchedule;
-
-                    DateTime[] nextOccurrences = cronSchedule.InnerSchedule.GetNextOccurrences(DateTime.Now, DateTime.Now + TimeSpan.FromMinutes(1)).ToArray();
-                    if (nextOccurrences.Length > 1)
+                    if (attribute.UseMonitor && ShouldDisableScheduleMonitor(cronSchedule, DateTime.Now))
                     {
-                        // if there is more than one occurrence due in the next minute,
-                        // assume that this is a sub-minute constant schedule and disable
-                        // persistence
+                        logger.LogDebug("UseMonitor changed to false based on schedule frequency.");
                         attribute.UseMonitor = false;
                     }
                 }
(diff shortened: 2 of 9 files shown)


src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -16,9 +16,9 @@ public class CronSchedule : TimerSchedule
         private readonly CrontabSchedule _cronSchedule;
 
         /// <summary>
-        /// Constructs a new instance based on the specified crontab schedule
+        /// Constructs a new instance based on the specified crontab schedule.
```
mathewc, line 19: Just fixing fxcop issues in a few of these files

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
```diff
@@ -98,5 +94,41 @@ internal static TimerSchedule Create(TimerTriggerAttribute attribute, INameResol
...
+        /// </summary>
+        /// <param name="cronSchedule">The cron schedule to check.</param>
+        /// <returns>True if monitoring should be disabled, false otherwise.</returns>
+        internal static bool ShouldDisableScheduleMonitor(CronSchedule cronSchedule, DateTime now)
+        {
+            // take the original expression minus the seconds portion
+            string expression = cronSchedule.InnerSchedule.ToString();
+            var expressions = expression.Split(' ');
+
+            // If any of the minute or higher fields contain non-wildcard expressions
```
mathewc, line 109: As per https://github.com/atifaziz/NCrontab/wiki/Crontab-Expression

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/TimerScheduleTests.cs
```diff
@@ -131,6 +132,40 @@ public void Create_InvalidSchedule()
...
+        [InlineData("* 30 * * * *", null, false)]
+        [InlineData("* */30 * * * *", null, false)]
+        [InlineData("0 * * * * *", null, false)]
+        [InlineData("30 * * * * *", null, false)]
+        [InlineData("* */20 * * * *", "10/30/2020 12:00:00 AM", false)]
+        [InlineData("* */20 * * * *", "10/30/2020 12:20:00 AM", false)]
+        [InlineData("* */20 * * * *", "10/30/2020 12:15:00 AM", false)]
+        [InlineData("* */20 * * * *", "10/30/2020 12:19:30 AM", false)]
+        [InlineData("* * 12 * * Mon", null, false)]
+        [InlineData("* 59 11 * * 1,2,3,4,5", null, false)]
```
wenhzha, line 151: This does not consider the range definition for cron.  ex: * 1-5 * * * *,  Every second during minute :02 of 1-5 of every hour.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
```diff
@@ -98,5 +94,41 @@ internal static TimerSchedule Create(TimerTriggerAttribute attribute, INameResol
...
+            // the schedule can be longer than 1 minute. I.e. the only way for all occurrences
+            // to be less than or equal to a minute is if all these fields are wild ("* * * * *").
+            bool hasNonSecondRestrictions = !expressions.Skip(1).All(p => p == "*");
+
```
wenhzha, line 113: To clear layout the expectation of this behavior change. This would start to enable schedule monitor for many of timer trigger that we used not to. For instance, cron: * 0 20 * * *,  since scale controller would only warm up the site a minute and half before, in the old behavior, due to the race condition,  we would not enable schedule monitor even we know each batch of operation is 24 hr apart. After this change, even each execution within one batch is every second, we would still enable schedule monitor.  This should not be too big of stress for the host or storage handle, but if we think about an worse case like * *20 * * *,  this might result so billing cost on the storage, but we can tell customer to explicit opt out. I am cool with overall approach but just point out the potential risk, thoughts?

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/TimerScheduleTests.cs
mathewc: That is also handled - I'll add a test case for that.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
mathewc: Yes, it's a bug fix, so cases where we weren't enabling but should have been, we'll now enable, which may improve reliability for previously affected apps.

Regarding storage transactions - yes, if you've defined a schedule that can run for a period of consecutive seconds, that will result in a storage update per invocation. But that's nothing storage can't handle, and operations are very inexpensive.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
```diff
@@ -98,5 +94,41 @@ internal static TimerSchedule Create(TimerTriggerAttribute attribute, INameResol
...
+                {
+                    // if there is more than one occurrence due in the next minute,
+                    // assume that this is a sub-minute constant schedule and disable
+                    // persistence
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
brettsam, line 132: It may be worth writing out (debug-level) logs here to help us diagnose any problems. Something that tells us if we enabled/disabled the monitor, and why.

brettsam (commented):
Looks good -- just one suggestion on some additional logging.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
mathewc: Added a debug log. The fact that our disable logic kicked in is sufficient - we log the schedule details elsewhere.

PR #681, opened 2020-11-03
Fixing binding data bug

mathewc:
Another issue discovered when trying to pull the updated http extension into Functions host, related to the lazy binding data changes. If a poco is being bound to, we must evaluate binding data immediately and apply it to the poco. In all other cases, the lazy binding data will be evaluated on demand correctly. I added a missing repro test for this scenario.

### DIFF
--- a/test/WebJobs.Extensions.Http.Tests/HttpTestHelpers.cs
+++ b/test/WebJobs.Extensions.Http.Tests/HttpTestHelpers.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Tests.Extensions.Http
 {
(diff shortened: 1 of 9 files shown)


brettsam (approved):

test/WebJobs.Extensions.Http.Tests/HttpTestHelpers.cs
```diff
@@ -7,15 +7,21 @@
...
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Tests.Extensions.Http
 {
     public class HttpTestHelpers
     {
         public static HttpRequest CreateHttpRequest(string method, string uriString, IHeaderDictionary headers = null, string body = null)
         {
+            var context = new DefaultHttpContext();
```
mathewc, line 18: Required for POST requests with a body. The http extension code is accessing these services, and we weren't mocking this out fully.

test/WebJobs.Extensions.Http.Tests/HttpTriggerEndToEndTests.cs
```diff
@@ -60,6 +60,29 @@ public async Task BasicInvoke()
             await _jobHost.CallAsync(method, new { req = request });
         }
 
+        [Fact]
+        public async Task BindToPoco()
```
mathewc, line 64: Failed before the fix.

PR #691, opened 2020-11-25
Adding support to provide start date time for cosmos feed start

adityavaish:
Exposing StartTime of ChangeFeedProcessorOptions to client attribute so that a client to specify the start point of the feed. The date time should ideally be in UTC but it can take any format that can be parsed by C# DateTime

Added default test case to check the added attribute.


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -166,5 +166,12 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
         /// </remarks>
         /// </summary>
         public bool UseDefaultJsonSerialization { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Gets or sets the DateTime from which feed should start from. 
+        /// This is useful when you need to skip past and start from a checkpoint. It is best to use UTC date / time format to avoid any discrepencies.
```
ealsur: We need to add validation, what happens if the user sets StartFrom and StartFromBeginning at the same time?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -166,5 +166,12 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
...
+        /// </summary>
+        public string StartFrom { get; set; }
```
ealsur: We already have a StartFromBeginning, should this be StartFromTime ?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,19 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
             ChangeFeedProcessorOptions processorOptions = BuildProcessorOptions(attribute);
+            
             processorOptions.StartFromBeginning = attribute.StartFromBeginning;
+
             if (attribute.MaxItemsPerInvocation > 0)
             {
                 processorOptions.MaxItemCount = attribute.MaxItemsPerInvocation;
             }
 
+            if (DateTime.TryParse(attribute.StartFrom, out var startFromDateTime))
```
ealsur: Can we provide a friendlier message if this does not parse? Other validations in the same file throw InvalidOperationException I believe

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -400,7 +402,7 @@ public static IEnumerable<ParameterInfo[]> GetParameters()
         // These will set ConnectionStringSetting, which LeaseConnectionStringSetting should also use by default
         private static class ValidCosmosDBTriggerBindigsWithChangeFeedOptions
         {
-            public static void Func1([CosmosDBTrigger("aDatabase", "aCollection", ConnectionStringSetting = "CosmosDBConnectionString", MaxItemsPerInvocation = 10, StartFromBeginning = true)] IReadOnlyList<Document> docs)
+            public static void Func1([CosmosDBTrigger("aDatabase", "aCollection", ConnectionStringSetting = "CosmosDBConnectionString", MaxItemsPerInvocation = 10, StartFromBeginning = true, StartFrom = "2020-11-25T22:36:29Z")] IReadOnlyList<Document> docs)
```
ealsur: Need to test scenarios where it does not parse correctly and if we add a validation to not be used with StartFromBeginning

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -359,6 +359,8 @@ public async Task ValidChangeFeedOptions_Succeed(ParameterInfo parameter)
             Assert.Equal(new Uri("https://fromSettings"), binding.LeaseCollectionLocation.Uri);
             Assert.Equal(10, binding.ChangeFeedProcessorOptions.MaxItemCount);
             Assert.True(binding.ChangeFeedProcessorOptions.StartFromBeginning);
+            Assert.NotNull(binding.ChangeFeedProcessorOptions.StartTime);
+            Assert.Equal(binding.ChangeFeedProcessorOptions.StartTime, DateTime.Parse("2020-11-25T22:36:29Z"));
```
ealsur: This needs a separate test, cannot be the same as setting StartFromBeginning

ealsur (commented):
@brettsam Do we have any other trigger that passes datetime information? What is the recommendation on the format?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
adityavaish: There should be no conflict. StartFromBeginning has no effect when leases are specified and created.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
adityavaish: StartFromTime is a fair option. I didn't want to restrict it to just time as its really a datetime :) . Let me know if you strongly feel the other way and I can make the change.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
adityavaish: Fixed

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
ealsur: StartFromTime also has no effect if leases are created, it only is useful to define the first state. Same as StartFromBeginning.

The validation is for the user's sake, if they set StartFromBeginning = true, and StartTime = "XX", which is the start point? It is not clear for the user, so we need to avoid combinations that can lead to an undefined behavior.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
ealsur: We could use StartFromDateTime too, anything that is more precise than StartFrom

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,23 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
+            {
+                processorOptions.StartTime = startFromDateTime;
+            }
+            else if (!string.IsNullOrEmpty(attribute.StartFrom))
+            {
+                throw new ArgumentException("Specified StartFrom parameter is not in appropriate DateTime format. Please refer: https://docs.microsoft.com/en-us/dotnet/api/system.datetime.parse?view=net-5.0#parsing-and-cultural-conventions");
```
ealsur: Remove the /en-us/ from any links

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,23 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
             ChangeFeedProcessorOptions processorOptions = BuildProcessorOptions(attribute);
+            
             processorOptions.StartFromBeginning = attribute.StartFromBeginning;
+
             if (attribute.MaxItemsPerInvocation > 0)
             {
                 processorOptions.MaxItemCount = attribute.MaxItemsPerInvocation;
             }
 
+            if (DateTime.TryParse(attribute.StartFrom, out var startFromDateTime))
```
ealsur: Should it be only attempting parsing if the value is not null or whitespace? To avoid it altogether when its not really required

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
adityavaish: updated

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
adityavaish: updated

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
adityavaish: fixed

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
adityavaish: updated

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
adityavaish: Added few combinations

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
adityavaish: Added few combinations

adityavaish:
A gentle ping on the review!

adityavaish:
@fabiocav @brettsam - please help review

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -138,6 +138,13 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
         /// </summary>
         public bool StartFromBeginning { get; set; } = false;
 
+        /// <summary>
+        /// Optional.
+        /// Gets or sets the DateTime from which feed should start from. 
+        /// This is useful when you need to skip past and start from a checkpoint. It is best to use UTC date / time format to avoid any discrepencies.
```
ealsur: I would not mention anything about checkpoints and the UTC timezone should be a `<remark>`.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -138,6 +138,13 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
         /// </summary>
         public bool StartFromBeginning { get; set; } = false;
 
+        /// <summary>
+        /// Optional.
+        /// Gets or sets the DateTime from which feed should start from.
```
ealsur: DateTime capitalized is related to C#, but the trigger is used in other languages.

```suggestion
        /// Gets or sets the a date and time to initialize the change feed read operation from.
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,36 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
+
             if (attribute.MaxItemsPerInvocation > 0)
             {
                 processorOptions.MaxItemCount = attribute.MaxItemsPerInvocation;
             }
 
+            var startFromTime = DateTime.UtcNow;
+
+            if (!string.IsNullOrEmpty(attribute.StartFromTime) && 
+                !DateTime.TryParse(attribute.StartFromTime, out startFromTime))
```
ealsur: You can defer the variable declaration to the `out` definition":

```suggestion
                !DateTime.TryParse(attribute.StartFromTime, out DateTime startFromTime))
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,36 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
+            {
+                throw new InvalidOperationException("Specified StartFrom parameter is not in appropriate DateTime format. Please refer: https://docs.microsoft.com/dotnet/api/system.datetime.parse?view=net-5.0#parsing-and-cultural-conventions");
```
ealsur:
```suggestion
                throw new InvalidOperationException("Specified StartFromTime parameter is not in appropriate DateTime format. Please refer: https://docs.microsoft.com/dotnet/api/system.datetime.parse?view=net-5.0#parsing-and-cultural-conventions");
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,36 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
+            }
+
+            if (attribute.StartFromBeginning && 
+                !string.IsNullOrEmpty(attribute.StartFromTime))
+            {
+                throw new InvalidOperationException("Only one of StartFromBeginning or StartFromTime can be used");
+            }
+
+            processorOptions.StartFromBeginning = attribute.StartFromBeginning;
```
ealsur: Remove, it's a duplicate. It is also set a couple of lines above.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,36 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
+
+            if (!attribute.StartFromBeginning &&
+                !string.IsNullOrEmpty(attribute.StartFromTime))
```
ealsur: I think you can simply check for the StartFromTime, at this point, you already validated that StartFromBeginning cannot be true if StartFromTime is not null.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,36 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
 
+            var startFromTime = DateTime.UtcNow;
+
+            if (!string.IsNullOrEmpty(attribute.StartFromTime) && 
+                !DateTime.TryParse(attribute.StartFromTime, out startFromTime))
+            {
+                throw new InvalidOperationException("Specified StartFrom parameter is not in appropriate DateTime format. Please refer: https://docs.microsoft.com/dotnet/api/system.datetime.parse?view=net-5.0#parsing-and-cultural-conventions");
+            }
+
+            if (attribute.StartFromBeginning &&
```
ealsur: Maybe a cleaner way to do this is:

```
if (!string.IsNullOrEmpty(attribute.StartFromTime))
{
    if (attribute.StartFromBeginning)
    {
        throw new InvalidOperationException(....);
    }

    if (!DateTime.TryParse(attribute.StartFromTime, out DateTime startFromTime))
    {
          throw new InvalidOperationException(...);
    }

    processorOptions.StartTime = startFromTime;
}
else
{
     processorOptions.StartFromBeginning = attribute.StartFromBeginning;
}

File: test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs [hunk 15]
```diff
@@ -354,11 +354,14 @@ public async Task ValidChangeFeedOptions_Succeed(ParameterInfo parameter)
 
             CosmosDBTriggerBinding binding = (CosmosDBTriggerBinding)await provider.TryCreateAsync(new TriggerBindingProviderContext(parameter, CancellationToken.None));
 
+            CosmosDBTriggerAttribute cosmosDBTriggerAttribute = parameter.GetCustomAttribute<CosmosDBTriggerAttribute>(inherit: false);
+
             Assert.Equal(typeof(IReadOnlyList<Document>), binding.TriggerValueType);
             Assert.Equal(new Uri("https://fromSettings"), binding.DocumentCollectionLocation.Uri);
             Assert.Equal(new Uri("https://fromSettings"), binding.LeaseCollectionLocation.Uri);
-            Assert.Equal(10, binding.ChangeFeedProcessorOptions.MaxItemCount);
-            Assert.True(binding.ChangeFeedProcessorOptions.StartFromBeginning);
```
ealsur, line 361: When is the StartFromBeginning true now validated?

adityavaish: its on line 363 where input (true) is being compared.

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBTriggerAttributeBindingProviderTests.cs
```diff
@@ -404,13 +407,23 @@ public static void Func1([CosmosDBTrigger("aDatabase", "aCollection", Connection
             {
             }
 
+            public static void Func2([CosmosDBTrigger("aDatabase", "aCollection", ConnectionStringSetting = "CosmosDBConnectionString", MaxItemsPerInvocation = 10, StartFromTime = "2020-11-25T22:36:29Z")] IReadOnlyList<Document> docs)
```
ealsur, line 410: Missing one case where StartFromBeginning is true, right?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -58,12 +59,33 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
...
+            {
+                if (attribute.StartFromBeginning)
+                {
+                    throw new InvalidOperationException("Only one of StartFromBeginning or StartFromTime can be used");
+                }
+
+                if (!DateTime.TryParse(attribute.StartFromTime, out DateTime startFromTime))
+                {
+                    throw new InvalidOperationException("Specified StartFromTime parameter is not in appropriate DateTime format. Please refer: https://docs.microsoft.com/dotnet/api/system.datetime.parse?view=net-5.0#parsing-and-cultural-conventions");
+                }
```
brettsam, line 80: This link may break at some point in the future. Can we instead provide an example of a format that will work?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
```diff
@@ -138,6 +138,13 @@ public CosmosDBTriggerAttribute(string databaseName, string collectionName)
         /// </summary>
         public bool StartFromBeginning { get; set; } = false;
 
+        /// <summary>
+        /// Optional.
+        /// Gets or sets the a date and time to initialize the change feed read operation from.
+        /// <remark>It is best to use UTC date / time format to avoid any discrepencies.</remark>
+        /// </summary>
+        public string StartFromTime { get; set; }
```
brettsam, line 146: Are we able to use `DateTime` (or nullable, probably) as the type here? Will that be correctly deserialized?

brettsam:
> @brettsam Do we have any other trigger that passes datetime information? What is the recommendation on the format?

I don't think we do -- I was just trying to think through that as well. I know we have some that use TimeSpan, let me double-check how we handle that.

ealsur:
My concern is more around other languages, if the ISO format would work fine for them or there is a different one. Otherwise I see no blockers on the PR other than the test gap at this point.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
ealsur: How does that get exposed to other languages like JS or Python if we do it as DateTime or DateTimeOffset?

brettsam: Nevermind... that won't work for an attribute anyway... so it'll have to be a string.

brettsam, line 146: My two biggest concerns with this is time zones and parsing.
1. Does the ChangeFeedProcessor expect UTC? Can we just say that it will be interpreted as UTC no matter what (and then force that?)
2. I mentioned elsewhere, but in the summary including an example of an ISO string would be useful here.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
adityavaish: This link captures all the possible combinations that the API can handle. Its a long doc to replicate?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
adityavaish: #1 The datetime is parsed in C# code. Ideally user can specify in any format that is parsable by C# library. Have added a comment to ideally use ISO 8601.

#2 Added ISO 8601 example in comments

adityavaish: @ealsur - thoughts?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
brettsam: Whoops -- I accidentally deleted your comment when I was trying to delete a draft comment of mine!

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
brettsam: I'd prefer something like this, which gives an explicit example right in the comment:

`The recommended format is ISO 8601 with the UTC designator. For example: "2021-02-16T14:19:29Z".`

Yes, other formats will work, but this one is guaranteed to work and reduces the need to lookup a separate documentation page and look through all of the options.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
brettsam: *Here's my comment in response to yours...*

Yes, but that link may break at some point in the future (it has `net-5.0` in the URL, for example). Also, this may be used by languages other than C# in Functions, which means linking to .NET `DateTime.Parse()` documentation can be confusing. 

I'd prefer something like this, which gives an explicit example right in the error message and doesn't mention .NET.:

`The specified StartFromTime parameter is not in the correct format. Please use the ISO 8601 format with the UTC designator. For example: "2021-02-16T14:19:29Z".`

As I mentioned in the doc comments for the `StartFromTime` property: Yes, other formats will work, but this one is guaranteed to work and reduces the need to lookup a separate documentation page and look through all of the options.

brettsam (commented):
I have one other comment that I can't directly link to (and maybe it can be an immediate follow-up PR by @ealsur?) -- but now that the parsing of attribute strings can have some interesting side effects (time zones...), it may be good to explicitly log the fully-resolved `ProcessorOptions` with a `_logger.LogDebug(...)` statement here: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs#L129. That would help immensely when trying to support folks that may have configured something slightly incorrectly.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
adityavaish: Resolved. Update the comment

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
adityavaish: Updated

ealsur (approved):

brettsam (approved):

phola:
Hi this is really useful thanks! Think there maybe an issue with node.js functions though, this is what I'm getting:

image

PR #696, opened 2020-12-18
Disable admin overrides except admin/warmup in WebJobsRouter.

safihamid:
addresses: https://github.com/Azure/azure-functions-host/pull/6993

### DIFF
--- a/src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
+++ b/src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
@@ -51,6 +51,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
 
         public Task RouteAsync(RouteContext context)
         {
+            // /admin/* routes should not be allowed to be overriden by proxies or function routes.
+            var path = context.HttpContext.Request.Path;
+            if (path.StartsWithSegments(new PathString("/admin"), System.StringComparison.OrdinalIgnoreCase))
+            {
+                // admin/warmup is handled by function routes
+                if (!path.StartsWithSegments(new PathString("/admin/warmup"), System.StringComparison.OrdinalIgnoreCase))
+                {
+                    return Task.CompletedTask;
+                }
+            }
+
             // If this key is set in HttpContext, we first match against Function routes then Proxies.
             if (context.HttpContext.Items.ContainsKey(HttpExtensionConstants.AzureWebJobsUseReverseRoutesKey))
             {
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
```diff
@@ -51,6 +51,17 @@ private void InitializeRouteCollections()
...
         public Task RouteAsync(RouteContext context)
         {
+            // /admin/* routes should not be allowed to be overriden by proxies or function routes.
+            var path = context.HttpContext.Request.Path;
+            if (path.StartsWithSegments(new PathString("/admin"), System.StringComparison.OrdinalIgnoreCase))
+            {
+                // admin/warmup is handled by function routes
+                if (!path.StartsWithSegments(new PathString("/admin/warmup"), System.StringComparison.OrdinalIgnoreCase))
+                {
+                    return Task.CompletedTask;
```
ConnorMcMahon, line 61: Just to clarify, if we return a `Task.CompletedTask`, presumably that means that some other router further down the pipeline will handle the request (or if none do, then we return a 404?).

Is there a reason we went with this approach instead of having the Admin controller get routing priority over HTTP triggers?

Also, this only bypasses admin requests, but what about extension webhook endpoints? I know both EventGrid and Durable Functions use the webhook endpoints extensively for functionality.

ConnorMcMahon (approved):
LGTM!

I think we still will want to check if this overrides the extension webhook endpoints, but at this point the priority is fixing the regression.

src/WebJobs.Extensions.Http/Routing/WebJobsRouter.cs
mathewc, line 61: So by returning a completed Task without routing below, this router skips and lets the base mvc routing handle the request? I wonder if we don't just have an ordering issue at the point where we register the WebJobs router [here](https://github.com/Azure/azure-functions-host/blob/9ac904e34b744d95a6f746921556235d4b2b3f0f/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L69). We register our custom router BEFORE we call AddMvc meaning our function routing takes precedence over mvc handling for all our admin controller routes? Perhaps if we AddMvc before we add our custom router all built in admin routes will match first?

It feels a bit messy to be putting these exclusions in here, since anytime we need to adjust this, it requires a separate nuget package update. If the above doesn't work, another option could be to create a simple filtering IWebJobsRouter in the host that composes the one returned from the http package, overriding RouteAsync, applying any filtering and delegating?

mathewc: I'd like to get @fabiocav's thoughts on this

safihamid: @mathewc yes the ordering was wrong. needed to add UseMVC before UseHttpBindingRouting. please look at other PR in webhost https://github.com/Azure/azure-functions-host/pull/6993 once all tests pass there and all look good there I will revert this merge here. thanks

mathewc: Approved the other PR

PR #664, opened 2020-09-22
SendGrid API response handling

nicbavetta:
This pull request contains modifications to enable extensibility through SendGrid API response handling. Previously, if a response from the SendGrid API indicated a result other than success (defined, by this extension, as status code >= 300), an exception would be thrown when the _SendGridMessageAsyncCollector_ was flushed. This implementation made it difficult to react to responses in meaningful ways.

**Additions**

1. **ISendGridResponseHandler** - This interface can be implemented to allow business logic to be executed against responses from the SendGrid API.

2. **DefaultSendGridResponseHandler** - The default implementation for the interface above. If consumers of this extension do not implement and register the _ISendGridResponseHandler_ with DI, then the _DefaultSendGridResponseHandler_ will be used. This handler implements the logic that was previously implemented when SendGrid responses were received with status codes >= 300. The status code evaluation has been changed to >= 400, per [SendGrid documentation](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html).

**Modifications**
1. The _ISendGridClient_  and  _SendGridClient_ have been removed in favor of leveraging the client interface/implementation provided by the SendGrid SDK.
2. Namespaces in this project have been made to be consistent.
3. As mentioned previously, SendGrid response status code evaluation has been changed to >= 400, per [SendGrid documentation](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html).
4. Some null-checks have been added in modified constructors to better convey explicit requirements.

PR #709, opened 2021-02-15
Adding SendGrid response handling

nicbavetta:
Per request, I have stripped down [my last pull request](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/664) to exclude any additional/unrelated refactoring.

**There were several tests failing prior to my work. These tests will also fail in the automated build. They are unrelated to my work and should be ignored when evaluating my pull request.** 

This pull request contains modifications to enable extensibility through SendGrid API response handling. Previously, if a response from the SendGrid API indicated a result other than success (defined, by this extension, as status code >= 300), an exception would be thrown when the _SendGridMessageAsyncCollector_ was flushed. This implementation made it difficult to react to responses in meaningful ways.

**Additions**

1. **ISendGridResponseHandler** - This interface can be implemented to allow business logic to be executed against responses from the SendGrid API.

2. **DefaultSendGridResponseHandler** - The default implementation for the interface above. If consumers of this extension do not implement and register the _ISendGridResponseHandler_ with DI, then the _DefaultSendGridResponseHandler_ will be used. This handler implements the logic that was previously implemented when SendGrid responses were received with status codes >= 300. The status code evaluation has been changed to >= 400, per [SendGrid documentation](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html).

### DIFF
--- a/src/WebJobs.Extensions.SendGrid/Config/DefaultSendGridResponseHandler.cs
+++ b/src/WebJobs.Extensions.SendGrid/Config/DefaultSendGridResponseHandler.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.Bindings;
+using SendGrid;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Config
+{
+    /// <summary>
+    /// The default handler of responses received from APIs call to SendGrid.
+    /// </summary>
+    public class DefaultSendGridResponseHandler : ISendGridResponseHandler
+    {
+        /// <summary>
+        /// Handles the response.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.SendGrid/Config/DefaultSendGridResponseHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Handles the response.
+        /// </summary>
+        /// <param name="response">The response to handle.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns>A task.</returns>
+        public async Task HandleAsync(Response response, CancellationToken cancellationToken)
+        {
+            // https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html
+            if ((int)response.StatusCode >= 400)
```
jviau, line 26: This changed from `>= 300` to `>= 400`. Any reason? Was 300 a mistake before?

jviau:
Closing PR due to inactivity.

@nicbavetta please re-open the PR if you believe it is still relevant and address the questions above.

PR #714, opened 2021-03-02
Migrating StorageScheduleMonitor to use Track2

karshinlin:
- StorageScheduleMonitor now uses AzureStorageProvider in WebJobs.Host.Storage for storage dependency

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitorV2.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitorV2.cs
@@ -1,0 +1,183 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Storage.Blobs;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Timers
+{
+    /// <summary>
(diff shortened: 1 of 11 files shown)


src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitorV2.cs
```diff
@@ -0,0 +1,183 @@
...
+                if (_containerClient == null)
+                {
+                    if (_storageOptions.InternalSasBlobContainer != null)
+                    {
+                        _containerClient = new BlobContainerClient(new Uri(_storageOptions.InternalSasBlobContainer));   
+                    }
+                    else if (_storageOptions.InternalContainerName != null)
+                    {
+                        // TODO: Should we throw if we can't create the client?
+                        if (!_azureStorageProvider.TryGetBlobServiceClientFromConnection(out BlobServiceClient blobServiceClient, ConnectionStringNames.Storage))
```
pakrym, line 97: Should we make AzureStorageProvider aware of JobHostInternalStorageOptions?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitorV2.cs
```diff
@@ -0,0 +1,183 @@
...
+                        {
+                            throw new InvalidOperationException($"Could not create BlobServiceClient with connection {ConnectionStringNames.Storage}");
+                        }
+                        _containerClient = blobServiceClient.GetBlobContainerClient(HostContainerName);
+                    }
+                    else
+                    {
+                        // TODO: Should we throw if we can't create the client?
+                        if (!_azureStorageProvider.TryGetBlobServiceClientFromConnection(out BlobServiceClient blobServiceClient, ConnectionStringNames.Storage))
```
pakrym, line 106: It might be useful to add a GetBlobServiceClientFromConnection method to the AzureStorageProvider. We can have a default implementation that would throw a nice, consistent error message

src/WebJobs.Extensions/Extensions/Timers/Scheduling/StorageScheduleMonitorV2.cs
karshinlin: Hmm good point. It might be wise to consolidate all this logic with container overrides. @fabiocav to chime in.

PR #730, opened 2021-07-08
Cosmos DB: Add support for Gremlin accounts

ealsur:
Gremlin accounts support Change Feed using the SQL API, but until V3 SDK, they could not use the Change Feed Processor.

V3 added support for Gremlin recently (https://github.com/Azure/azure-cosmos-dotnet-v3/pull/2491).

This PR adds the required logic to the CreateIfNotExist for lease collections.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
         private const string CosmosDBTriggerUserAgentSuffix = "CosmosDBTriggerFunctions";
         private const string SharedThroughputRequirementException = "Shared throughput collection should have a partition key";
         private const string LeaseCollectionRequiredPartitionKey = "/id";
+        private const string LeaseCollectionRequiredPartitionKeyFromGremlin = "/partitionKey";
         private readonly IConfiguration _configuration;
         private readonly INameResolver _nameResolver;
         private readonly CosmosDBOptions _options;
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -134,7 +136,16 @@ internal static TimeSpan ResolveTimeSpanFromMilliseconds(string nameOfProperty,
...
         private static async Task CreateLeaseCollectionIfNotExistsAsync(CosmosClient cosmosClient, string databaseName, string collectionName, int? throughput)
         {
-            await CosmosDBUtility.CreateDatabaseAndCollectionIfNotExistAsync(cosmosClient, databaseName, collectionName, LeaseCollectionRequiredPartitionKey, throughput);
+            try
+            {
+                await CosmosDBUtility.CreateDatabaseAndCollectionIfNotExistAsync(cosmosClient, databaseName, collectionName, LeaseCollectionRequiredPartitionKey, throughput);
+            }
+            catch (CosmosException cosmosException) 
+                when (cosmosException.StatusCode == System.Net.HttpStatusCode.BadRequest
+                    && cosmosException.Message.Contains("invalid for Gremlin API"))
```
brettsam, line 145: Is this message guaranteed to stay the same? Wondering if there's any error code or something that would be safer to check.

brettsam (approved):
Just one question -- if there's no good alternative, go ahead and get it in.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: Sadly no :( The backend does not return a specific error code (Status Code + Sub status code). It's a plain HTTP 400, only way to detect it at this point is the error message.

PR #733, opened 2021-08-02
[Cosmos DB] Rename ConnectionStringSetting and collection

ealsur:
Other extensions are using the name `Connection` for attribute mapping that works for connection strings (example https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/EventHubAttribute.cs#L33) but would also work for AAD information.

Since we are already on a breaking change train (moving to V3 SDK), might as well rename it to be consistent (anyway, using `ConnectionStringSetting` would not sound great for AAD).

As part of the naming alignment, V3 SDK documentation talks about Containers, not Collections anymore, so the property names like `CollectionName` are renamed to `ContainerName`.


fabiocav (approved):

fabiocav:
@ealsur looks good. As long as this is going in a new major version.

Thanks for sending refactoring work like this as a standalone PR!

PR #736, opened 2021-08-11
Cosmos DB - Add Support for AAD authentication

ealsur:
This PR expands the support for authentication to include AAD.

Users can define on the Cosmos DB Trigger or binding a Connection value that points to a configuration that either:

* Is a string and represents a valid Cosmos DB Connection String
```json
{
        "MyConnection": "<cosmos db connection string>"
}
```
* Is a JSON node with valid AAD information:

```json
{
	"MyConnection" : {
             "accountEndpoint": "https://myCosmosDBaccount.documents.azure.com:443/",
	     "clientId": "AAD client Id",
	     "tenantId": "AAD Tenant Id",
	     "clientSecret": "AAD secret"
       }
}
```

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Config/DefaultCosmosDBServiceFactory.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Config/DefaultCosmosDBServiceFactory.cs
@@ -1,15 +1,107 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
+using Azure.Core;
 using Microsoft.Azure.Cosmos;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 {
     internal class DefaultCosmosDBServiceFactory : ICosmosDBServiceFactory
     {
-        public CosmosClient CreateService(string connectionString, CosmosClientOptions cosmosClientOptions)
+        private readonly IConfiguration _configuration;
+        private readonly AzureComponentFactory _componentFactory;
+
+        public DefaultCosmosDBServiceFactory(
+            IConfiguration configuration,
+            AzureComponentFactory componentFactory)
         {
-            return new CosmosClient(connectionString, cosmosClientOptions);
+            this._configuration = configuration;
+            this._componentFactory = componentFactory;
+        }
+
+        public CosmosClient CreateService(string connectionName, CosmosClientOptions cosmosClientOptions)
+        {
+            CosmosConnectionInformation cosmosConnectionInformation = this.ResolveConnectionInformation(connectionName);
+            if (cosmosConnectionInformation.UsesConnectionString)
+            {
+                // Connection string based auth
+                return new CosmosClient(cosmosConnectionInformation.ConnectionString, cosmosClientOptions);
+            }
+
+            // AAD auth
+            return new CosmosClient(cosmosConnectionInformation.AccountEndpoint, cosmosConnectionInformation.Credential, cosmosClientOptions);
+        }
+
+        private CosmosConnectionInformation ResolveConnectionInformation(string connection)
+        {
+            var connectionSetting = connection ?? Constants.DefaultConnectionStringName;
+            IConfigurationSection connectionSection = GetWebJobsConnectionStringSectionCosmos(this._configuration, connectionSetting);
+            if (!connectionSection.Exists())
+            {
+                // Not found
+                throw new InvalidOperationException($"Cosmos DB connection configuration '{connectionSetting}' does not exist. " +
+                                                    $"Make sure that it is a defined App Setting.");
+            }
+
+            if (!string.IsNullOrWhiteSpace(connectionSection.Value))
+            {
+                return new CosmosConnectionInformation(connectionSection.Value);
+            }
+            else
+            {
+                string accountEndpoint = connectionSection["accountEndpoint"];
+                if (string.IsNullOrWhiteSpace(accountEndpoint))
+                {
+                    // Not found
+                    throw new InvalidOperationException($"Connection should have an 'accountEndpoint' property or be a " +
+                        $"string representing a connection string.");
+                }
+
+                TokenCredential credential = _componentFactory.CreateTokenCredential(connectionSection);
+                return new CosmosConnectionInformation(accountEndpoint, credential);
+            }
+        }
+
+        public static IConfigurationSection GetWebJobsConnectionStringSectionCosmos(IConfiguration configuration, string connectionStringName)
+        {
+            // first try a direct unprefixed lookup
+            IConfigurationSection section = WebJobsConfigurationExtensions.GetConnectionStringOrSetting(configuration, connectionStringName);
+
+            if (!section.Exists())
+            {
+                // next try prefixing
+                string prefixedConnectionStringName = WebJobsConfigurationExtensions.GetPrefixedConnectionStringName(connectionStringName);
+                section = WebJobsConfigurationExtensions.GetConnectionStringOrSetting(configuration, prefixedConnectionStringName);
+            }
+
+            return section;
+        }
+
+        private class CosmosConnectionInformation
+        {
+            public CosmosConnectionInformation(string connectionString)
+            {
+                this.ConnectionString = connectionString;
+                this.UsesConnectionString = true;
+            }
+
+            public CosmosConnectionInformation(string accountEndpoint, TokenCredential tokenCredential)
+            {
+                this.AccountEndpoint = accountEndpoint;
+                this.Credential = tokenCredential;
+                this.UsesConnectionString = false;
+            }
+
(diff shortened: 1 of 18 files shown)


philipthomas-MSFT (approved):

src/WebJobs.Extensions.CosmosDB/WebJobsConfigurationExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
+{
+    internal static class WebJobsConfigurationExtensions
+    {
+        private const string WebJobsConfigurationSectionName = "AzureWebJobs";
+
+        public static IConfigurationSection GetWebJobsConnectionStringSection(this IConfiguration configuration, string connectionStringName)
```
brettsam, line 12: Can you use these helpers directly? https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs#L15

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
-    <Version>$(CosmosDBVersion)</Version>
+    <Version>$(CosmosDBVersion)-preview</Version>
```
brettsam: Will this be the only preview? I usually use `preview1` just in case I need another one...

brettsam:
A couple of minor comments....

src/WebJobs.Extensions.CosmosDB/WebJobsConfigurationExtensions.cs
pakrym: Those return a string and we need an `IConfigurationSection` here.

pakrym: It might be nice if we added a  `IConfigurationSection`-returning overload to `IConfigurationExtensions`. Most extensions that do AAD auth use it.

src/WebJobs.Extensions.CosmosDB/Config/DefaultCosmosDBServiceFactory.cs
```diff
@@ -1,15 +1,107 @@
...
+                    throw new InvalidOperationException($"Connection should have an 'accountEndpoint' property or be a " +
+                        $"string representing a connection string.");
+                }
+
+                TokenCredential credential = _componentFactory.CreateTokenCredential(connectionSection);
+                return new CosmosConnectionInformation(accountEndpoint, credential);
+            }
+        }
+
+        public static IConfigurationSection GetWebJobsConnectionStringSectionCosmos(IConfiguration configuration, string connectionStringName)
```
pakrym: Why not use the GetWebJobsConnectionStringSection from `WebJobsConfigurationExtensions.cs`

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -185,31 +182,10 @@ private void ThrowMissingConnectionStringException(bool isLeaseConnectionString
...
-                {
-                    throw new InvalidOperationException($"Unable to resolve app setting for property '{nameof(CosmosDBTriggerAttribute)}.{propertyName}'. Make sure the app setting exists and has a valid value.");
-                }
-
-                return resolvedString;
-            }
-
-            // If that didn't exist, fall back to options.
-            return _options.ConnectionString;
+                $"The CosmosDBTrigger {leaseString}connection must be set either via a '{Constants.DefaultConnectionStringName}' configuration or via the {attributeProperty} property.");
```
pakrym, line 188: We've changed a similar message in CosmosDBExtensionConfigProvider.cs to "The {attributeProperty} property cannot be an empty value."

src/WebJobs.Extensions.CosmosDB/Config/DefaultCosmosDBServiceFactory.cs
ealsur: @brettsam had the same question, the problem is, `WebJobsConfigurationExtensions` resolve using the WebJobs root path `AzureWebJobs:extensions:<config name>`: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs#L42-L46

The other method available `GetConnectionStringOrSetting` only returns a string, not the Section, so we cannot use it.

If we take a look at the other repository uses these methods, which are the ones I copied: https://github.com/Azure/azure-sdk-for-net/blob/303b75341ba17aecf239a3232d28ca76209a5346/sdk/extensions/Microsoft.Azure.WebJobs.Extensions.Clients/src/Shared/WebJobsConfigurationExtensions.cs#L12

src/WebJobs.Extensions.CosmosDB/WebJobsConfigurationExtensions.cs
ealsur: The problem is, `WebJobsConfigurationExtensions` resolve using the WebJobs root path `AzureWebJobs:extensions:<config name>`: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs#L42-L46

The other method available `GetConnectionStringOrSetting` only returns a string, not the Section, so we cannot use it.

If we take a look at the other repository uses these methods, which are the ones I copied: https://github.com/Azure/azure-sdk-for-net/blob/303b75341ba17aecf239a3232d28ca76209a5346/sdk/extensions/Microsoft.Azure.WebJobs.Extensions.Clients/src/Shared/WebJobsConfigurationExtensions.cs#L12

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
ealsur: Done

src/WebJobs.Extensions.CosmosDB/WebJobsConfigurationExtensions.cs
pakrym: Yeah, I agree with you about the need for getting a section but you already have a copy of the WebJobsConfigurationExtensions.cs in this PR. Why not use it?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
ealsur: It's a bit of a different scenario, the `CosmosDBExtensionConfigProvider.cs` validates that a Binding does not have a Connection that is empty (it is set to an empty value). If it's not set, it will attempt later on to use the default config section, and if it's not found, error out.

For Triggers, that Validation step earlier does not occur, so we need here to cover 2 things, that they are not setting it to empty and that the default is present. Hence the message covers both cases.

src/WebJobs.Extensions.CosmosDB/WebJobsConfigurationExtensions.cs
brettsam: Ah, I see now... makes sense...

brettsam (approved):

src/WebJobs.Extensions.CosmosDB/WebJobsConfigurationExtensions.cs
ealsur: > Yeah, I agree with you about the need for getting a section but you already have a copy of the WebJobsConfigurationExtensions.cs in this PR. Why not use it?

Not sure I follow. I did create the copy (of the Azure SDK repo) and I am using it in the `DefaultCosmosDBServiceFactory.cs` file.

pakrym: Looking at GetWebJobsConnectionStringSectionCosmos from DefaultCosmosDBServiceFactory. It seems to be very similar to this method (WebJobsConfigurationExtensions.GetWebJobsConnectionStringSection). Do we need both?

ealsur: Now I get it, yup, makes sense. Removed the `GetWebJobsConnectionStringSectionCosmos` in `DefaultCosmosDBServiceFactory`

src/WebJobs.Extensions.CosmosDB/Config/DefaultCosmosDBServiceFactory.cs
ealsur: Removed

pakrym:
LGTM

PR #737, opened 2021-08-25
Moving Timer startup invocations off the listener startup path

mathewc:
Addresses https://github.com/Azure/azure-functions-host/issues/5529

For full Functions E2E testing I verified the following scenario:

1) Functions app with a TimerTrigger (RunOnStartup=True) that delays for 30 seconds when run, and an HttpTrigger function
2) I started the functions host, then immediately pinged the http function

Before my fix, the startup invocation of the timer function delays host startup, causing the http request to hang. With the fix, the timer invocation is run in the background and doesn't block host startup, allowing it to return relatively quickly.

The issue is WebScriptHostService is an IHostedService, and as part of startup, it starts the ScriptHost, which starts all the listeners. The app can't start processing http requests until all the hosted services are started up. So the fact that WebScriptHostService awaits host/listener startup [here](https://github.com/Azure/azure-functions-host/blob/ed814c569434d65abcc22a169ac2d95f1ef79c32/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L286) means http requests can't be processed until that's complete. That means that http requests weren't processed until the 30s startup timer invocation completed.

### DIFF
--- a/test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
+++ b/test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Tests.Timers
         }
 
         [Fact]
-        public async Task StartAsync_SchedulePastDue_InvokesJobFunctionImmediately()
+        public async Task StartAsync_SchedulePastDue_SchedulesImmediateInvocation()
         {
             // Set this to true to ensure that the function is only executed once
             // In this case, because it is run on startup due to being behind schedule,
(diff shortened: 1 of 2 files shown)


test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
```diff
@@ -234,26 +245,39 @@ public async Task StartAsync_ScheduleNotPastDue_DoesNotInvokeJobFunctionImmediat
             CancellationToken cancellationToken = CancellationToken.None;
             await _listener.StartAsync(cancellationToken);
 
+            Assert.Null(_listener.StartupInvocation);
```
paulbatum, line 248: Nit: I am not sure this null check is necessary, given that the next line verifies that the function did not run, and it just ties the test close to the internal implementation details.

test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
```diff
@@ -203,6 +203,15 @@ public async Task StartAsync_SchedulePastDue_InvokesJobFunctionImmediately()
             CancellationToken cancellationToken = CancellationToken.None;
             await _listener.StartAsync(cancellationToken);
 
+            var startupInvocation = _listener.StartupInvocation;
+            Assert.NotNull(startupInvocation);
+            Assert.False(startupInvocation.RunOnStartup);
+            Assert.Equal(TimerListener.StartupInvocationContext.IntervalMS, _listener.Timer.Interval);
+            Assert.True(startupInvocation.IsPastDue);
+            Assert.Equal(default(DateTime), startupInvocation.OriginalSchedule);
```
paulbatum, line 211: Similar to my other comment - I would err on the side of testing the behavior (the function ran/did not run, the right logs were emitted), rather than checking the internal state.. It makes the tests require more changes if the internal implementation has to change again. Sometimes you have to do this type of stuff in tests because its hard to verify behavior but it looks like you have it pretty well covered in this case.

paulbatum (approved):
Looks good. Just a couple of personal nits about the tests being heavily tied to internal implementation of TimerListener.

brettsam (approved):
This is great -- a long-time subtle issue fixed!

test/WebJobs.Extensions.Tests/Extensions/Timers/Listener/TimerListenerTests.cs
mathewc: Thanks. I could go either way on this, but I like the stronger tests, so I'll leave these as is for now.

PR #742, opened 2021-09-01
Cosmos DB - AddAzureClientsCore should be called during registration

### LINKED ISSUE #741: CosmosDB 4.0.0-preview1 always throws an exception when AAD authentication is configured
Whenever I set up and run `CosmosDBTrigger` or `CosmosDB` to use AAD authentication, it throws a `NullReferenceException` and fails to start.

I checked and found that `AzureComponentFactory` is always passed null for `DefaultCosmosDBServiceFactory`. The root cause seems to be that `AddAzureClientsCore` is not registered in the DI.

Example (Service Bus Extension) :
https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs#L91-L93

#### Repro steps

1. Create new function (v3) project
2. Add `CosmosDBTrigger` function
3. Update `Microsoft.Azure.WebJobs.Extensions.CosmosDB` version `4.0.0-preview1`
4. Setup RBAC and connection string
5. Run function app

#### Expected behavior

An error always occurs when starting Function.

```
[2021-09-01T15:07:18.499Z] Microsoft.Azure.WebJobs.Host: Error indexing method 'Function2'. 
Microsoft.Azure.WebJobs.Extensions.CosmosDB: Cannot create container information for TodoItem in database SampleDB with lease leases in database SampleDB : Object reference not set to an instance of an object. 
Microsoft.Azure.WebJobs.Extensions.CosmosDB: Object reference not set to an instance of an object.
[2021-09-01T15:07:18.536Z] Error indexing method 'Function2'
```

#### Actual behavior

`CosmosDBTrigger` and Managed Identity can be used without any errors.

#### Known workarounds

- Install `Microsoft.Azure.WebJobs.Extensions.Storage` (At the same time, `AddAzureClientsCore` is added to the DI, so)

#### Related information 

* `Microsoft.Azure.WebJobs.Extensions.CosmosDB` : `4.0.0-preview1`
* https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs#L21-L38

### LINKED ISSUE #743: Serializer customization using `ICosmosDBSerializerFactory` does not work
I created a class that implements `ICosmosDBSerializerFactory` and tried to customize the default serializer, but it did not work correctly because `CreateSerializer` was not called.

It seems that after adding the customized `ICosmosDBSerializerFactory` to the DI, the Cosmos DB Extension overrides it with the default SerializerFactory.

The initialization by Startup seems to be done in the order of `Function App` => `Extensions`, so if the Extension provides an extensible interface, it need to use `TryAddSingleton`. (In Function Extensions, it could be a convention to use `TryAddSingleton`)

- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/StorageBlobsWebJobsBuilderExtensions.cs
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues/src/StorageQueuesWebJobsBuilderExtensions.cs

#### Repro steps

```csharp
public class Startup : FunctionsStartup
{
    public override void Configure(IFunctionsHostBuilder builder)
    {
        builder.Services.AddSingleton<ICosmosDBSerializerFactory, MyCosmosDBSerializerFactory>();
    }
}

public class MyCosmosDBSerializerFactory : ICosmosDBSerializerFactory
{
    public CosmosSerializer CreateSerializer()
    {
        var options = new JsonSerializerOptions
        {
            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
        };

        // From https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/SystemTextJson/CosmosSystemTextJsonSerializer.cs
        return new CosmosSystemTextJsonSerializer(options);
    }
}
```

#### Expected behavior

The customized SerializerFactory will be used.

#### Actual behavior

The default implementation of SerializerFactory is always used.

#### Known workarounds

Not available.

#### Related information 

* `Microsoft.Azure.WebJobs.Extensions.CosmosDB` : `4.0.0-preview1`
* https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs#L36

ealsur:
Current preview build expects users to register and call `AddAzureClientsCore` during webjob initialization, this can be moved to be part of the Cosmos DB Extension registration.

We also detected that binding types `int?` do not work on runtime.

And if a user was trying to customize the `ICosmosDBSerializerFactory`, current code was only working if the customization happened after the `AddCosmosDB` call because we were registering the default implementation with `AddSingleton` instead of `TryAddSingleton`, the tests did not catch it because the order on which they were written. The PR contains the fix plus new test coverage.

Closes #741
Closes #743

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs
@@ -7,6 +7,7 @@ using Microsoft.Azure.WebJobs.Extensions.CosmosDB;
 using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace Microsoft.Extensions.Hosting
 {
@@ -34,8 +35,8 @@ namespace Microsoft.Extensions.Hosting
                 });
 
             builder.Services.AddAzureClientsCore();
-            builder.Services.AddSingleton<ICosmosDBServiceFactory, DefaultCosmosDBServiceFactory>();
-            builder.Services.AddSingleton<ICosmosDBSerializerFactory, DefaultCosmosDBSerializerFactory>();
+            builder.Services.TryAddSingleton<ICosmosDBServiceFactory, DefaultCosmosDBServiceFactory>();
+            builder.Services.TryAddSingleton<ICosmosDBSerializerFactory, DefaultCosmosDBSerializerFactory>();
             
             return builder;
         }
--- a/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Extensions.CosmosDB;
+using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
@@ -30,11 +31,12 @@ namespace Microsoft.Extensions.Hosting
                 {
                     IConfigurationSection section = config.GetSection(path);
                     section.Bind(options);
-                });                
+                });
 
+            builder.Services.AddAzureClientsCore();
             builder.Services.AddSingleton<ICosmosDBServiceFactory, DefaultCosmosDBServiceFactory>();
             builder.Services.AddSingleton<ICosmosDBSerializerFactory, DefaultCosmosDBSerializerFactory>();
-
+            
             return builder;
         }
 
--- a/src/ExtensionsSample/Program.cs
+++ b/src/ExtensionsSample/Program.cs
@@ -4,7 +4,6 @@
 using System.IO;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs;
-using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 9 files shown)


src/WebJobs.Extensions.CosmosDB/Config/CosmosDBWebJobsBuilderExtensions.cs
```diff
@@ -30,11 +31,12 @@ public static IWebJobsBuilder AddCosmosDB(this IWebJobsBuilder builder)
                 {
                     IConfigurationSection section = config.GetSection(path);
                     section.Bind(options);
-                });                
+                });
 
+            builder.Services.AddAzureClientsCore();
```
brettsam, line 37: Do you know what services this adds that are required for AAD support? I'm wondering if it can be scoped down, or if adding all the services is okay...

ealsur: The other Azure extensions call the same method: 
* https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs#L91-L93
* https://github.com/Azure/azure-sdk-for-net/blob/6c2d70003ba77ef2b82d661cf70c641d1fa62338/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues/src/StorageQueuesWebJobsBuilderExtensions.cs
* https://github.com/Azure/azure-sdk-for-net/blob/f52c5e7fd7ca0f631a3e069dafb7ff0dd47affce/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubWebJobsBuilderExtensions.cs#L85

brettsam (approved):

PR #745, opened 2021-09-20
Adding storagev12 ScheduleMonitor implementation

karshinlin:
- New package to hold the storage-back implementation of ScheduleMonitor for Timers. 
- Removes dependency on WebJobs.Host.Storage in WebJobs.Extensions in lieu of dependency from WebJobs.Extensions.Timer.Storage to WebJobs.Host.Storage v5.0.0-beta.1


build/common.props
```diff
@@ -2,12 +2,13 @@
...
     <!-- Extensions can have independent versions and only increment when released -->
     <Version>3.0.0$(VersionSuffix)</Version>
-    <ExtensionsVersion>4.0.4$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
-    <CosmosDBVersion>4.0.0$(VersionSuffix)</CosmosDBVersion>
-    <HttpVersion>3.1.1$(VersionSuffix)</HttpVersion>
-    <MobileAppsVersion>3.0.0$(VersionSuffix)</MobileAppsVersion>
-    <SendGridVersion>3.0.2$(VersionSuffix)</SendGridVersion>
-    <TwilioVersion>3.0.1$(VersionSuffix)</TwilioVersion>
+    <ExtensionsVersion>5.0.0-beta.1$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
+    <CosmosDBVersion>4.0.1$(VersionSuffix)</CosmosDBVersion>
```
mathewc, line 6: I don't think we need to rev the versions of all the other packages - they're not changing at all and don't have a proj dependency on Extensions, right?

src/WebJobs.Extensions/Extensions/ExtensionsWebJobsStartup.cs
```diff
@@ -13,7 +13,6 @@ public class ExtensionsWebJobsStartup : IWebJobsStartup
     {
         public void Configure(IWebJobsBuilder builder)
         {
-            builder.AddTimers();
```
mathewc: Timers is still part of the default extensions package, and should register its non-storage specific functionality. E.g. for someone running a TimerTrigger in a WebJob on a single instance, they won't have to pull in Timers.Storage and can use the file based implementation of ScheduleMonitor.


Previously to get storage based schedule tracking and distributed locking, a user had to call `builder.AddTimers()` and `builder.AddAzureStorageCoreServices()`. With your new Timers.Storage package, I'd expect you'd have a `builder.AddTimersStorage()` or something, and that guy could internally add the StorageScheduleMonitor, and could also call `builder.AddAzureStorageCoreServices()`.

src/WebJobs.Extensions/Properties/AssemblyInfo.cs
```diff
@@ -3,4 +3,5 @@
 
 using System.Runtime.CompilerServices;
 
-[assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.Extensions.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
\ No newline at end of file
+[assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.Extensions.Timers.Storage, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
mathewc: We don't want internals visible between product assemblies, only for tests

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.20.1" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.27" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32-dev637683692850667066" />
```
mathewc: The referencing of these dev packages is temporary just to get a PR out right? We can't check in until we have actual published packages.

src/WebJobs.Extensions.Timers.Storage/StorageScheduleMonitor.cs
```diff
@@ -18,30 +19,21 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
     /// <summary>
     /// <see cref="ScheduleMonitor"/> that stores schedule information in blob storage.
     /// </summary>
-    public class StorageScheduleMonitor : ScheduleMonitor
+    internal class StorageScheduleMonitor : ScheduleMonitor
```
mathewc: Why the change to internal? Was public before and that was fine.

src/WebJobs.Extensions.Timers.Storage/WebJobs.Extensions.Timers.Storage.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <AssemblyName>Microsoft.Azure.WebJobs.Extensions.Timers.Storage</AssemblyName>
+    <RootNamespace>Microsoft.Azure.WebJobs.Extensions.Timers.Storage</RootNamespace>
+    <PackageId>Microsoft.Azure.WebJobs.Extensions.Timers.Storage</PackageId>
+    <Description>This package contains storage-based implementations for Timers. For more information, please visit https://azure.microsoft.com/en-us/documentation/articles/websites-webjobs-resources.</Description>
```
mathewc: nit: "... contains Azure Storage based service implementations for Timers"

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/StorageScheduleMonitorTests.cs
```diff
@@ -25,26 +29,26 @@ public StorageScheduleMonitorTests()
         {
             _scheduleMonitor = CreateScheduleMonitor(TestHostId);
 
-            Cleanup();
+            Cleanup().GetAwaiter().GetResult();
```
mathewc, line 32: Why this change?

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/StorageScheduleMonitorTests.cs
```diff
@@ -25,26 +29,26 @@ public StorageScheduleMonitorTests()
...
+            Assert.Equal(expectedPath, path);
         }
 
         [Fact]
         public void TimerStatusDirectory_HostIdNull_Throws()
         {
             StorageScheduleMonitor localScheduleMonitor = CreateScheduleMonitor(null);
 
-            CloudBlobDirectory directory = null;
+            string path = null;
```
mathewc: nit: fix name of test

test/WebJobs.Extensions.Tests/PublicSurfaceTests.cs
```diff
@@ -26,13 +26,11 @@ public void ExtensionsPublicSurface_LimitedToSpecificTypes()
                 "FilesOptions",
                 "FilesWebJobsBuilderExtensions",
                 "FileSystemScheduleMonitor",
-                "StorageScheduleMonitor",
                 "FileTriggerAttribute",
                 "IFileProcessorFactory",
                 "ScheduleMonitor",
                 "ScheduleStatus",
                 "TimerInfo",
-                "TimerWebJobsBuilderExtensions",
```
mathewc: This should remain as noted above. For your new assembly, you'll want to add a new set of public surface tests

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
karshinlin: correct - temporary stop-gap to get the pipelines moving.

src/WebJobs.Extensions/Extensions/ExtensionsWebJobsStartup.cs
karshinlin: Added these changes. We have a new builder method in the Timers.Storage project for builder.AddTimersStorage() to add the blob storage backed schedule monitor. 

One other note is that FileSystemScheduleMonitor doesn't seem to work in the Host - DI defaults to the FileSystemScheduleMonitor(string currentDirectory) ctor instead of the parameterless ctor. Do we need a change here? The Host will use the builder.AddTimersStorage() but consumers of WebJobs.Extensions that choose to use File based schedule monitor may fail here.

build/common.props
```diff
@@ -2,12 +2,13 @@
   <PropertyGroup>
     <!-- Extensions can have independent versions and only increment when released -->
     <Version>3.0.0$(VersionSuffix)</Version>
-    <ExtensionsVersion>4.0.4$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
-    <CosmosDBVersion>4.0.0$(VersionSuffix)</CosmosDBVersion>
-    <HttpVersion>3.1.1$(VersionSuffix)</HttpVersion>
-    <MobileAppsVersion>3.0.0$(VersionSuffix)</MobileAppsVersion>
-    <SendGridVersion>3.0.2$(VersionSuffix)</SendGridVersion>
-    <TwilioVersion>3.0.1$(VersionSuffix)</TwilioVersion>
+    <ExtensionsVersion>5.0.0-beta.1$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
```
mathewc, line 5: Need to review these one by one. Many of them don't need to be incremented because they've already been since the last release. We don't want to unnecessarily skip released versions.

src/WebJobs.Extensions.Timers.Storage/StorageScheduleMonitor.cs
```diff
@@ -136,12 +145,17 @@ public override async Task UpdateStatusAsync(string timerName, ScheduleStatus st
             }
         }
 
-        private CloudBlockBlob GetStatusBlobReference(string timerName)
+        private async Task<BlobClient> GetStatusBlobReference(string timerName, bool createContainerIfNotExists = false)
```
mathewc: Rename to GetStatusBlobClient

src/WebJobs.Extensions.Timers.Storage/TimersStorageWebJobsBuilderExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// </summary>
+    public static class TimersStorageWebJobsBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the Timers Azure storage backed implementation to the provided <see cref="IWebJobsBuilder"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebJobsBuilder"/> to configure.</param>
+        /// <remarks>
+        /// Note: Certain Azure services must be registered by the consumer to use the storage-backed implementation.
+        /// This is done in <see path="" cref="RuntimeStorageWebJobsBuilderExtensions.AddAzureStorageCoreServices(IWebJobsBuilder)"/>.
```
mathewc: Definitely I agree with you that in most user scenarios they want Storage based distributed locking for timers, so they want to add these core services. However, from the standpoint of the implementations in this package it's not required, so the "must" is a little strong I think. I'd say just remove the comment for now and we'll see if we get user confusion. Even before your changes here, it was already the case that customers had to know to call both builder.AddTimers() and builder.AddAzureStorageCoreServices().

src/WebJobs.Extensions.Timers.Storage/TimersStorageWebJobsBuilderExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Extension methods for Timers storage integration.
+    /// </summary>
+    public static class TimersStorageWebJobsBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the Timers Azure storage backed implementation to the provided <see cref="IWebJobsBuilder"/>.
```
mathewc: Here and below, let's reword this to something like "Adds Azure Storage based implementations for Timers services..." here and below

src/WebJobs.Extensions/Extensions/Timers/Config/TimerWebJobsBuilderExtensions.cs
```diff
@@ -27,8 +27,7 @@ public static IWebJobsBuilder AddTimers(this IWebJobsBuilder builder)
 
             builder.AddExtension<TimersExtensionConfigProvider>()
                 .BindOptions<TimersOptions>();
-            builder.Services.AddSingleton<ScheduleMonitor, StorageScheduleMonitor>();
-
+            builder.Services.AddSingleton<ScheduleMonitor>(serviceProvider => new FileSystemScheduleMonitor());
```
mathewc, line 30: Why can't you do AddSingleton<ScheduleMonitor, FileSystemScheduleMonitor>()?

test/WebJobs.Extensions.Http.Tests/WebJobs.Extensions.Http.Tests.csproj
```diff
@@ -24,6 +24,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\WebJobs.Extensions.Http\WebJobs.Extensions.Http.csproj" />
+    <ProjectReference Include="..\..\src\WebJobs.Extensions.Timers.Storage\WebJobs.Extensions.Timers.Storage.csproj" />
```
mathewc: Why do these tests need Timers?

test/WebJobs.Extensions.Tests.Common/WebJobs.Extensions.Tests.Common.csproj
```diff
@@ -23,6 +23,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\WebJobs.Extensions\WebJobs.Extensions.csproj" />
+    <ProjectReference Include="..\..\src\WebJobs.Extensions.Timers.Storage\WebJobs.Extensions.Timers.Storage.csproj" />
```
mathewc: Why is this needed?

test/WebJobs.Extensions.Timers.Storage.Tests/PublicSurfaceTests.cs
```diff
@@ -0,0 +1,62 @@
...
+            var expected = new[]
+            {
+                "StorageScheduleMonitor",
+                "TimersStorageWebJobsBuilderExtensions",
+            };
+
+            AssertPublicTypes(expected, assembly);
+        }
+
+        private static void AssertPublicTypes(IEnumerable<string> expected, Assembly assembly)
```
mathewc: Since we have 3 copies of the below methods now, I wonder if you can move them to TestHelpers in TestCommon, so the implementations can be shared? Should have been done previously.

mathewc: Actually it already looks like JobHostTestHelpers in TestCommon has these, but the tests aren't using them. So we can just use this and delete the copies.

src/WebJobs.Extensions/Extensions/Timers/Config/TimerWebJobsBuilderExtensions.cs
karshinlin: FileSystemScheduleMonitor has two constructors. DI will default to the constructor that takes an input string (which will throw an exception). We need to use the factory method to choose the constructor explicitly.

test/WebJobs.Extensions.Http.Tests/WebJobs.Extensions.Http.Tests.csproj
karshinlin: it's not... removing.

test/WebJobs.Extensions.Tests.Common/WebJobs.Extensions.Tests.Common.csproj
karshinlin: it's not... removing.

src/WebJobs.Extensions/Extensions/Timers/Config/TimerWebJobsBuilderExtensions.cs
mathewc: Got it

build/common.props
mathewc: Specifically, I think you only need to increment the following:

- ExtensionsVersion 4.0.4 -> 5.0.0-beta.1
- SendGrid 3.0.2 -> 3.0.3
- Twilio 3.0.1 -> 3.0.3 (I actually don't understand how the version is 3.0.1 here since we've released 3.0.2?)

Other package notes:

- CosmosDB: is still in 4.0.0 preview (last released version 4.0.0-preview2)
- Http: 3.1.1 had never been released. Latest on nuget is 3.0.12 and on myget it was 3.1.1-10859
- MobileApps: still in 3.0.0 preview (last released 3.0.0-beta8)

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.20.1" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.27" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32-dev637696509934498632" />
```
mathewc: This is just temporary to unblock the PR for CI right? Before checking this in, we'll need to have an actual published version, since public packages we release can only reference other public packages

karshinlin: Yes, I will update with the right packages before completing this PR.

mathewc (approved):

src/WebJobs.Extensions.Timers.Storage/TimersStorageWebJobsBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static IWebJobsBuilder AddTimersWithStorage(this IWebJobsBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            builder.AddTimers();
+            builder.Services.AddSingleton<ScheduleMonitor, StorageScheduleMonitor>();
```
mathewc: Rather than duplicate the code here, should call builder.AddTimersStorage() so the core logic is centralized

mathewc: Also for this new method, you didn't add an overload that takes an options config. Should probably do that.

src/WebJobs.Extensions.Http/WebJobs.Extensions.Http.csproj
```diff
@@ -11,8 +11,9 @@
     <Version>$(HttpVersion)</Version>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <WarningsAsErrors />
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
mathewc: Why are you changing these project settings?

src/WebJobs.Extensions.Timers.Storage/WebJobs.Extensions.Timers.Storage.csproj
```diff
@@ -13,8 +13,9 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <DefineConstants>TRACE;DEBUG;NETSTANDARD;NETSTANDARD2_0</DefineConstants>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <WarningsAsErrors />
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
mathewc: Same here and elsewhere.

mathewc (approved):

mathewc (approved):

PR #753, opened 2021-11-15
[Cosmos DB] Adds more Change Feed logs

ealsur:
This PR adds back the `CosmosDBTriggerHealthMonitor` leveraging the new Notification APIs to log:

* When a lease is acquired
* When a lease is released
* When changes are delivered to a Function (before user code is processed)
* When there are errors handling a lease, due to user unhandled exceptions or any Cosmos related failures


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerHealthMonitor.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                case ChangeFeedProcessorUserException userException:
+                    this.logger.LogWarning(userException.InnerException, "Lease {LeaseToken} encountered an unhandled user exception during processing.", leaseToken);
+                    this.logger.LogDebug("Lease {LeaseToken} has error diagnostics {Diagnostics}", leaseToken, userException.ChangeFeedProcessorContext.Diagnostics);
+                    break;
+                case CosmosException cosmosException when cosmosException.StatusCode == HttpStatusCode.RequestTimeout || cosmosException.StatusCode == HttpStatusCode.ServiceUnavailable:
+                    this.logger.LogWarning(cosmosException, "Lease {LeaseToken} experiencing transient connectivity issues.", leaseToken);
+                    break;
+                default:
+                    this.logger.LogError(exception, "Lease {LeaseToken} experienced an error during processing.", leaseToken);
```
kirankumarkolli: Do errors impact live site? 

Or what's the logging guidance for extensions?

ealsur: Extensions have access to logging into the customer configured platform (App Insights for example). Users configure which extensions they want to log for and which is the level per extension. We have docs that cover it here for our extension: https://docs.microsoft.com/en-us/azure/cosmos-db/sql/how-to-configure-cosmos-db-trigger#enabling-logging

ealsur: Enabling these logs helps users have insight on what is going on in their Trigger. We also had this log for V2 Change Feed Processor library (reference https://github.com/Azure/azure-webjobs-sdk-extensions/blob/cosmos/v3.x/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerHealthMonitor.cs), but were missing when using the V3 SDK until the Notifications APIs were added, which add more scenarios and more detailed error cases (telling apart user error from other kind for example)

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerHealthMonitor.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                this.logger.LogDebug("Lease {LeaseToken} has error diagnostics {Diagnostics}", leaseToken, asCosmosException.Diagnostics);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public Task OnLeaseAcquireAsync(string leaseToken)
+        {
+            this.logger.LogInformation("Lease {LeaseToken} was acquired to start processing.", leaseToken);
```
brettsam: Should this be `LogDebug` instead of info? Not sure everyone will want to see these logs every time.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerHealthMonitor.cs
```diff
@@ -0,0 +1,62 @@
...
+            this.logger = logger;
+        }
+
+        public Task OnErrorAsync(string leaseToken, Exception exception)
+        {
+            switch (exception)
+            {
+                case ChangeFeedProcessorUserException userException:
+                    this.logger.LogWarning(userException.InnerException, "Lease {LeaseToken} encountered an unhandled user exception during processing.", leaseToken);
+                    this.logger.LogDebug("Lease {LeaseToken} has error diagnostics {Diagnostics}", leaseToken, userException.ChangeFeedProcessorContext.Diagnostics);
```
brettsam: I'd suggest that you assign an EventId to give a better description for all of these new messages.

brettsam (commented):
Couple of suggestions for the logging approach... otherwise looks good.

brettsam (approved):

PR #758, opened 2021-12-21
Cosmos DB - Adding support for dynamic StartFromTime binding

### LINKED ISSUE #757: CosmosDBTrigger: The value for StartFromTime attribute cannot be specified by using biding expression
The value of StartFromTime attribute is used without being resolved thus it cannot be specified through config.
Specifying the value of  StartFromTime using biding expression (eg %StartFromTime%) with the actual value set in FunctionApp Configuration always throws an exception: ```The specified StartFromTime parameter is not in the correct format. Please use the ISO 8601 format with the UTC designator. For example: '2021-02-16T14:19:29Z'.``` (https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs#L151)

ealsur:
Adds support to `StartFromTime` can be bound to dynamic binding expressions.

Closes https://github.com/Azure/azure-webjobs-sdk-extensions/issues/757

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
@@ -136,7 +136,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             return TimeSpan.FromMilliseconds(attributeValue.Value);
         }
 
-        internal static string ResolveAttributeValue(INameResolver nameResolver, string attributeValue)
+        private static string ResolveAttributeValue(INameResolver nameResolver, string attributeValue)
         {
             return nameResolver.ResolveWholeString(attributeValue) ?? attributeValue;
         }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs
```diff
@@ -131,6 +136,11 @@ internal static TimeSpan ResolveTimeSpanFromMilliseconds(string nameOfProperty,
             return TimeSpan.FromMilliseconds(attributeValue.Value);
         }
 
+        private static string ResolveAttributeValue(INameResolver nameResolver, string attributeValue)
```
brettsam: Why did you make this static?

ealsur: Fair point, undid static change

brettsam (approved):

PR #767, opened 2022-04-07
[Cosmos DB] Add Listener logs for debugging

ealsur:


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -70,6 +71,8 @@ internal class CosmosDBTriggerListener<T> : IListener, IScaleMonitor<CosmosDBTri
             this._cosmosDBAttribute = cosmosDBAttribute;
             this._scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-CosmosDBTrigger-{_monitoredContainer.Database.Id}-{_monitoredContainer.Id}".ToLower());
             this._healthMonitor = new CosmosDBTriggerHealthMonitor(logger);
+            this._details = new Lazy<string>(() => $"prefix='{this._processorName}', monitoredContainer='{this._monitoredContainer.Id}', monitoredDatabase='{this._monitoredContainer.Database.Id}', " +
+                $"leaseContainer='{this._leaseContainer.Id}', leaseDatabase='{this._leaseContainer.Database.Id}', functionId='{this._functionId}'");
```
alrod: why do we need to use `Lazy` here?

alrod: I am wondering what is "`prefix='{this._processorName}`" ?

ealsur: We were directed to follow https://github.com/Azure/azure-sdk-for-net/commit/6583def79155b0dbfcc5a6561ede370e0b2fd41b as an example, there is no real need I guess, we can just use a normal string

ealsur: Prefix is the value of the `LeaseContainerPrefix` attribute configuration, which is as important/relevant as the other parameters.

alrod (approved):

brettsam (approved):

Issue #775, opened 2022-05-25
Update Twilio dependency


Swimburger:
One of your and our customers is running into an issue when using the Twilio binding.
They are using the `Microsoft.Azure.WebJobs.Extensions.Twilio` package, but also added the most recent version of the `Twilio` package, so it is using a more recent version than what your package is depending on.
Since the last time the Azure package was updated (1/26/2021), the `Twilio` package has updated the constructor of the `TwilioRestClient` with one new optional attribute. 
It looks like this change breaks the integration between your package and our package. I will also open an issue for the `GitHub` package to add the old signature back as a separate constructor overload.

Whether it is to fix this issue, or just to get the latest bug fixes and security patches, I still think the Azure package should update its dependency on `Twilio` to the latest version.

The version of `Twilio` you are using now is `5.6.3` and the latest is `5.75.1`.
The version of `Microsoft.Azure.WebJobs.Extensions.Twilio` is `3.0.2`.

I learned about this when a customer asked a question on StackOverflow: https://stackoverflow.com/questions/72371458/exception-binding-parameter-message/72372337#72372337

Swimburger:
I requested that Twilio adds back the old constructor signature to maintain compatibility in [this GitHub issue](https://github.com/twilio/twilio-csharp/issues/609).

Swimburger:
Should this be closed? Was this ever released?

PR #778, opened 2022-06-02
Update Twilio dependency

Swimburger:
When our customers install the Twilio WebJobs helper and install the latest version of the Twilio SDK, the helper throws an error because it was compiled against an older version of the Twilio SDK. See #775.

This PR updates the Twilio SDK to the latest version. This change is a breaking change so I bumped the major version of `<TwilioVersion>`.


Swimburger:
@microsoft-github-policy-service agree [company="default"]

Swimburger:
@microsoft-github-policy-service agree

src/WebJobs.Extensions.Twilio/WebJobs.Extensions.Twilio.csproj
```diff
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
-    <PackageReference Include="Twilio" Version="5.6.3" />
+    <PackageReference Include="Twilio" Version="6.*" />
```
fabiocav: Can you please update this to a specific version instead of a floating version here? If the intent is to support anything >6.0.0 (or 6.0.1 which is what is published), you can specify that here and that customers will be able to use a higher version in the project.

src/WebJobs.Extensions.Twilio/Config/TwilioExtensionConfigProvider.cs
```diff
@@ -83,10 +83,8 @@ internal static CreateMessageOptions CreateMessageOptions(JObject messageOptions
         {
             var options = new CreateMessageOptions(new PhoneNumber(GetValueOrDefault<string>(messageOptions, "to")))
             {
-                ProviderSid = GetValueOrDefault<string>(messageOptions, "providerSid"),
```
fabiocav, line 86: Looks like a number of new properties have been added as well:

Screenshot 2023-01-04 at 2 22 15 PM

Would it make sense for those to be supported here?

fabiocav:
Also looks like the tests fail with the updated dependency because of the old TFM. Should be fine to update those to something more recent/supported (including .NET 6)

src/WebJobs.Extensions.Twilio/WebJobs.Extensions.Twilio.csproj
Swimburger: Is that the best practice? I couldn't really find guidance on this for .NET libraries in general.

src/WebJobs.Extensions.Twilio/Config/TwilioExtensionConfigProvider.cs
Swimburger: I added them just in case.

src/WebJobs.Extensions.Twilio/WebJobs.Extensions.Twilio.csproj
Swimburger: Updated to 6.2.0

Swimburger:
I updated the test libraries and had to change the `DocumentationFile` file because it was being written to the root of my disk which it has no permissions for.
The Twilio tests pass, but some other tests don't and some tests peg my computer, maybe because I'm running it on Linux.

Swimburger:
@fabiocav I resolved your concerns. Can you give it another look?

fabiocav (approved):

fabiocav:
@kashimiz are we good to remove the AppVeyor check here?

Swimburger:
Bump. The AppVeyor looks very broken :)

Swimburger:
When will this change be released?

Swimburger:
@fabiocav, was this change ever released?

PR #787, opened 2022-07-15
Add support for custom names for values in data binding

MO2k4:
Currently it is not possible to use different names for binding than the property names. To be more flexible, i added a check if DataMemberAttribute is set and use this name instead of the property name

```cs
public class TestPocoWithDataMember
{
    [DataMember(Name = "custom_value")]
    public string CustomValue { get; set; }

    public string Location { get; set; }
}
```

This would allow the developer to bind the data against query parameter like custom_value.

### DIFF
--- a/src/WebJobs.Extensions.Http/PropertyHelper.cs
+++ b/src/WebJobs.Extensions.Http/PropertyHelper.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Serialization;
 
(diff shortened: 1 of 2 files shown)


MO2k4:
@mathewc @fabiocav Sorry for contacting you directly, but i've seen that you closed some prs in the last weeks.

Could you please take a look at my pr? If you are not the correct persons mentioned, could you please forward my pr for review? Is there anything i could do or improve in my pr to get this merged?

Thanks in advance

jviau (dismissed):

MO2k4:
Hey @jviau do you need anything from my end?

jviau:
@MO2k4 we are just discussing internally if we foresee any problems with this feature.

PR #797, opened 2022-09-06
Clarifying xml doc on ScheduleStatus

mathewc:


kshyju (approved):

Kryptos-FR:
I encountered the same issue as https://github.com/Azure/azure-functions-dotnet-worker/issues/954. The documentation of `TimerInfo` for isolated workers hasn't been updated. On top of that, the `TimerInfo` class in isolated worked doesn't have a `Schedule` property, so the workaround `schedule.GetNextOccurrence(DateTime.UtcNow);` isn't possible. How do we obtain the next occurrence of the trigger?

PR #802, opened 2022-10-04
CosmosDB support for ParameterBindingData reference type

liliankasem:
The Azure Functions team is working on enabling SDK-type bindings for out-of-proc language workers. To enable this, we are introducing a reference type that contains all the information the worker would require to hydrate the SDK-type on the worker side, instead of passing the payload through the host. You can read internal design document [here](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-benbyrd/app-service-web-apps/app-service-team-documents/functionteamdocs/design/oopworkers/sdktypebindings).

This PR adds rules and converters for `ParameterBindingData` which is the reference type mentioned above ^ 

- You can read this [PR description](https://github.com/Azure/azure-functions-dotnet-worker/pull/1406) to learn more about how this is used in the Functions Dotnet-Isolated worker.
- You can learn more about how this is used on the Functions host in this [PR description](https://github.com/Azure/azure-functions-host/pull/8815)

Related issue: https://github.com/Azure/azure-functions-dotnet-worker/issues/1034

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
@@ -4,13 +4,13 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.WebJobs.Description;
 using Microsoft.Azure.WebJobs.Extensions.CosmosDB.Bindings;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Config;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json.Linq;
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
     /// <summary>
     /// Defines the configuration options for the CosmosDB binding.
     /// </summary>
-    [Extension("CosmosDB")]
+    [Extension(Constants.WebJobsCosmosExtensionName)]
     internal class CosmosDBExtensionConfigProvider : IExtensionConfigProvider
     {
         private readonly ICosmosDBServiceFactory _cosmosDBServiceFactory;
@@ -30,10 +30,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
         private readonly ILoggerFactory _loggerFactory;
 
         public CosmosDBExtensionConfigProvider(
-            IOptions<CosmosDBOptions> options, 
-            ICosmosDBServiceFactory cosmosDBServiceFactory, 
+            IOptions<CosmosDBOptions> options,
+            ICosmosDBServiceFactory cosmosDBServiceFactory,
             ICosmosDBSerializerFactory cosmosSerializerFactory,
-            INameResolver nameResolver, 
+            INameResolver nameResolver,
             ILoggerFactory loggerFactory)
         {
             _cosmosDBServiceFactory = cosmosDBServiceFactory;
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -87,6 +89,13 @@ internal void ValidateConnection(CosmosDBAttribute attribute, Type paramType)
             }
         }
 
+        internal ParameterBindingData CreateBindingData(CosmosDBAttribute cosmosAttribute)
```
satvu: nit: CreateParameterBindingData

liliankasem:
@ealsur Hi Matias, this PR is ready for review, is there anything more that you feel might be missing? It should be a simple enough change, especially since we're only supporting input bindings. Also curious about what the release cycle is like and the earliest this could be released?

Let me know if you have any questions!

src/WebJobs.Extensions.CosmosDB/Constants.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
     public static class Constants
     {
         public const string DefaultConnectionStringName = "CosmosDB";
+        public const string WebJobsCosmosExtensionName = "CosmosDB";
```
liliankasem, line 11: Feels silly to have two strings with the same value, but didn't want to rename the connection string const. We could just have one const called "CosmosDB" but that requires making changes to 29 references that use `DefaultConnectionStringName`. I think that's best left for a different PR

ealsur (approved):

surgupta-msft (approved):

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -158,5 +168,49 @@ public override bool IsMatch(Type type, OpenTypeMatchContext context)
                 return base.IsMatch(type, context);
             }
         }
+
+        // Cannot use `new BinaryData(cosmosAttribute)` because this "may only be called on a Type for
+        // which Type.IsGenericParameter is true"; therefore we use our own reference type.
+        // Has the same properties as CosmosDBAttribute
+        private class CosmosParameterBindingDataContent
+        {
+            public CosmosParameterBindingDataContent()
```
surgupta-msft: Curious, if we need this default constructor for any specific scenarios?

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
```diff
@@ -53,12 +53,13 @@ public void Initialize(ExtensionConfigContext context)
...
             }
 
-            // Apply ValidateConnection to all on this rule. 
+            // Apply ValidateConnection to all on this rule.
             var rule = context.AddBindingRule<CosmosDBAttribute>();
             rule.AddValidator(ValidateConnection);
             rule.BindToCollector<DocumentOpenType>(typeof(CosmosDBCollectorBuilder<>), this);
 
             rule.BindToInput<CosmosClient>(new CosmosDBClientBuilder(this));
+            rule.BindToInput<ParameterBindingData>((attr) => CreateParameterBindingData(attr));
```
surgupta-msft, line 62: For Blob extension, we had changes in [equivalent file](https://github.com/surgupta-msft/azure-sdk-for-net/blob/60bc0fc45ab345c83aa4eebec17a27adb1b1b892/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Config/BlobsExtensionConfigProvider.cs#L111) for supporting collection. Do we not need similar changes in case of cosmos extension or may be I am missing something?

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
liliankasem: I don't think there is a scenario for this other than maybe for testing? Happy to remove it :)

src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs
liliankasem: We do not, cosmos doesn't have an "IsBatched" concept so we don't need to support ParameterBindingData collection - this is all handled in the worker

surgupta-msft: Nice, good to know.

PR #806, opened 2022-11-01
Target Based Scaling for CosmosDB + Separate out CosmosDBScaleMonitor from Listener

chiangvincent:
## Overview
This PR refactors the `CosmosDBTriggerListener` into two components, `CosmosDBScaleMonitor`, and `CosmosDBMetricsProvider`, and introduces new class called the `CosmosDBTargetScaler` to support target based scaling. The refactor separates CosmosDBScaleMonitor for the redesigned scale controller, which calls into `CosmosDBScaleMonitor` methods for scaling decisions. This is an effort to unify scaling logic between the scale controller and extensions SDKs, which have diverged over time. The `CosmosDBMetricsProvider` is shared by the `CosmosDBScaleMonitor` and `CosmosDBTargetScaler` classes, which use the provider class to retrieve metrics from the monitored container.

Here is a [sample PR](https://github.com/Azure/azure-sdk-for-net/pull/31367) for the Service Bus SDK for reference.

Note: this PR includes a fix for the `StatusFilePath_ThrowsWhenPathDoesNotExist` unit test for timer triggers after upgrading the test csproj to netcore3.1: 
To address the following changes in Exception logging, see:
- [Github PR](https://github.com/dotnet/coreclr/pull/25185)
- [Github PR Files](https://github.com/dotnet/coreclr/pull/25185/files#diff-44c858b5a24906b9be41cf77a46534485f23f583c613b1fc2483561e8f000599)

## TODO
- [x] create CosmosDBScaleMonitor
- [x] create CosmosDBMetricsProvider
- [x] remove duplicate logic between CosmosDBScaleMonitor, CosmosDBMetricsProvider, and CosmosDBTriggerListener
- [x] CosmosDBTriggerListener implement IScaleMonitorProvider and ITargetScalerProvider
- [x] fix unit tests
- [x] add CosmosDBTargetScaler
- [x] add unit tests for CosmosDBTargetScaler
- [x] E2E tests with cosmosdb trigger
- [ ] test with scale controller prototype

## Logs from E2E Verification With Local Function App
1. Generate a local nuget package with ComosDB changes from this PR
2. Create function app referencing local CosmosDB nuget packages
3. Publish and run the application locally with `azure-functions-host` repo 
4. Enable runtime driven scaling, target based scaling, and add `ScmHostingConfigurations.txt` to enable CosmosDB TBS
5. Create timer app that sends CosmosDb events every second to a production CosmosDB container
6. Ping `admin/host/scale/status` API using Postman, observe target worker count and local `azure-functions-host` logs emitted 

_Screenshot showing the target scaler requesting a worker count of 2, capped by the partition count which is 1 in this case_
image


src/ExtensionsSample/ExtensionsSample.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
```
ealsur, line 5: undo?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerMetrics.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 {
-    internal class CosmosDBTriggerMetrics : ScaleMetrics
+    public class CosmosDBTriggerMetrics : ScaleMetrics
```
ealsur: This should be internal, not public

test/WebJobs.Extensions.CosmosDB.Tests/WebJobs.Extensions.CosmosDB.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
ealsur, line 4: undo?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerMetrics.cs
chiangvincent: This was needed to make the `CosmosDBScaleMonitor` class public -- I added an overview in the PR description for more context here

ealsur: And this needs to be public? Who will be consuming it? Won't normal users see the type and have access to it?

chiangvincent: Yes, this needs to be public for CosmosDBScaleMonitor to be public -- see below:
image

Users would be able to see the type in this case

chiangvincent: In the scale controller redesign, we call the GetMetricsAsync and GetScaleStatus methods directly, see [here](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController?path=/src/ScaleController/Services/SiteScaleMonitorService.cs&version=GBv3.x&line=79&lineEnd=79&lineStartColumn=16&lineEndColumn=81&lineStyle=plain&_a=contents)

ealsur: But you need the interface to be public (ScaleMetrics) not the implementation. The consumer (scale controller) gets the ScaleMetrics reference, not the CosmosDBTriggerMetrics reference. Otherwise, you would not have been able to merge your linked code. So this class can remain internal.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -0,0 +1,135 @@
...
+                using (FeedIterator<ChangeFeedProcessorState> iterator = estimator.GetCurrentStateIterator())
+                {
+                    while (iterator.HasMoreResults)
+                    {
+                        FeedResponse<ChangeFeedProcessorState> response = await iterator.ReadNextAsync();
+                        partitionWorkList.AddRange(response);
+                    }
+                }
+
+                partitionCount = partitionWorkList.Count;
```
alrod, line 61: Just curios. Is it logical or physical partitions?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -0,0 +1,135 @@
...
+                remainingWork = partitionWorkList.Sum(item => item.EstimatedLag);
+            }
+            catch (Exception e) when (e is CosmosException || e is InvalidOperationException)
+            {
+                if (!TryHandleCosmosException(e))
```
alrod, line 67: Why do we swallow CosmosDb exceptions? In this case we end up with "ScaleIn". I would expect to throw the exception so this metric would not count in voting.

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Scale/FunctionsScaleManager.cs#L111

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Scale/FunctionsScaleManager.cs#L177

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTargetScaler.cs
```diff
@@ -0,0 +1,85 @@
...
+            return GetScaleResultInternal(context, metrics.RemainingWork, metrics.PartitionCount);
+        }
+
+        internal TargetScalerResult GetScaleResultInternal(TargetScalerContext context, long remainingWork, int partitionCount)
+        {
+            int concurrency;
+
+            if (!context.InstanceConcurrency.HasValue)
+            {
+                if (_cosmosDBTriggerAttribute.MaxItemsPerInvocation > 0)
```
alrod:
```suggestion
concurrency = _cosmosDBTriggerAttribute.MaxItemsPerInvocation > 0 ? _cosmosDBTriggerAttribute.MaxItemsPerInvocation : DefaultMaxItemsPerInvocation;

```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTargetScaler.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            if (concurrency <= 0)
+            {
+                _logger.LogWarning($"Concurrency value for target based scale must be greater than 0. Using default value of {DefaultMaxItemsPerInvocation} as concurrency value.");
+                concurrency = DefaultMaxItemsPerInvocation;
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(remainingWork / (decimal)concurrency);
+
+            string targetScaleMessage = $"'Target worker count for function '{_functionId}' is '{targetWorkerCount}' (MonitoredContainerId='{_monitoredContainer.Id}', MonitoredContainerDatabaseId='{_monitoredContainer.Database.Id}', RemainingWork ='{remainingWork}', Concurrency='{concurrency}').";
```
alrod: remove a single quote `"'Target`

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBListenerTests.cs
```diff
@@ -95,7 +96,7 @@ public async Task GetMetrics_ReturnsExpectedResult()
                 .Setup(m => m.ReadNextAsync(It.IsAny<CancellationToken>()))
                 .Returns(Task.FromResult(response.Object));
 
-            var metrics = await _listener.GetMetricsAsync();
+            var metrics = await ((CosmosDBScaleMonitor)_listener.GetMonitor()).GetMetricsAsync();
```
alrod: Should we create `CosmosDBMretricsProviderTest`? GetMetrics is not part in `CosmosDbListener `anymore.

Also should we add `CosmosDBScaleMonitorTest` and move some tests from here? As part of the PR we introduced `CosmosDBTargetScalerTests`.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTargetScaler.cs
```diff
@@ -0,0 +1,85 @@
...
+                concurrency = context.InstanceConcurrency.Value;
+            }
+
+            if (concurrency <= 0)
+            {
+                _logger.LogWarning($"Concurrency value for target based scale must be greater than 0. Using default value of {DefaultMaxItemsPerInvocation} as concurrency value.");
+                concurrency = DefaultMaxItemsPerInvocation;
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(remainingWork / (decimal)concurrency);
```
alrod, line 60: What if the load is in one partition? Will we overscale in this case?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -0,0 +1,135 @@
...
+                    {
+                        FeedResponse<ChangeFeedProcessorState> response = await iterator.ReadNextAsync();
+                        partitionWorkList.AddRange(response);
+                    }
+                }
+
+                partitionCount = partitionWorkList.Count;
+                remainingWork = partitionWorkList.Sum(item => item.EstimatedLag);
+            }
+            catch (Exception e) when (e is CosmosException || e is InvalidOperationException)
```
alrod, line 64: We want to handle all `CosmosExceptions` but throw `InvalidOperationException` ?

```suggestion
            catch (Exception e) when (e is CosmosException)
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerMetrics.cs
chiangvincent: Copying notes from offline conversation with Matias: 

Ahh got it, I see the concern. From the scale controller redesign perspective, we do not need access to specific attributes in the CosmosDBTriggerMetrics class, we just need to be able to call GetScaleStatus and GetMetricsAsync. If we are able to rely on the generic implementations of these methods (for example [here](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/806/files#diff-87779aaf26d7ac619b8e357dd921c23dbd1ab7b725ac337a30b6fb26e8072046:~:text=%7D-,async%20Task%3CScaleMetrics%3E%20IScaleMonitor.GetMetricsAsync(),-%7B)), then we do not have to make the metrics class public. This would also entail changing the implemented class for ScaleMonitor from IScaleMonitor<CosmosDBTriggerMetrics> to IScaleMonitor<ScaleMetrics> https://github.com/Azure/azure-webjobs-sdk-extensions/pull/806/files#diff-87779aaf26d7ac619b8e357dd921c23dbd1ab7b725ac337a30b6fb26e8072046:~:text=class%20CosmosDBScaleMonitor%20%3A-,IScaleMonitor%3CCosmosDBTriggerMetrics%3E,-%7B

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTargetScaler.cs
chiangvincent: In line 71, we check if the targetWorkerCount is greater than partitionCount, if it is, then we cap the targetWorkerCount at the partitionCount

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
chiangvincent: I believe this corresponds to physical partitions, but deferring to @ealsur

ealsur: Physical

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTargetScaler.cs
chiangvincent: removed

test/WebJobs.Extensions.CosmosDB.Tests/Trigger/CosmosDBListenerTests.cs
chiangvincent: Yeah I think we should create a new class for scale monitor

chiangvincent: Ported over the tests to CosmosDBScaleMonitorTests

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
chiangvincent: From my understanding, we only want to rethrow InvalidOperationException if the error trace is empty (see TryHandleCosmosException). We could be missing context on this, but I see specific unit tests that test InvalidOperationException is thrown with a specific error message. We should probably clarify with the cosmos team in case there's more context around this. @ealsur any advice here?

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.31.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.35" />
```
alrod: please update to 3.0.36

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBScaleMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+        private readonly Container _monitoredContainer;
+        private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
+        private readonly CosmosDBMetricsProvider _cosmosDBMetricsProvider;
+
+        public CosmosDBScaleMonitor(string functionId, ILogger logger, Container monitoredContainer, Container leaseContainer, string processorName)
+        {
+            _logger = logger;
+            _functionId = functionId;
+            _monitoredContainer = monitoredContainer;
+            _scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-CosmosDBTrigger-{_monitoredContainer.Database.Id}-{_monitoredContainer.Id}".ToLower());
```
alrod: use a new ctor taking 2 params

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
chiangvincent: done

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBScaleMonitor.cs
chiangvincent: done

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
ealsur: The SDK throws CosmosException. If the consumer of this API understands InvalidOperationException that would be a contract between this code and that consumer, I cannot chip in here because we are neither the consumer nor the authors or this code originally.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerMetrics.cs
ealsur: Why were these made public again? I thought the interfaces/base types provided enough access, these do not need to be public

ealsur: You just need to work with `ScaleMetrics` or `ITargetScaler` or `IScaleMonitor`, you don't need `CosmosDBMetrics`.

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.31.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.36-11991" />
```
chiangvincent: Need to remove this when the webjobs SDK is released

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerMetrics.cs
chiangvincent: Created a scaler factory class which returns a generic `IScaleMonitor` implementation

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -0,0 +1,135 @@
...
+        private bool TryHandleCosmosException(Exception exception)
+        {
+            string errormsg = null;
+            string exceptionMessage = exception.Message;
+
+            if (!string.IsNullOrEmpty(exceptionMessage))
+            {
+                foreach (KeyValuePair<string, string> exceptionString in KnownDocumentClientErrors)
+                {
+                    if (exceptionMessage.IndexOf(exceptionString.Key, StringComparison.OrdinalIgnoreCase) >= 0)
```
ealsur, line 118: Not a blocking concern but one idea could be to use Status Codes instead of messages. CosmosException contains a StatusCode for each failure, which is probably better in the future if the service decides to change error messages

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -0,0 +1,135 @@
...
+        private static readonly Dictionary<string, string> KnownDocumentClientErrors = new Dictionary<string, string>()
+        {
+            { "Resource Not Found", "Please check that the CosmosDB container and leases container exist and are listed correctly in Functions config files." },
+            { "The input authorization token can't serve the request", string.Empty },
+            { "The MAC signature found in the HTTP request is not the same", string.Empty },
+            { "Service is currently unavailable.", string.Empty },
+            { "Entity with the specified id does not exist in the system.", string.Empty },
+            { "Subscription owning the database account is disabled.", string.Empty },
+            { "Request rate is large", string.Empty },
+            { "PartitionKey value must be supplied for this operation.", "We do not support lease containers with partitions at this time. Please create a new lease collection without partitions." },
```
ealsur: nit: This error does no longer happen on the new extension

ealsur (approved):
Looks good on the trigger changes

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
chiangvincent: Ah I can't see the specific lines you are highlighting, is it just `{ "PartitionKey value must be supplied for this operation.", "We do not support lease containers with partitions at this time. Please create a new lease collection without partitions." },`? If so, I will remove from dictionary

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
chiangvincent: Thanks for the callout! It might be better to separate this part into a different PR as I do not have context on the code here (was copy/paste from existing scale monitor), just to keep changes separate from TBS

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
ealsur: Yes, just that one.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
ealsur: Completely agree, I was just writing it for our future selves

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
chiangvincent: Got it, removed

ealsur (approved):

chiangvincent:
@ealsur are the scale monitors used for Cosmos V3 the same ones here? If not, can you send a pointer to the existing listener + scale monitor we will have to cherry-pick to?

src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
chiangvincent: done

brettsam (approved):

PR #825, opened 2023-02-15
Update default ScheduleStatus values to include timezone indicator

liliankasem:
Related https://github.com/Azure/azure-functions-host/issues/7391

We're getting a couple CRIs where customers are not able to parse the schedule data because a the timestamp indicator is missing. 

Currently, when we get the schedule information it looks like this:

```json
{
	"Schedule": {
		"AdjustForDST": true
	},
	"ScheduleStatus": {
		"Last": "0001-01-01T00:00:00",
		"Next": "2023-02-14T16:50:00-08:00",
		"LastUpdated": "0001-01-01T00:00:00"
	},
	"IsPastDue": false
}
```

With java and custom functions, this leads to the following exception:

```log
Microsoft.Azure.WebJobs.Host.FunctionInvocationException : Exception while executing function: Functions.az3125-func ---> Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException : Result: Failure
Exception: IllegalArgumentException: No time zone indicator
Stack: com.google.gson.JsonSyntaxException: Failed parsing '0001-01-01T00:00:00' as Date; at path $.ScheduleStatus.
	at com.google.gson.internal.bind.DateTypeAdapter.deserializeToDate(DateTypeAdapter.java:90)
	at com.google.gson.internal.bind.DateTypeAdapter.read(DateTypeAdapter.java:75)
	at com.google.gson.internal.bind.DateTypeAdapter.read(DateTypeAdapter.java:46)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:41)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:187)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:145)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:130)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:221)
```

We can use the Utc or Local kind to give values that include timezone information. Open to better suggestions on how to approach this.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 ScheduleStatus = new ScheduleStatus
                 {
                     Last = default(DateTime).ToLocalTime(),
-                    Next = _schedule.GetNextOccurrence(now),
+                    Next = _schedule.GetNextOccurrence(now)
                 };
             }
 
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -110,14 +110,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
 
             if (ScheduleStatus == null)
             {
-                var defaultDateTime = new DateTime(1, 1, 1, 0, 0, 0, DateTimeKind.Local);
-
                 // no schedule status has been stored yet, so initialize
                 ScheduleStatus = new ScheduleStatus
                 {
-                    Last = defaultDateTime,
+                    Last = default(DateTime).ToLocalTime(),
                     Next = _schedule.GetNextOccurrence(now),
-                    LastUpdated = defaultDateTime
                 };
             }
 
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -114,7 +114,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 ScheduleStatus = new ScheduleStatus
                 {
                     Last = default(DateTime).ToLocalTime(),
-                    Next = _schedule.GetNextOccurrence(now)
+                    Next = _schedule.GetNextOccurrence(now),
+                    LastUpdated = default(DateTime).ToLocalTime()
                 };
             }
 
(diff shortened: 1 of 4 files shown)


brettsam:
Only other thing I can think of is to make it `default(DateTime).ToLocalTime()` so it's not so verbose.

But we do have scenarios where we rely on this being default(DateTime) -- 

https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs#L72-L85

Do we need to adjust those checks to match the new default? I would assume so -- maybe we can store this in a static so we can compare?

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -110,15 +110,18 @@ public async Task StartAsync(CancellationToken cancellationToken)
 
             if (ScheduleStatus == null)
             {
+                var defaultDateTime = new DateTime(1, 1, 1, 0, 0, 0, DateTimeKind.Local);
```
kshyju: `DateTime.MinValue` ?

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -110,15 +110,18 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
+
                 // no schedule status has been stored yet, so initialize
                 ScheduleStatus = new ScheduleStatus
                 {
-                    Last = default(DateTime),
-                    Next = _schedule.GetNextOccurrence(now)
+                    Last = defaultDateTime,
+                    Next = _schedule.GetNextOccurrence(now),
+                    LastUpdated = defaultDateTime
```
kshyju: Would it be safe to make this change? I wonder some host code is relying on this property to be null to execute certain code path and we will break it with this change.

liliankasem: I will need to double check this, somewhere LastUpdated is being set (to a the same value as Last). If that is in the host and expected to be null, I can remove this change and we will just need to fix Last

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
liliankasem: Using  default(DateTime).ToLocalTime()

liliankasem:
> 

Yup that makes sense! I can adjust to make sure we update those checks.

liliankasem:
@mathewc this wouldn't be a breaking change anywhere do you think?

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
liliankasem: I tested this E2E with the host and the host does not expect them to be null, doesn't make changes to the value. Without this change the kind is **unspecified** for LastUpdated. I'm not sure where else LastUpdated is being set but so far looks like this is a fine change to make

mathewc (approved):

PR #838, opened 2023-03-16
Removing AddTimersStorage extension method

### LINKED ISSUE #833: Remove .AddTimersWithStorage() from beta ScheduleMonitor configuration
We think that this convenience wrapper is confusing in its interaction with some of the other configuration options and just in terms of general naming. Instead, we should have users to call both `.AddTimers()` and `.AddTimersStorage()`.

As part of this change we want to update the XML docs to make sure that it is clear in all cases which options are being used for the distributed lock manager and ScheduleMonitor in each configuration setup.

fabiocav:
Resolves #833

### DIFF
--- a/src/WebJobs.Extensions.Timers.Storage/TimersStorageWebJobsBuilderExtensions.cs
+++ b/src/WebJobs.Extensions.Timers.Storage/TimersStorageWebJobsBuilderExtensions.cs
@@ -16,6 +16,12 @@ namespace Microsoft.Extensions.Hosting
         /// <summary>
         /// Adds Azure Storage based implementations for Timers services to the provided <see cref="IWebJobsBuilder"/>.
         /// </summary>
+        /// <remarks>
+        /// The following services are registered with the <see cref="IWebJobsBuilder"/>:
+        /// <list type="bullet">
+        ///     <item><description>An Azure Storage based implementation of a <see cref="ScheduleMonitor"/>.</description></item>
+        /// </list>
+        /// </remarks>
         /// <param name="builder">The <see cref="IWebJobsBuilder"/> to configure.</param>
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.Timers.Storage/TimersStorageWebJobsBuilderExtensions.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Extensions.Hosting
     public static class TimersStorageWebJobsBuilderExtensions
     {
         /// <summary>
-        /// Adds Azure Storage based implementations for Timers services to the provided <see cref="IWebJobsBuilder"/>.
+        /// Adds an Azure Storage based implementation of the <see cref="ScheduleMonitor"/> for timer services to the provided <see cref="IWebJobsBuilder"/>.
```
mathewc: The doc was intentionally general here since the intent for this method isn't to register just the single ScheduleMonitor service, but to register any other storage based services we might have in the future. If/when we add additional storage services in the future, I don't think we'll want to keep listing each of the services in this xml doc?

If your goal is to explicitly call out which services are being added, perhaps leave the general comment as before and list services in the remarks section?

fabiocav: This pattern is what we follow in the worker and in some ASP.NET APIs. For this particular one, the generic description should be sufficient, but I modified this based on the comment in the linked issue.

fabiocav: Updated.

mathewc (approved):

PR #842, opened 2023-04-11
order HTTP routes in a logical way

briandunnington:
Consider two routes:

/api/recent
/api/{id}

Currently, Azure Functions orders routes by the alphabetical order of the function name, so it is non-deterministic as to which route will be choosen if a request for '/api/recent' comes in depending on the names of the functions.

This PR makes the order deterministic and sensible. It prefers static segments over parameterized segments (will always pick '/api/recent' for a request of '/api/recent' even though '/api/{id}' is also a match) and prefers more-specific (longer) routes over less-specific (shorter) routes if everything else is the same.

Addresses:
https://github.com/Azure/azure-functions-host/issues/3153
https://github.com/Azure/azure-functions-core-tools/issues/2442
https://github.com/Azure/azure-functions-host/issues/3152
https://github.com/MicrosoftDocs/azure-docs/issues/11755

cc @fabiocav

### DIFF
--- a/src/WebJobs.Extensions.Http/Routing/WebJobsRouteBuilder.cs
+++ b/src/WebJobs.Extensions.Http/Routing/WebJobsRouteBuilder.cs
@@ -3,9 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Http
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
     public class WebJobsRouteBuilder
     {
         private readonly IWebJobsRouteHandler _handler;
-        private readonly List<IRouter> _routes = new List<IRouter>();
+        private readonly List<Route> _routes = new List<Route>();
         private readonly IInlineConstraintResolver _constraintResolver;
         private readonly string _routePrefix;
 
@@ -75,6 +75,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
         {
             var routes = new RouteCollection();
 
+            var routePrecedence = Comparer<Route>.Create(RouteComparison);
+            _routes.Sort(routePrecedence);
+
             foreach (var route in _routes)
             {
                 routes.Add(route);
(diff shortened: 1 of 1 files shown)


karstenjMSFT:
Please consider this PR. It dramatically improves the logic of how functions handles routes. The PR has been open since April!

grantborthwick:
Any progress on this? Our team would use this - our use case is that we have multiple routes defined, but also need to add a route at the root of the cname and want the more specific routes to take precedence over a wildcard match at the root. Either this or supporting Order on the HttpTriggerAttribute would support this scenario. Otherwise, we're implementing our own custom logic to route to sub-paths in a single function at the root.

Just also found this note that states that routes are considered alphabetically by FunctionName - https://github.com/Azure/azure-functions-host/issues/3153#issuecomment-1073158123 - this should work for our scenario by prefixing the function with z_.

pregress:
Could also use this in the workers SDK...

jviau:
@briandunnington apologies for the delay, we are working through some older PRs now.

Is this PR still relevant? If so, it would be great to see some tests added to verify and demonstrate the before & after behavior changes.

jviau (commented):
I wonder if you can basically copy and adapt https://github.com/dotnet/aspnetcore/blob/main/src/Http/Routing/src/Template/RoutePrecedence.cs ?

## Code Review -- CoPilot CLI

This is generated by GitHub Copilot CLI. 

### :red_circle: Multi-part segments: only `Parts[0]` is examined
**`src/WebJobs.Extensions.Http/Routing/WebJobsRouteBuilder.cs`** — `RouteComparison` method

ASP.NET Core routing segments can have multiple parts (e.g., `prefix{id}suffix` → 3 parts). The code accesses `Parts[0]` unconditionally, ignoring all other parts. This leads to incorrect ordering for routes with complex segments.

---

### :red_circle: Catch-all parameters (`{*path}`) not handled
**`src/WebJobs.Extensions.Http/Routing/WebJobsRouteBuilder.cs`** — `RouteComparison` method

Catch-all parameters should always have the lowest precedence, but `IsCatchAll` is never checked. A route like `/api/{*rest}` could sort ahead of `/api/{id}`, breaking specificity-based ordering.

---

### :yellow_circle: Optional parameters (`{id?}`) not considered
**`src/WebJobs.Extensions.Http/Routing/WebJobsRouteBuilder.cs`** — `RouteComparison` method

`TemplatePart.IsOptional` is not checked. A route with an optional parameter should rank below one with a required parameter at the same position.

---

### :yellow_circle: No tests
The PR adds non-trivial sorting logic but includes zero unit tests. Given the number of edge cases (catch-all, multi-part, optional, constraint counts, varying lengths), this needs test coverage before merging.

briandunnington:
thanks @jviau - this PR is almost three years old so a lot has changed since then (and i appreciate you taking a look even though it is so old). that the time, i copied the ASP.NET routing logic directly from the ASP.NET source code, but will revisit this and make the updated for handling additional segments, catchall, and optional parameters - as well as provide tests for all of the cases

PR #852, opened 2023-07-04
Check for cancellation after triggering function

### LINKED ISSUE #851: Cosmos DB triggered function advances checkpoint after cancellation
Our Cosmos DB change feed triggered Function App is sometimes skipping processing of items in the source container. I believe this is due to the functions host being stopped during scaling and the change feed checkpoint still advancing.

#### Repro steps

1. Clone the sample Function App from https://github.com/MikeNicholls/change-feed-sample

2. Configure the Function App in `local.settings.json` and run it locally.

3. Create a document in Cosmos DB to trigger the function.

4. While the function is executing, initiate a graceful shutdown of the functions host, either by pressing Ctrl+C or raising a SIGINT (e.g. `killall -2 func`). Note the ID of the document that was being processed.

5. Run the Function App locally again.

#### Expected behavior

The function triggers again on the same Cosmos DB document.

#### Actual behavior

The function will not trigger again (or will move onto the next document, if there is one).

#### Known workarounds

By not passing a cancellation token into the function, it will try to continue on while the host is shutting down. Eventually it will either complete or be killed, and if killed will will eventually resume after restart once the lease expires.

I'm not sure how reliable this workaround is, I don't know if it's possible for the user function code to be terminated early while the checkpoint still advances during a forced shutdown.

#### Related information 

* Microsoft.Azure.WebJobs.Extensions.CosmosDB version 4.3.0
* Microsoft.Azure.Cosmos version 3.35.1
* Sample project: https://github.com/MikeNicholls/change-feed-sample

MikeNicholls:
Fixes #851 by throwing an exception if cancellation is requested, preventing automatic checkpointing from occurring.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
@@ -219,6 +219,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
                 ChangeFeedProcessorUserException userException = new ChangeFeedProcessorUserException(result.Exception, context);
                 await this._healthMonitor.OnErrorAsync(context.LeaseToken, userException);
             }
+            // Prevent the change feed lease from being checkpointed if cancellation was requested
+            cancellationToken.ThrowIfCancellationRequested();
         }
 
         public IScaleMonitor GetMonitor()
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
@@ -219,6 +219,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
                 ChangeFeedProcessorUserException userException = new ChangeFeedProcessorUserException(result.Exception, context);
                 await this._healthMonitor.OnErrorAsync(context.LeaseToken, userException);
             }
+
             // Prevent the change feed lease from being checkpointed if cancellation was requested
             cancellationToken.ThrowIfCancellationRequested();
         }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -219,6 +219,8 @@ private async Task ProcessChangesAsync(ChangeFeedProcessorContext context, IRead
                 ChangeFeedProcessorUserException userException = new ChangeFeedProcessorUserException(result.Exception, context);
                 await this._healthMonitor.OnErrorAsync(context.LeaseToken, userException);
             }
+            // Prevent the change feed lease from being checkpointed if cancellation was requested
+            cancellationToken.ThrowIfCancellationRequested();
```
ealsur, line 224: Can you please add tests to cover the scenario? This cancellationToken is related to the call to StopAsync, is the Function shutdown tied to that? 
Should we throw a different error so the message is clear to on the user logs?

ealsur: Why are we cancelling at this point? The user Function successfully processed the changes (`FunctionResult result = await this._executor.TryExecuteAsync(new TriggeredFunctionData() { TriggerValue = docs }, cancellationToken);` was completed), why not checkpoint? It would cause an unnecessary duplication of the batch?

ealsur: After the discussion on the linked issue, it is clear.

ealsur (changes requested):
Please add tests for the scenario if possible

MikeNicholls:
@microsoft-github-policy-service agree company="Smudge"

MikeNicholls:
@ealsur I've added an end-to-end test which covers this scenario.

ealsur (approved):

ealsur:
@brettsam Do you know how can we get the CI tests to run for external contributors that have no write permissions?

ealsur:
CI in this repo is not working for external forks, we'll try to test the fork manually

ealsur:
Manually running the test in the fork passes:

image

PR #860, opened 2023-08-11
Use JArray.Parse instead of JArray.FromObject for string values

FHajHusein:
JArray.FromObject fails if passed a string value. As a result, manual triggering (Test/Run in portal, "Execute Function Now" in VSC, etc.) fails with:

```
System.ArgumentException : Object serialized to String. JArray instance expected.
at Newtonsoft.Json.Linq.JArray.FromObject(Object o,JsonSerializer)
at Newtonsoft.Json.Linq.JArray.FromObject(Object o)
at Microsoft.Azure.WebJobs.Extensions.CosmosDB.CosmosDBTriggerValueBinder.GetValueAsync() at ...\WebJobs.Extensions.CosmosDB\Trigger\CosmosdbTriggerValueBinder.cs : 49
at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ParameterHelper.PrepareParametersAsync()
...
```
The proposed change replaces JArray.FromObject with JArray.Parse for string values.

Note: This is for the out-of-process scenario (any language but dotnet). For dotnet, manual triggering also fails, but with an exception originating from the WebJobs SDK.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosdbTriggerValueBinder.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosdbTriggerValueBinder.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
         {
             if (_isString)
             {
-                return Task.FromResult<object>(JArray.FromObject(_value).ToString(Newtonsoft.Json.Formatting.None));
+                return Task.FromResult<object>(JArray.Parse(_value as string).ToString(Newtonsoft.Json.Formatting.None));
             }
 
             if (_isJArray)
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosdbTriggerValueBinder.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosdbTriggerValueBinder.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


ealsur:
`CosmosDBEndToEnd` Tests are failing with the proposed change:

image

The scenario is:

https://github.com/Azure/azure-webjobs-sdk-extensions/blob/b12d0c9f18262ad58bef1be431475fefe4099b27/test/WebJobs.Extensions.CosmosDB.Tests/CosmosDBEndToEndTests.cs#L200-L208

This ensures that if the bound type is `string`, the Trigger works, and this test is currently passing with the existing code.

ealsur:
```
if (_isString)
{
    if (_value is string)
    {
        Task.FromResult<object>(JArray.Parse(_value as string).ToString(Newtonsoft.Json.Formatting.None));
    }

    return Task.FromResult<object>(JArray.FromObject(_value).ToString(Newtonsoft.Json.Formatting.None));
}
```

Would probably be a better fix. The binding has two components, the type in the Function definition (`IReadOnlyList<X>` or `string` for bundles) and the other is the input. The input is normally coming from the C# SDK and that scenario works with `JArray.FromObject`, but when it's from the Portal UI, it seems the type is `string`, hence your attempt to fix would fix that but is breaking bundles when the input is coming from the C# SDK.

FHajHusein:
@microsoft-github-policy-service agree
________________________________

FHajHusein:
Thanks @ealsur.

Obviously, I confused the parameter's type with the value's type.

Your fix still doesn't address the cases where the parameter type is `JArray` or `IReadOnlyList<T>`, when manually triggering the function (e.g., via the portal).

I modified my proposed fix and tested _normal_ and _manual_ triggering, when using bundles, as well as with parameter types `JArray`, `IReadOnlyList<T>`, and `string`. Would you please check?

ealsur:
Proposed changes pass End To End tests:

image

ealsur (approved):
Manually running the tests is green

PR #862, opened 2023-08-21
Remove old dependencies by adding net6 target framework

Havunen:
Remove old dependencies by adding net6 target framework and ignore WebApiCompatShim code in net6

### DIFF
--- a/src/WebJobs.Extensions.Http/Shim/HttpRequestMessageFeature.cs
+++ b/src/WebJobs.Extensions.Http/Shim/HttpRequestMessageFeature.cs
@@ -11,7 +11,7 @@ using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Http.Shim
 {
-    public class HttpRequestMessageFeature : IHttpRequestMessageFeature
+    internal sealed class HttpRequestMessageFeature : IHttpRequestMessageFeature
     {
         private readonly HttpContext _httpContext;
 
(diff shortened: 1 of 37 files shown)


src/WebJobs.Extensions.Http/Shim/HttpRequestMessageFeature.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Http.Shim
+{
+    public class HttpRequestMessageFeature : IHttpRequestMessageFeature
```
Havunen: this code is copied here from the `Microsoft.AspNetCore.Mvc.WebApiCompatShim` package: 
ref: https://github.com/aspnet/Mvc/blob/b18526cdc8d8c868e8d7f9c6c8484064215e1002/src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/HttpRequestMessage/HttpRequestMessageFeature.cs

in case somebody somehow manages to call this code path

Havunen:
It seems the tests are not passing in dev branch either...

Frulfump:
Should this be `net8.0` now instead since .NET 6 went out of support in November 2024?

Unfortunate that no maintainer has commented on this PR...

jviau:
Closing this as these projects have all been updated to net8. #957

PR #878, opened 2023-11-28
[Cosmos DB] InputBinding returns 404 Error instead of null when pk+id does not exist when using string binding

ealsur:
Mirror of https://github.com/Azure/azure-webjobs-sdk-extensions/pull/877

For cases where `string` is used as `<T>` for binding or for Bundles (they use `string`) - there was a regression compared with the V3 Extension that it would throw an exception instead of returning `null` for input bindings.


FabianMeiswinkel (approved):
LGTM

PR #884, opened 2024-02-09
[Cosmos DB] Release 4.5.0

### LINKED ISSUE #762: Cosmos DB trigger is not connecting to CosmosDB Emulator
Hi Azure Functions Team,
I have an Azure Functions App consisting of HttpTrigger Functions and CosmosDbTrigger Functions. Both kind of functions work fine and can connect against an actual CosmosDB instance. Unfortunately, when running against the CosmosDB emulator the CosmosDBTrigger Functions fail to connect against the CosmosDB emulator.
The errors that I receive are of the following form:
```
[2022-01-15T18:17:06.601Z] Error indexing method 'Functions.feed'
[2022-01-15T18:17:06.601Z] Microsoft.Azure.WebJobs.Host: Error indexing method 'Functions.feed'. Microsoft.Azure.WebJobs.Extensions.CosmosDB: Cannot create Collection Information for Article in database ##-db with lease leases in database ##-db : The SSL connection could not be established, see inner exception. System.Net.Http: The SSL connection could not be established, see inner exception. System.Private.CoreLib: The remote certificate is invalid according to the validation procedure: RemoteCertificateNameMismatch.
```

The settings of the app look like below: 
```
{
  "IsEncrypted": false,
  "Values": {
    "AzureWebJobsStorage": "DefaultEndpointsProtocol=https;AccountName=###;AccountKey=###;EndpointSuffix=core.windows.net",
    "FUNCTIONS_WORKER_RUNTIME": "node",
    "DATABASE_CONNECTION_STRING": "AccountEndpoint=https://##.###.###.###:8081/;AccountKey=<DEFAULT_PRIMARY_KEY>"
   }
}
```
Though, the remaining HttpTrigger functions can communicate with the CosmosDB emulator via the SDK's CosmosClient. For this to work `NODE_TLS_REJECT_UNAUTHORIZED=0` is set. But as the error looks like coming form the .NET functions tool itself not from any of our node application, it looks like this setting is not helping. 

The CosmosDB emulator is setup as lined out in an [answer](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/1232#issuecomment-621422833) to [#1232](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/1232) with the deviation that the local development is done on OSX, which is why the certificate import is different. And it looks like the imported certificate is ignored by .NET.

I believe a similar, if not same question was raised in [#2024](https://github.com/Azure/azure-functions-core-tools/issues/2024#issue-632209260) was closed, but not solved. At least nothing I can find there helps me in resolving the described issues.

I'm not sure what else to try and would really appreciate any help at all :)

ealsur:
This PR prepares the 4.5.0 release which includes 

1. Bumping the SDK to `3.38.1` 
2. Applying a temporary fix to Scale Controller Metrics APIs. When the Scale Controller runs and the customer Function instances are dormant, there is a chance that the monitored container splits. In this case, the Metrics API is throwing a CosmosException with StatusCode 410. The solution is to return a Metric value that would prompt the Scale Controller to at least start 1 instance, which will take care of handling the split.

Bumping to SDK to `3.38.1` includes related fixes to the Extension:

- [4220](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4220) Change Feed Processor: Fixes disposal of unused CancellationTokenSource (#4220)
- [4276](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4276) Change Feed Processor: Fixes LeaseLostException on Notifications API for Renewer (#4276)
- [4251](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4251) Emulator : Adds support for flag in connection string to ignore SSL check (#4251)

Along other general improvements: https://raw.githubusercontent.com/Azure/azure-cosmos-dotnet-v3/master/changelog.md

Closes https://github.com/Azure/azure-webjobs-sdk-extensions/issues/762

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
@@ -61,6 +61,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB.Trigger
                 partitionCount = partitionWorkList.Count;
                 remainingWork = partitionWorkList.Sum(item => item.EstimatedLag);
             }
+            catch (CosmosException cosmosException) when (cosmosException.StatusCode == HttpStatusCode.Gone)
+            {
+                // Temporary handling of split issue described in https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4285
+                // This happens if the main instance is not running, potentially using Consumption Plan
+                // In this case, we return a positive value to make the Scale Controller consider spinning at least a single instance
+                partitionCount = 1;
+                remainingWork = 1;
+            }
             catch (Exception e) when (e is CosmosException || e is InvalidOperationException)
             {
                 if (!TryHandleCosmosException(e))
--- a/build/common.props
+++ b/build/common.props
@@ -3,7 +3,7 @@
     <!-- Extensions can have independent versions and only increment when released -->
     <Version>3.0.0$(VersionSuffix)</Version>
     <ExtensionsVersion>5.0.0$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
-    <CosmosDBVersion>4.4.2$(VersionSuffix)</CosmosDBVersion>
+    <CosmosDBVersion>4.5.0$(VersionSuffix)</CosmosDBVersion>
     <HttpVersion>3.2.0$(VersionSuffix)</HttpVersion>
     <MobileAppsVersion>3.0.0$(VersionSuffix)</MobileAppsVersion>
     <SendGridVersion>3.0.3$(VersionSuffix)</SendGridVersion>
--- a/src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
+++ b/src/WebJobs.Extensions.CosmosDB/WebJobs.Extensions.CosmosDB.csproj
@@ -19,7 +19,7 @@
     <WarningsAsErrors />
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.35.4" />
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.38.1" />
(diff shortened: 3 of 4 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -61,6 +61,14 @@ public async Task<CosmosDBTriggerMetrics> GetMetricsAsync()
                 partitionCount = partitionWorkList.Count;
                 remainingWork = partitionWorkList.Sum(item => item.EstimatedLag);
             }
+            catch (CosmosException cosmosException) when (cosmosException.StatusCode == HttpStatusCode.Gone)
+            {
+                // Temporary handling of split issue described in https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4285
+                // This happens if the main instance is not running, potentially using Consumption Plan
+                // In this case, we return a positive value to make the Scale Controller consider spinning at least a single instance
+                partitionCount = 1;
```
kirankumarkolli, line 69: Clarification: Idea is that it can only happen if the processor is not up and setting to some min-threshold will spin it-up right? 

For a large collection can the processor might take sometime to get caught-up to the impacted partition?

kirankumarkolli (approved):

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
ealsur: Yes, setting it to at least 1 will force Scale Controller to spin 1 instance. That instance will hit the split exception and create the leases, then Scale Controller can decide to scale further up.

ealsur: > For a large collection can the processor might take sometime to get caught-up to the impacted partition?

It might but that's not important. Once the split is handled, then the Scale Controller Metrics will work again and produce valid metrics that can help scale up to more instances if needed

PR #888, opened 2024-04-08
Fix scaling for empty lease.

alrod:
CosmosDb trigger need at least one execution to initiate correct lease/checkpoint,

actually it works, we just missing a document in the beginning:
1. FunctionApp is deployed, no assigned workers.
2. SC handles "Non Found" - SC detect empty leases and add a new worker -> worker creates leases container and "ContinuationToken": null doc.
3. SC votes for 0 workers as Delta = 0.
4. Add a doc1 -> SC always votes for 1 worker as Delta = 1. We always have 1 assigned worker.
5. Add doc2 -> triggers executes on doc2 as we have an active worker - proper lease doc is created. The trigger executes fine from this point of time.
doc1 has never triggered.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
@@ -17,6 +17,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB.Trigger
         private readonly Container _monitoredContainer;
         private readonly Container _leaseContainer;
         private readonly string _processorName;
+        private int _assignWorkerOnNotFoundCount = 0;
+        private int _maxAssignWorkerOnNotFoundCount = 5;
+
 
         private static readonly Dictionary<string, string> KnownDocumentClientErrors = new Dictionary<string, string>()
         {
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -17,6 +17,9 @@ internal class CosmosDBMetricsProvider
         private readonly Container _monitoredContainer;
         private readonly Container _leaseContainer;
         private readonly string _processorName;
+        private int _assignWorkerOnNotFoundCount = 0;
+        private int _maxAssignWorkerOnNotFoundCount = 5;
```
ealsur:
```suggestion
        private readonly int _maxAssignWorkerOnNotFoundCount = 5;
```

And move it with the other readonly variables?

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -59,7 +62,17 @@ public async Task<CosmosDBTriggerMetrics> GetMetricsAsync()
                 }
 
                 partitionCount = partitionWorkList.Count;
-                remainingWork = partitionWorkList.Sum(item => item.EstimatedLag);
+                if (partitionCount == 0)
+                {
+                    // When partitionCount is 0, the lease exists but there have not been any cosmos db trigger executions.
```
ealsur:
```suggestion
                    // When partitionCount is 0, the lease container exists but it has not been initialized.
```

ealsur (approved):
Consider adding tests to cover the new behavior

PR #895, opened 2024-06-10
Fix pipeline

Francisco-Gamino:
This PR contains the following changes:

* Set Configuration to 'Release' when building packing by default

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -4,8 +4,14 @@
   [string]$packageSuffix = "0",
   [bool]$isLocal = $false,
   [string]$outputDirectory = (Join-Path -Path $PSScriptRoot -ChildPath "buildoutput"),
-  [bool]$pack = $false
+  [bool]$pack = $false,
+  [string]$Configuration
 )
+
+if (-not $Configuration) {
+  Write-Host "Configuration not specified, defaulting to 'Release'" -ForegroundColor Yellow
+  $Configuration = "Release"
+}
 
 if ($null -eq $buildNumber) {
   throw 'Parameter $buildNumber cannot be null or empty. Exiting script.'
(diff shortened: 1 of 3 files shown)


Build.ps1
```diff
@@ -4,9 +4,15 @@
   [string]$packageSuffix = "0",
   [bool]$isLocal = $false,
   [string]$outputDirectory = (Join-Path -Path $PSScriptRoot -ChildPath "buildoutput"),
-  [bool]$pack = $false
+  [bool]$pack = $false,
+  [string]$Configuration
```
jviau, line 8: Consider setting the default in param itself:
```suggestion
  [string]$Configuration = "Release"
```

build/pipelines/templates/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       Write-Host $testsArray -separator "`n"
 
       # building tests will automatically build the main project as well
-      .\run-tests.ps1 -tests $testsArray
+      .\run-tests.ps1 -tests $testsArray -Configuration $env:Configuration
```
jviau, line 65: You can also perform variable substitution with ADO syntax if you want:
```suggestion
      .\run-tests.ps1 -tests $testsArray -Configuration $(Configuration)
```

jviau (approved):

Build.ps1
Francisco-Gamino: Fixed.

build/pipelines/templates/ci.yml
Francisco-Gamino: Fixed -- I've updated the script to use ADO syntax.

Build.ps1
Francisco-Gamino: @jviau -  The variable `$Configuration` is set at runtime and might not exist, so the value passed to `Build.ps1` and `run-tests.ps1` might be an empty string. If this happens and `$Configuration` defaults to `'Release'`, the script will fail. To prevent this, check if `$Configuration` is null or empty, and if so, set it to `'Release'` within the script, which was the original approach.

PR #907, opened 2024-09-13
Catch and log all exception in CosmosDBMetricsProvider, 4.8.1 version upgrade

chiangvincent:
We should be catching and logging all exceptions thrown during the CosmosDBMetricsProvider

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -3,7 +3,7 @@
     <!-- Extensions can have independent versions and only increment when released -->
     <Version>3.0.0$(VersionSuffix)</Version>
     <ExtensionsVersion>5.0.0$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
-    <CosmosDBVersion>4.9.0$(VersionSuffix)</CosmosDBVersion>
+    <CosmosDBVersion>4.8.1$(VersionSuffix)</CosmosDBVersion>
     <HttpVersion>3.2.0$(VersionSuffix)</HttpVersion>
     <MobileAppsVersion>3.0.0$(VersionSuffix)</MobileAppsVersion>
     <SendGridVersion>3.0.3$(VersionSuffix)</SendGridVersion>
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -96,6 +96,10 @@ public async Task<CosmosDBTriggerMetrics> GetMetricsAsync()
 
                 _logger.LogWarning(Events.OnScaling, errormsg);
             }
+            catch (Exception e)
+            {
+                _logger.LogWarning(Events.OnScaling, "Exception occurred while obtaining metrics for CosmosDB {0}: {1}.", e.GetType().ToString(), e.Message);
```
ealsur: Log the exception as is, not just the message?

ealsur:
```suggestion
                _logger.LogWarning(Events.OnScaling, "Exception occurred while obtaining metrics for CosmosDB {0}: {1}.", e.GetType().ToString(), e);
```

ealsur: Given it's an unknown you probably want to log as much as you can

ealsur (approved):

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
chiangvincent: changed to log e.tostring()!

build/common.props
```diff
@@ -3,7 +3,7 @@
     <!-- Extensions can have independent versions and only increment when released -->
     <Version>3.0.0$(VersionSuffix)</Version>
     <ExtensionsVersion>5.0.0$(VersionSuffix)</ExtensionsVersion> <!-- WebJobs.Extensions -->
-    <CosmosDBVersion>4.8.0$(VersionSuffix)</CosmosDBVersion>
+    <CosmosDBVersion>4.9.0$(VersionSuffix)</CosmosDBVersion>
```
ealsur:
```suggestion
    <CosmosDBVersion>4.8.1$(VersionSuffix)</CosmosDBVersion>
```

Use a patch version. Minor version is used when dependencies or APIs change

PR #908, opened 2024-09-13
[TimerTrigger] fixing daylight savings bug

### LINKED ISSUE #880: TimerTrigger issue with Daylight Saving Time change
Daylight saving time change caused the timer trigger not to trigger for 1hour for below repro app.

#### Repro app details -

Function App : prod-mcz-balance-app
Timer trigger: SendImbalanceToFiftyoneAsync
OS: Windows
App setting 'WEBSITE_TIME_ZONE' is set to: 'Romance Standard Time'

When Daylight saving time was set 1 hour back on 29th of October the TimerTrigger did not trigger for 1 hour between 12.00 AM - 01.00 AM UTC. 

#### Related information 

https://github.com/Azure/azure-webjobs-sdk-extensions/pull/492#issuecomment-479862622

brettsam:
Fixes #880

This change refactors our handling of Daylight Savings Time (DST). Previously this was all done in `GetNextTimerInterval()`, but after researching and testing, I realized that there wasn't enough information available to correctly calculate Cron schedules during the transition into and out of DST. Instead, I've moved the DST handling into the `CronSchedule` (and `ConstantSchedule`, although this one was easier) itself. This has the added benefit that when we print out the next several occurrences at the beginning of the Timer startup, they'll now be correct. Previously, we'd generate the "next" value and then re-calculate it based on DST to start our internal timer.

I've tried to add comments with explanations of terms to the code itself but have called a few things out in comments as well.

Side note that this only touches `Cron` and `Constant` schedules, which are the only ones supported by Functions. I briefly tested the behavior of `Daily` and `Weekly` and they seem... fine? It's hard to know exactly how they should behave, but they're so little used that it wasn't worth the effort to re-think them.

Some other notes: 
- `DateTime` can *sometimes* store enough details internally to allow it to roundtrip to/from `DateTimeOffset`. I've changed our usage to start from `DateTimeOffset.Now` wherever possible to ensure we have all the data for offsets stored internally. 
- However, NCronTab internally does *not* return `DateTime`s with offset details. This means that in ambiguous times (times during "fall back" from DST where there are duplicate times in a time zone), we will always think that this is the Standard (later) offset because of how `GetUtcOffset()` works. I've added calculations for the `CronSchedule` to account for this.
    > Side note. `GetUtcOffset()` only returns the Standard time in ambiguous cases when it doesn't have enough internal information to choose the correct offset. Sometimes it does! Like if you use `dateTimeOffset.LocalDateTime`, or even `DateTime.Now` it [stores some details internally](https://github.com/dotnet/runtime/blob/f722e5c5b72000a06ce267a9bee222a8c3be4696/src/libraries/System.Private.CoreLib/src/System/DateTime.cs#L133-L137) that let it choose the correct offset in ambiguous cases.
- I had previously fixed an issue when "springing forward" to DST where we'd calculate the wrong interval. In my testing, I found some other incidents where we'd schedule a trigger for a nonexistent time (i.e 02:30 when springing forward from 01:00 -> 03:00). I've taken this into account now as well.
- For `CronSchedule`, the behavior is different when you have an "interval" or a "point-in-time" schedule as the behavior around the "ambiguous" times going from DST -> Standard are different. I followed similar behavior to this: https://github.com/HangfireIO/Cronos?tab=readme-ov-file#transition-from-summer-time-in-autumn.

A succinct example of what this fixes can be shown here in this log for when daylight savings exits:

before change:
```
The next 5 occurrences of the 'TimerTrigger' schedule (Cron: '0 0,30 * * * *') will be:
      11/03/2024 00:30:00-07:00 (11/03/2024 07:30:00Z)
      11/03/2024 01:00:00-08:00 (11/03/2024 09:00:00Z)
      11/03/2024 01:30:00-08:00 (11/03/2024 09:30:00Z)
      11/03/2024 02:00:00-08:00 (11/03/2024 10:00:00Z)
      11/03/2024 02:30:00-08:00 (11/03/2024 10:30:00Z)
```

after change:
```
The next 5 occurrences of the 'TimerTrigger' schedule (Cron: '0 0,30 * * * *') will be:
      11/03/2024 00:30:00-07:00 (11/03/2024 07:30:00Z)
      11/03/2024 01:00:00-07:00 (11/03/2024 08:00:00Z)
      11/03/2024 01:30:00-07:00 (11/03/2024 08:30:00Z)
      11/03/2024 01:00:00-08:00 (11/03/2024 09:00:00Z)
      11/03/2024 01:30:00-08:00 (11/03/2024 09:30:00Z)
```

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -94,8 +94,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
             bool isPastDue = false;
 
             // we use DateTime.Now rather than DateTime.UtcNow to allow the local machine to set the time zone. In Azure this will be
-            // UTC by default, but can be configured to use any time zone if it makes scheduling easier.
-            DateTime now = DateTime.Now;
+            // UTC by default, but can be configured to use any time zone if it makes scheduling easier.                        
+            DateTimeOffset now = DateTimeOffset.Now;
+
             Logger.ScheduleAndTimeZone(_logger, _functionLogName, _schedule, TimeZoneInfo.Local.DisplayName);
 
             if (ScheduleMonitor != null)
@@ -104,7 +105,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // If we have, invoke it immediately.
                 ScheduleStatus = await ScheduleMonitor.GetStatusAsync(_timerLookupName);
                 Logger.InitialStatus(_logger, _functionLogName, ScheduleStatus?.Last.ToString("o"), ScheduleStatus?.Next.ToString("o"), ScheduleStatus?.LastUpdated.ToString("o"));
-                TimeSpan pastDueDuration = await ScheduleMonitor.CheckPastDueAsync(_timerLookupName, now, _schedule, ScheduleStatus);
+                TimeSpan pastDueDuration = await ScheduleMonitor.CheckPastDueAsync(_timerLookupName, now.LocalDateTime, _schedule, ScheduleStatus);
                 isPastDue = pastDueDuration != TimeSpan.Zero;
             }
 
@@ -114,7 +115,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 ScheduleStatus = new ScheduleStatus
                 {
                     Last = default(DateTime).ToLocalTime(),
-                    Next = _schedule.GetNextOccurrence(now),
+                    Next = _schedule.GetNextOccurrence(now.LocalDateTime),
                     LastUpdated = default(DateTime).ToLocalTime()
                 };
             }
@@ -145,7 +146,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
             else
             {
                 // start the regular schedule
-                StartTimer(DateTime.Now);
+                StartTimer(DateTimeOffset.Now);
             }
 
             _logger.LogDebug($"Timer listener started ({_functionLogName})");
@@ -255,7 +256,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
             {
                 if (!timerStarted)
                 {
-                    StartTimer(DateTime.Now);
+                    StartTimer(DateTimeOffset.Now);
                 }
             }
         }
@@ -351,9 +352,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
             }
         }
 
-        private void StartTimer(DateTime now)
+        private void StartTimer(DateTimeOffset now)
         {
-            var nextInterval = GetNextTimerInterval(ScheduleStatus.Next, now, _schedule.AdjustForDST);
+            var nextInterval = GetNextTimerInterval(ScheduleStatus.Next, now, _schedule.AdjustForDST, _schedule.IsInterval, _logger);
             StartTimer(nextInterval);
         }
 
@@ -372,7 +373,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
         /// <param name="adjustForDST">True to adjust for daylight savings time (if crossing DST boundary), false otherwise.</param>
         /// <param name="timeZone">The time zone info to use. Will use <see cref="TimeZoneInfo.Local"/> if not supplied.</param>
         /// <returns>The next timer interval.</returns>
-        internal static TimeSpan GetNextTimerInterval(DateTime next, DateTime now, bool adjustForDST, TimeZoneInfo timeZone = null)
+        internal static TimeSpan GetNextTimerInterval(DateTime next, DateTimeOffset now, bool adjustForDST, bool isInterval, ILogger logger, TimeZoneInfo timeZone = null)
         {
             TimeSpan nextInterval;
 
@@ -381,13 +382,39 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // For calculations, we use DateTimeOffsets and TimeZoneInfo to ensure we honor time zone
                 // changes (e.g. Daylight Savings Time)
                 timeZone = timeZone ?? TimeZoneInfo.Local;
-                var nowOffset = new DateTimeOffset(now, timeZone.GetUtcOffset(now));
-                var nextOffset = new DateTimeOffset(next, timeZone.GetUtcOffset(next));
-                nextInterval = nextOffset - nowOffset;
+
+                // Note: An ambigious time is one one where a local time maps to multiple UTC times due to a time zone change. For example, 
+                // in the Pacific time zone, there are two 1:30am DateTimes when transitioning from Daylight Savings Time to Standard Time.
+                // Our 'next' calculation will never handle this correctly as it does not consider time zones.
+                bool isNowAmbiguous = timeZone.IsAmbiguousTime(now.LocalDateTime);
+                bool isNextAmbiguous = timeZone.IsAmbiguousTime(next);
+
+                // Note: If a time is ambiguous, GetUtcOffset() always returns the Standard (not Daylight Savings) offset.
+                var nextTimeZoneOffset = timeZone.GetUtcOffset(next);
+
+                // Only use this calculation if this is an interval schedule (as opposed to a point-in-time schedule).
+                if (isInterval && (isNowAmbiguous || isNextAmbiguous) && (nextTimeZoneOffset != now.Offset))
+                {
+                    // This scenario covers calculations where either of the times in our calculation are ambiguous and cross the
+                    // 'Daylight Savings Time' -> 'Standard Time' boundary.
+                    // In order to calculate the next interval correctly in this scenario, we need to ignore the time zones.
+                    nextInterval = next - now.LocalDateTime;
+
+                    DateTime ambiguousTime = isNowAmbiguous ? now.LocalDateTime : next;
+                    Logger.AmbiguousTimeAdjustment(logger, ambiguousTime, timeZone.DisplayName);
+                }
+                else
+                {
+                    // This scenario covers calculations involving all other scenarios, including where the current time is in one 
+                    // offset (e.g. Standard time) and the next occurrence is in a different offset (e.g. Daylight Savings time).
+                    var nextOffset = new DateTimeOffset(next, nextTimeZoneOffset);
+                    nextInterval = nextOffset - now;
+                }
             }
             else
             {
-                nextInterval = next - now;
+                // Ignore the offset and time zone if we're not adjusting for DST
+                nextInterval = next - now.LocalDateTime;
             }
 
             // If the interval happens to be negative (due to slow storage, for example), adjust the
--- a/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // If we have, invoke it immediately.
                 ScheduleStatus = await ScheduleMonitor.GetStatusAsync(_timerLookupName);
                 Logger.InitialStatus(_logger, _functionLogName, ScheduleStatus?.Last.ToString("o"), ScheduleStatus?.Next.ToString("o"), ScheduleStatus?.LastUpdated.ToString("o"));
-                TimeSpan pastDueDuration = await ScheduleMonitor.CheckPastDueAsync(_timerLookupName, now.LocalDateTime, _schedule, ScheduleStatus);
+                TimeSpan pastDueDuration = await ScheduleMonitor.CheckPastDueAsync(_timerLookupName, now, _schedule, ScheduleStatus);
                 isPastDue = pastDueDuration != TimeSpan.Zero;
             }
 
@@ -114,9 +114,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // no schedule status has been stored yet, so initialize
                 ScheduleStatus = new ScheduleStatus
                 {
-                    Last = default(DateTime).ToLocalTime(),
+                    Last = default(DateTime),
                     Next = _schedule.GetNextOccurrence(now.LocalDateTime),
-                    LastUpdated = default(DateTime).ToLocalTime()
+                    LastUpdated = default(DateTime)
                 };
             }
 
@@ -267,7 +267,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
         /// <param name="invocationTime">The time of the invocation, likely DateTime.Now.</param>
         /// <param name="isPastDue">True if the invocation is because the invocation is due to a past due timer.</param>
         /// <param name="runOnStartup">True if the invocation is because the timer is configured to run on startup.</param>
-        internal async Task InvokeJobFunction(DateTime invocationTime, bool isPastDue = false, bool runOnStartup = false, DateTime? originalSchedule = null)
+        internal async Task InvokeJobFunction(DateTimeOffset invocationTime, bool isPastDue = false, bool runOnStartup = false, DateTimeOffset? originalSchedule = null)
         {
             try
             {
@@ -325,7 +325,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // adjust the invocation time forward for the purposes of calculating the next occurrence.
                 // Without this, it's possible to set the 'Next' value to the same time twice in a row, 
                 // which results in duplicate triggers if the site restarts.
-                DateTime adjustedInvocationTime = invocationTime;
+                DateTimeOffset adjustedInvocationTime = invocationTime;
                 if (!isPastDue && !runOnStartup && ScheduleStatus?.Next > invocationTime)
                 {
                     adjustedInvocationTime = ScheduleStatus.Next;
@@ -335,9 +335,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
                 // consider this a schedule change when the host next starts.
                 ScheduleStatus = new ScheduleStatus
                 {
-                    Last = adjustedInvocationTime,
-                    Next = _schedule.GetNextOccurrence(adjustedInvocationTime),
-                    LastUpdated = adjustedInvocationTime
+                    Last = adjustedInvocationTime.LocalDateTime,
+                    Next = _schedule.GetNextOccurrence(adjustedInvocationTime.LocalDateTime),
+                    LastUpdated = adjustedInvocationTime.LocalDateTime
                 };
 
                 if (ScheduleMonitor != null)
@@ -354,7 +354,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Listeners
 
         private void StartTimer(DateTimeOffset now)
         {
-            var nextInterval = GetNextTimerInterval(ScheduleStatus.Next, now, _schedule.AdjustForDST, _schedule.IsInterval, _logger);
+            var nextInterval = GetNextTimerInterval(ScheduleStatus.Next, now);
             StartTimer(nextInterval);
         }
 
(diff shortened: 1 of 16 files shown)


src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +51,103 @@ internal CrontabSchedule InnerSchedule
...
+            // Scenario 1 -- When Daylight Savings begins and the clock springs forward, we need to ensure that we are 
+            //   not scheduling a job to run at an invalid time. If the next occurrence is invalid, we need to skip
+            //   it and look for the next possible valid time.
+            bool isInvalidTime = TimeZone.IsInvalidTime(DateTime.SpecifyKind(nextDateTime, DateTimeKind.Unspecified));
+
+            while (isInvalidTime)
+            {
+                nextDateTime = _cronSchedule.GetNextOccurrence(nextDateTime);
+                isInvalidTime = TimeZone.IsInvalidTime(DateTime.SpecifyKind(nextDateTime, DateTimeKind.Unspecified));
+            }
```
liliankasem: Does it matter when the next occurrence happens after the invalid occurrences are discarded?

Or is this a case of only matters if interval and doesn't matter if "point-in-time"?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +51,103 @@ internal CrontabSchedule InnerSchedule
...
+            // Begin evaluating scenarios when Daylight Savings ends and time falls back. This leads to "ambiguous" times
+            // as the clock repeats an hour. For example, times from 1:00 - 1:59 AM will occur twice in Pacific Standard Time
+            // and are therefore considered "ambiguous" for this time zone.            
+            bool isNowAmbiguous = TimeZone.IsAmbiguousTime(nowDateTime);
+            bool isNextAmbiguous = TimeZone.IsAmbiguousTime(nextDateTime);
+
+            // Because of how NCronTab constructs it's "next" DateTime, if the next time is ambiguous, it will *always*
+            // return the Standard Time offset when calling GetUtcOffset().
+            var nextOffset = TimeZone.GetUtcOffset(nextDateTime);
+            var next = new DateTimeOffset(nextDateTime, nextOffset);
```
liliankasem: What is the specific scenario if we end here? What is `next` at this point? (i.e. all the if-else statements below don't evaluate, and we go straight to `return next;`) Is this still scenario 1 described above?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +51,103 @@ internal CrontabSchedule InnerSchedule
...
+            {
+                // Scenario 4 -- Interval schedule where either point is ambiguous and we're crossing the offset boundary. For example, 
+                //   an "every 30 minute" interval in Pacific time:
+                //   - 00:30-7 -> 01:00-8 (because all ambiguous "next" use Standard time) -> adjust back 01:00-7
+                //   - 01:30-7 -> 02:00-8 -> adjust back to 01:00-8
+                var offsetDiff = nextOffset - now.Offset;
+                next = TimeZoneInfo.ConvertTime(next.Add(offsetDiff), TimeZone);
+            }
+
+            return next;
```
liliankasem: I think this all mostly makes sense to me; I'm just struggling to wrap my head around the difference in the "spring forward" vs "fall back" calculations since they're all happening in this one function i.e. we're checking for DST in both directions every time which is harder to follow for me

liliankasem:
I just want to call out that the code comments provided here are fantastic, I was dreading to look at this PR but the descriptions were concise yet detailed enough. It was really easy to follow along and understand the changes being made, and why!

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: I guess technically this is really `Scenario 0` -- where we're not in any kind of invalid or ambiguous time frame and the time that we get back from NCronTab simply needs to be created using the non-confusing TimeZone offset. I'll add comments for that...

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: Yes, I agree. Let me break them out to make this more clear. I realized that as I was reading through it again also.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -82,28 +86,72 @@ private DateTimeOffset GetNextOccurrence(DateTimeOffset now)
...
+            if (TryAdjustInvalidTime(next, out DateTimeOffset? adjusted))
+            {
+                // "Spring Forward": "next" was invalid and was adjusted forward to a valid time.
+                next = adjusted.Value;
+            }
+            else if (TryAdjustAmbiguousTime(now, next, out adjusted))
+            {
+                // "Fall Back": Either "now" or "next" was ambiguous and the time was adjusted.
+                next = adjusted.Value;
+            }
```
liliankasem: Soooo much cleaner!

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -82,28 +86,72 @@ private DateTimeOffset GetNextOccurrence(DateTimeOffset now)
...
                 nextDateTime = _cronSchedule.GetNextOccurrence(nextDateTime);
-                isInvalidTime = timeZone.IsInvalidTime(DateTime.SpecifyKind(nextDateTime, DateTimeKind.Unspecified));
+                isInvalidTime = TimeZoneInfo.Local.IsInvalidTime(nextDateTime);
+            }
+
+            adjusted = new DateTimeOffset(nextDateTime);
+            return true;
+        }
+
+        internal bool TryAdjustAmbiguousTime(DateTimeOffset now, DateTimeOffset next, out DateTimeOffset? adjusted)
```
liliankasem, line 125: Can you add a summary comment here too?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -82,28 +86,72 @@ private DateTimeOffset GetNextOccurrence(DateTimeOffset now)
...
+        {
+            adjusted = null;
+
+            TimeZoneInfo timeZone = TimeZoneInfo.Local;
+            if (!timeZone.SupportsDaylightSavingTime)
```
liliankasem: Can there be an invalid time if there is no DST? Do we need to add this check to ` TryAdjustInvalidTime` as well?Maybe it makes more sense to have this check be part of `GetNextOccurrence` before we do the "TryAdjust" bits

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -116,38 +164,40 @@ private DateTimeOffset GetNextOccurrence(DateTimeOffset now)
...
                 //   - 00:30-7 -> 01:00-8 (because all ambiguous "next" use Standard time) -> adjust back 01:00-7
                 //   - 01:30-7 -> 02:00-8 -> adjust back to 01:00-8
-                var offsetDiff = nextOffset - now.Offset;
-                next = TimeZoneInfo.ConvertTime(next.Add(offsetDiff), timeZone);
+                var offsetDiff = next.Offset - now.Offset;
+                adjusted = TimeZoneInfo.ConvertTime(next.Add(offsetDiff), timeZone);
             }
 
-            return next;
+            return adjusted != null;
```
liliankasem: nit: I'd like to see us using the `is` operator more throughout our repos for improved code readability and consistency

```suggestion
            return adjusted is not null;
```

liliankasem (approved):
lgtm, minor comments

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +61,149 @@ internal CrontabSchedule InnerSchedule
...
+        internal bool TryAdjustInvalidTime(DateTimeOffset next, out DateTimeOffset? adjusted)
+        {
+            adjusted = null;
+
+            if (TimeZoneInfo.Local.IsInvalidTime(next.DateTime))
+            {
+                // This forces the offset to be recalculated for the local time zone. It will also signal that we don't need to do the
+                // ambiguous time calculations as this cannot happen during ambiguous times.
+                adjusted = next.ToLocalTime();
+                _logger.LogDebug(new EventId(900, "DstInvalidTime"), $"The calculated next occurrence for the schedule '{_cronSchedule}' of '{next}' was determined to be an invalid time in the time zone '{TimeZoneInfo.Local.DisplayName}' due to Daylight Saving Time. The adjusted next occurrence is '{adjusted}'.");
```
cjaliaga: Would make sense to have these with `LoggerMessage.Define` (on a partial class or maybe using extension methods) and using structured logging? Not a hot path (well maybe for one hour a year) but I like having them listed with their EventId/EventName one after the other.

brettsam: Sure, I moved them to their own Logger class like others.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: I actually just refactored that as well -- this was just an optimization, but I moved it up to GetNextOccurrence now (as you suggested) so it's clearer that if there's no DaylightSavingTime, nothing else needs to happen.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +61,149 @@ internal CrontabSchedule InnerSchedule
...
+        /// <summary>
+        /// If the provided DateTime is not valid, moves forward to the next occurrence in the schedule
+        /// until a valid DateTime is found. For example, 02:30 does not exist when "springing forward"
+        /// to Daylight Saving Time in the Pacific Standard Time Zone. Uses TimeZoneInfo.Local to make
+        /// the determination.
+        /// </summary>
+        /// <param name="next">The DateTimeOffset to evaluate.</param>
+        /// <param name="adjusted">The adjusted DateTimeOffset, if "next" is invalid.</param>
+        /// <returns></returns>
+        internal bool TryAdjustInvalidTime(DateTimeOffset next, out DateTimeOffset? adjusted)
```
brettsam: I realized that this was completely unnecessary. Just calling `next.ToLocalTime()` takes care of any invalid time that may have been created. So I've removed the entire method in the next iteration.

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -94,8 +94,9 @@ public async Task StartAsync(CancellationToken cancellationToken)
             bool isPastDue = false;
 
             // we use DateTime.Now rather than DateTime.UtcNow to allow the local machine to set the time zone. In Azure this will be
-            // UTC by default, but can be configured to use any time zone if it makes scheduling easier.
-            DateTime now = DateTime.Now;
```
mathewc, line 98: nit: fix DateTime.Now comments above, updating them with DateTimeOffset

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -266,7 +267,7 @@ internal async Task HandleTimerEvent()
         /// <param name="invocationTime">The time of the invocation, likely DateTime.Now.</param>
```
mathewc: DateTimeOffset.Now

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
```diff
@@ -145,7 +146,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
             else
             {
                 // start the regular schedule
-                StartTimer(DateTime.Now);
+                StartTimer(DateTimeOffset.Now);
```
mathewc, line 149: I wonder why we've never used the "now" local captured above here?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
```diff
@@ -19,6 +18,7 @@ public abstract class TimerSchedule
         /// <summary>
         /// Gets a value indicating whether intervals between invocations should account for DST.        
         /// </summary>
+        [Obsolete("This property is obsolete and will be removed in a future version. All TimerSchedule implementations should now handle their own DST transitions.")]
```
mathewc, line 21: Does this show up in the .NET Isolated mapping interfaces?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -14,14 +16,41 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
     public class CronSchedule : TimerSchedule
     {
         private readonly CrontabSchedule _cronSchedule;
+        private readonly ILogger _logger;
+
+        internal CronSchedule(CrontabSchedule schedule, ILogger logger)
+        {
+            _cronSchedule = schedule;
+            var cron = schedule.ToString();
+            var parts = cron.Split(' ');
```
mathewc: I think you may want to split like the HasSeconds ensuring we remove empty entries. E.g. maybe add a test case like "     
          * * * * * *" to your tests and see what happens?

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +61,133 @@ internal CrontabSchedule InnerSchedule
...
         public override DateTime GetNextOccurrence(DateTime now)
         {
-            return _cronSchedule.GetNextOccurrence(now);
+            // Note: TimeZoneInfo.Local is mocked in tests
+            return GetNextOccurrence(new DateTimeOffset(now)).LocalDateTime;
+        }
+
+        private DateTimeOffset GetNextOccurrence(DateTimeOffset now)
+        {
+            // The cron library only supports DateTime and does not take TimeZones into account. We need to
```
mathewc, line 79: Looks like there's an ncrontab version 3.3.3 with bug fixes that we might consider taking. I only checked to see if there was some updated version with TimeZone support :)

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +61,133 @@ internal CrontabSchedule InnerSchedule
...
+                // "Fall Back": Either "now" or "next" was ambiguous and the time was adjusted.
+                next = adjusted.Value;
+            }
+
+            return next;
+        }
+
+        /// <summary>
+        /// Uses the 'now' and 'next' values to determine if either is ambiguous. If they are, adjusts the times
+        /// so that they are unambiguous, depending on the time of schedule it is (interval or point-in-time).
```
mathewc: "type" of schedule it is

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +61,133 @@ internal CrontabSchedule InnerSchedule
...
+            //    - Run on November 3rd at 01:30-7
+            //    - Run on November 4th at 01:30-7 (and not run at 1:30-8)
+            //    - Run on November 5th at 01:30-8
+            if (!IsInterval && isNextAmbiguous && next.Offset != now.Offset)
+            {
+                // "Fall Back" Scenario 1 -- Point-in-time schedule where the next time is ambiguous and the offsets have changed.
+                //   This means that "now" is in DST and "next" is in Standard Time. Since we never want to run an ambiguous
+                //   point-in-time schedule on Standard time, move the offset back to the Daylight offset.
+                var offsetDiff = next.Offset - now.Offset;
+                adjusted = TimeZoneInfo.ConvertTime(next.Add(offsetDiff), timeZone);
```
mathewc, line 168: I'm having a hard time following this offset adjustment logic. Can you tie it to your example above, or update the example above to explicitly give examples for these two different scenarios? Or maybe just include a "For example" here to make it clear.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
```diff
@@ -32,13 +61,133 @@ internal CrontabSchedule InnerSchedule
...
+                Logger.DstAmbiguousTimeSecondOccurrence(_logger, now, _cronSchedule.ToString(), next, TimeZoneInfo.Local.DisplayName, adjusted.Value);
+            }
+            else if (IsInterval && (isNowAmbiguous || isNextAmbiguous) && next.Offset != now.Offset)
+            {
+                // "Fall Back" Scenario 3 -- Interval schedule where either point is ambiguous and we're crossing the offset boundary. For example, 
+                //   an "every 30 minute" interval in Pacific time:
+                //   - 00:30-7 -> 01:00-8 (because all ambiguous "next" use Standard time) -> adjust back 01:00-7
+                //   - 01:30-7 -> 02:00-8 -> adjust back to 01:00-8
+                var offsetDiff = next.Offset - now.Offset;
+                adjusted = TimeZoneInfo.ConvertTime(next.Add(offsetDiff), timeZone);
```
mathewc, line 201: To be sure I follow, you're saying:

next=01:00-8
now=00:30-7
offsetDiff=8-7=1
adjusted=01:00-8 + 1 = 01:00-7

That example makes sense to me. But I don't see how you get 01:00-8 from 01:30-7. Is TimeZoneInfo.ConvertTime doing this internally for you?

src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs
brettsam: Hmm, yeah, me too. Maybe because when we first captured it and now could make the interval incorrect? I'm hesitant to touch this part at this point...

src/WebJobs.Extensions/Extensions/Timers/Scheduling/TimerSchedule.cs
brettsam: No -- looks like we never added the `Schedule` to that payload. I think we were trying to map the fewest required properties that we could?

https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Timer/src/TimerInfo.cs

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: Good callout. I tested this just now and turns out we don't explicitly need to do that b/c we're looking at this after `CrontabSchedule` has already done all of its internal normalization. But I'm adding it anyway.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: Yeah, I checked too :-)

Good point, I'll pull it in.

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: `DateTimeOffset` represents an exact point in time, UTC, and the `Offset` is giving you a Local view of that time, but it knows nothing about the Time Zone or if the time is valid, for example, because none of that information exists in the way that NCronTab creates the "next".

So, during ambiguous times, the current view we have is always of the Standard/later time. In this case, with a -8 offset.

By adding the offset difference to the time (which in this case `(-8) - (-7) == -1` is actually subtracting an hour), we're subtracting one hour from the UTC point-in-time and telling it to give us the Offset based on the local time zone of that UTC point-in-time.

Put simply, we're an hour late and we want to move it back. But just moving it back doesn't adjust the internal offset (again, b/c it doesn't know about time zones inherently). So, we use `TimeZoneInfo.ConvertTime()` to get the offset correct. This is really just for outputting logs, though -- because I believe that just subracting an hour would work for all our interval calculations. But it makes the output look odd and it would show a time that doesn't actually exist in that time zone. (`11/4/2018 00:30-8` in this case isn't a real time; it should be `11/4/2018 01:30-7`).

src/WebJobs.Extensions/Extensions/Timers/Scheduling/CronSchedule.cs
brettsam: Removed this invalid time stuff... it's all handled differently now.

mathewc (approved):

PR #923, opened 2025-01-10
Add functionality to update serializer settings to fix datetime parsing issue

surgupta-msft:
Resolves - https://github.com/Azure/azure-functions-dotnet-worker/issues/2442

Users can provide configuration for cosmosDB extension via Host.json - [docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-cosmosdb-v2?tabs=isolated-process%2Cextensionv4&pivots=programming-language-csharp#hostjson-settings)

```json
{
    "version": "2.0",
    "extensions": {
        "cosmosDB": {
            "serializerSettings": {
              "dateParseHandling":  "None"
            }
        }
    }
}
```


src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
jviau, line 1: Do we need to use `ObjectSerializer` here? If our goal is to customize Newtonsoft.Json can we accomplish that without bringing in `Azure.Core`? We could use the Newtonsoft.Json JsonSerializer directly in this class.

src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.IO;
+using Azure.Core.Serialization;
+using Microsoft.Azure.Cosmos;
+
+namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
+{
+    /// <summary>
+    /// This is a wrapper that allows us to use cosmosDB options ObjectSerializer to create a CosmosSerializer.
+    /// </summary>
+    internal sealed class ObjectCosmosSerializer : CosmosSerializer
```
jviau, line 15: I think we may want to inherit `CosmosLinqSerializer`.

jviau (changes requested):

src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
surgupta-msft: The project already depends on `Azure.Core` via [Microsoft.Azure.Cosmos](https://www.nuget.org/packages/Microsoft.Azure.Cosmos/) dependency, if that's your concern. If I have to use NewtonSoft.Json JsonSerializer directly here, I think I will need to convert between `System.IO.Stream` and `System.IO.TextReader` for every call to `ToStream()` and `FromStream()`? May be ObjectSerializer also does that, but in that case it will be better to use already available ObjectSerializer. Let me know your thoughts.

src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
surgupta-msft: `CosmosLinqSerializer` also inherits from `CosmosSerializer`. Is there any added benefit to using `CosmosLinqSerializer`?

For context, we have implemented same serializer on Worker side too - https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.CosmosDB/src/WorkerCosmosSerializer.cs

jviau: I think `CosmosLinqSerializer` is needed for the LINQ IQueryable support with Cosmos. I am not sure if it is needed, but I wonder if there are features not working/broken without it?

src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
jviau: It does already do the wrapping: https://github.com/Azure/azure-sdk-for-net/blob/7739c55dd0975c23cc20eac9d09b7d239b5310dc/sdk/core/Microsoft.Azure.Core.NewtonsoftJson/src/NewtonsoftJsonObjectSerializer.cs#L70

I have a preference for including that code directly instead of bringing in an additional package for it (the Microsoft.Azure.Core.NewtonsoftJson package)

src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
surgupta-msft: The tests ran fine. This code will only execute if host.json has `dateParseHandling` serializerSettings.

src/WebJobs.Extensions.CosmosDB/Config/ObjectCosmosSerializer.cs
surgupta-msft: For context, the package `Microsoft.Azure.Core.NewtonsoftJson` only has 2 dependencies - `Azure.Core` and `Newtonsoft.Json` - [link](https://www.nuget.org/packages/Microsoft.Azure.Core.NewtonsoftJson),  which were already part of this extension because of other dependencies. So the package is not bringing in anything new. Is there any benefit of writing our own type here with similar logic instead of using the available one. 

I think I forgot to mention, we are also using same class in worker too - https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.CosmosDB/src/WorkerCosmosSerializer.cs

jviau (approved):

src/WebJobs.Extensions.CosmosDB/Config/SerializerSettings.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
+{
+    public class SerializerSettings
+    {
+        /// <summary>
+        /// Gets or sets a string to be included in the User Agent for all operations by Cosmos DB bindings and triggers.
```
ealsur, line 9: This string does not represent the property

PR #929, opened 2025-02-19
fixing datetimeoffset minvalue implicit conversion when in time zone that is ahead of UTC

### LINKED ISSUE #928: Timer trigger listener does not work when WEBSITE_TIME_ZONE is set due to DateTimeOffset
A recent CRI found that a FunctionApp's timer trigger did not start due to the following exception:

`The UTC time represented when the offset is applied must be between year 0 and 10,000`
 
The customer mitigated the issue by removing WEBSITE_TIME_ZONE.
 
We can see that the CRON is picked up properly:

```log 
The 'name' timer is using the schedule 'Cron: '0 0 0 1 1 *'' and the local time zone: '(UTC+09:00) Osaka, Sapporo, Tokyo'
Function 'name' initial status: Last='0001-01-01T00:00:00.0000000', Next='2026-01-01T00:00:00.0000000+09:00', LastUpdated='2025-02-05T16:01:00.0016370+09:00'
```
 
However, when `WEBSITE_TIME_ZONE` is removed, we see a different result:
 
```log
The 'name' timer is using the schedule 'Cron: '0 0 0 1 1 *'' and the local time zone: '(UTC+09:00) Osaka, Sapporo, Tokyo'
Function 'name' initial status: Last='0001-01-01T09:00:00.0000000+09:00', Next='2026-01-01T00:00:00.0000000+09:00', LastUpdated='2025-02-12T17:35:00.0000000+09:00'
 ```

The difference is the value of `Last`:

- With `WEBSITE_TIME_ZONE`: Last='0001-01-01T00:00:00.0000000'
- Without `WEBSITE_TIME_ZONE`: Last='0001-01-01T09:00:00.0000000+09:00'

### So I think the exception is happening because the converted date is outside the offset min and max value because of the timezone offset.

The stack trace for the exception points to the  DateTime to DateTimeOffest change we recently made to address a daylight savings bug:
 
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/2c9483cbdb769615ff66df2b92e8792e34ea0050/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs#L48
 
Full stack:

```log
Microsoft.Azure.WebJobs.Host.Listeners.FunctionListenerException : The listener for function 'Functions.name' was unable to start. ---> System.ArgumentOutOfRangeException : The UTC time represented when the offset is applied must be between year 0 and 10,000. (Parameter 'offset')
   at System.DateTimeOffset.ValidateDate(DateTime dateTime,TimeSpan offset)
   at System.DateTimeOffset..ctor(DateTime dateTime)
   at System.DateTimeOffset.op_Implicit(DateTime dateTime)
   at async Microsoft.Azure.WebJobs.Extensions.Timers.ScheduleMonitor.CheckPastDueAsync(String timerName,DateTimeOffset now,TimerSchedule schedule,ScheduleStatus lastStatus) at D:\a\_work\1\s\src\WebJobs.Extensions\Extensions\Timers\Scheduling\ScheduleMonitor.cs : 66
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Extensions.Timers.Listeners.TimerListener.StartAsync(CancellationToken cancellationToken) at D:\a\_work\1\s\src\WebJobs.Extensions\Extensions\Timers\Listener\TimerListener.cs : 108
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Listeners.SingletonListener.StartAsync(CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Singleton\SingletonListener.cs : 70
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Listeners.FunctionListener.StartAsync(??) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Listeners\FunctionListener.cs : 68
```

 [Where the date validation happens in System.DateTime](https://github.com/dotnet/runtime/blob/cc2019bc0b25e724275dba81ab9c295879ad8cb3/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.cs#L933-L949)

### Mitigation

Remove `WEBSITE_TIME_ZONE` from app settings

brettsam:
fixes #928 

this bug was introduced with #908 as several places moved toward `DateTimeOffset`

Ultimately, this throws an exception:
```csharp
// tokyo is +9 hours from utc
using var timeZoneSetter = TimeZoneSetter.TokyoStandard;
if (default(DateTime) == default(DateTimeOffset)) // this throws due to implicit conversion
{
}
```

So we were doing a comparison between `DateTime.MinValue` and `DateTimeOffset.Now`... which performed an implicit conversion and threw an exception when that MinValue converted to an out-of-range value. This only happened when in a + time zone from UTC.

The fix is to remove that implicit conversion by only comparing to DateTimes in this situation.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
     /// </summary>
     public abstract class ScheduleMonitor
     {
-        private static DateTimeOffset defaultDateTime = default(DateTimeOffset).ToLocalTime();
+        private static DateTime defaultDateTime = new DateTime(0, DateTimeKind.Local);
 
         /// <summary>
         /// Gets the last recorded schedule status for the specified timer.
@@ -57,7 +57,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
                 DateTimeOffset nextOccurrence = schedule.GetNextOccurrence(now.LocalDateTime);
                 lastStatus = new ScheduleStatus
                 {
-                    Last = defaultDateTime.LocalDateTime,
+                    Last = defaultDateTime,
                     Next = nextOccurrence.LocalDateTime,
                     LastUpdated = now.LocalDateTime
                 };
@@ -106,7 +106,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
                         lastUpdated = now;
                     }
 
-                    lastStatus.Last = defaultDateTime.LocalDateTime;
+                    lastStatus.Last = defaultDateTime;
                     lastStatus.Next = expectedNextOccurrence.LocalDateTime;
                     lastStatus.LastUpdated = lastUpdated.LocalDateTime;
                     await UpdateStatusAsync(timerName, lastStatus);
--- a/test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/CronScheduleTests.cs
+++ b/test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/CronScheduleTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Azure.WebJobs.Extensions.Tests.Common;
+using Microsoft.Azure.WebJobs.Extensions.Tests.Extensions.Timers;
 using Microsoft.Azure.WebJobs.Extensions.Timers;
 using Xunit;
 
@@ -54,7 +55,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Tests.Timers.Scheduling
         [Fact]
         public void Interval_IntoDST_ReturnsExpectedValue()
         {
-            TimerListenerTests.SetLocalTimeZoneToPacific();
+            using var timeZoneSetter = TimeZoneSetter.PacificStandard;
             var testLogger = new TestLogger("Test");
 
             // Every hour at the 30 min mark
@@ -79,7 +80,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Tests.Timers.Scheduling
         [Fact]
         public void PointInTime_IntoDST_ReturnsExpectedValue()
         {
-            TimerListenerTests.SetLocalTimeZoneToPacific();
+            using var timeZoneSetter = TimeZoneSetter.PacificStandard;
 
             // 01:30 every day
             CronSchedule.TryCreate("0 30 1 * * *", out CronSchedule schedule);
@@ -103,7 +104,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Tests.Timers.Scheduling
         [Fact]
         public void PointInTime_WithinAmbiguousHour_ReturnsExpectedValue()
         {
-            TimerListenerTests.SetLocalTimeZoneToPacific();
+            using var timeZoneSetter = TimeZoneSetter.PacificStandard;
 
             // every 20 minutes
             CronSchedule.TryCreate("0 */20 * * * *", out CronSchedule schedule);
@@ -133,7 +134,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Tests.Timers.Scheduling
         [Fact]
         public void Interval_OutOfDST_ReturnsExpectedValue()
         {
-            TimerListenerTests.SetLocalTimeZoneToPacific();
+            using var timeZoneSetter = TimeZoneSetter.PacificStandard;
             var logger = new TestLogger("Test");
 
             // Every hour at the 30 min mark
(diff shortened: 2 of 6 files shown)


test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/ScheduleMonitorTests.cs
```diff
@@ -64,6 +62,32 @@ public async Task CheckPastDue(bool lastSet, bool lastUpdatedSet)
             Assert.Null(monitor.CurrentStatus);
         }
 
+        [Theory]
+        [InlineData(true, false)]
+        [InlineData(false, false)]
+        [InlineData(true, true)]
+        [InlineData(false, true)]
+
+        public Task CheckPastDue_PlusTimeZone(bool lastSet, bool lastUpdatedSet)
```
brettsam, line 76: this test failed before my change

src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
     /// </summary>
     public abstract class ScheduleMonitor
     {
-        private static DateTimeOffset defaultDateTime = default(DateTimeOffset).ToLocalTime();
+        private static DateTime defaultDateTime = new DateTime(0, DateTimeKind.Local);
```
brettsam, line 17: the fix... just use DateTime here since all our comparisons are DateTime below

brettsam:
here's where the exception was being thrown: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/2c9483cbdb769615ff66df2b92e8792e34ea0050/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleMonitor.cs#L74

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/ScheduleMonitorTests.cs
```diff
@@ -64,6 +62,32 @@ public async Task CheckPastDue(bool lastSet, bool lastUpdatedSet)
...
+            using var timeZoneSetter = TimeZoneSetter.TokyoStandard;
+            return CheckPastDue(lastSet, lastUpdatedSet);
+        }
+
+        [Theory]
+        [InlineData(true, false)]
+        [InlineData(false, false)]
+        [InlineData(true, true)]
+        [InlineData(false, true)]
+
```
liliankasem, line 88: there shouldnt be a new line here? or on line 70

test/WebJobs.Extensions.Tests/Extensions/Timers/TimeZoneSetter.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        var info = typeof(TimeZoneInfo).GetField("s_cachedData", BindingFlags.NonPublic | BindingFlags.Static);
+        var cachedData = info.GetValue(null);
+        var field = cachedData.GetType().GetField("_localTimeZone", BindingFlags.NonPublic | BindingFlags.Instance);
+        field.SetValue(cachedData, TimeZoneInfo.FindSystemTimeZoneById(timeZoneId));
+    }
+
+    public static TimeZoneSetter PacificStandard => new("Pacific Standard Time");
+
+    public static TimeZoneSetter TokyoStandard => new("Tokyo Standard Time");
```
liliankasem, line 24: Would it be worth adding another test case here for GMT or UTC?

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/CronScheduleTests.cs
```diff
@@ -54,7 +55,7 @@ public void GetNextOccurrence_ThreeDaySchedule_MultipleScheduleIterations()
         [Fact]
         public void Interval_IntoDST_ReturnsExpectedValue()
         {
-            TimerListenerTests.SetLocalTimeZoneToPacific();
+            using var timeZoneSetter = TimeZoneSetter.PacificStandard;
```
liliankasem, line 58: It is possible to make these theories and test them all on a -, + and 0 timezone?

liliankasem (approved):

test/WebJobs.Extensions.Tests/Extensions/Timers/TimeZoneSetter.cs
brettsam: CI runs on UTC anyway -- but I'll add that for these tests and make it more explicit.

test/WebJobs.Extensions.Tests/Extensions/Timers/Scheduling/CronScheduleTests.cs
brettsam: We could, but it'd take some work to identify the dates for some other time zones. And it may not be worth it as whether it's +/- UTC doesn't matter to this logic as much b/c they're not at the "edges" of min/max values like DateTime.Min/Max are (and we never use DateTime.Max anywhere, which is why we have no test for it).

PR #933, opened 2025-03-05
Fixing default DateTime bug with TimeZones

brettsam:
Fixes https://github.com/Azure/azure-functions-dotnet-worker/issues/2971 

There was an issue with us using a default `DateTime` that, when parsed by certain Json serializers would throw an exeption (notably `System.Text.Json` in .NET... although `Newtonsoft.Json` worked fine)

This is because using `new DateTime(0, DateTimeKind.Local)` results in a usable `DateTime`, but when cast or parsed as a `DateTimeOffset`, would throw.

This change fixes that and adds a bunch of tests around the scenarios we were seeing. It also adds an additional detail to the TriggerDetails collection, which would have helped us in our analysis.

PR #934, opened 2025-03-07
fixing issue in Timer's BindAsync with invalid dates

brettsam:
During testing of #933 I realized that portal invocations were still susceptible to "invalid" dates coming from the status blob. I had been correcting them in one place, but not all (turns out there were only two).

This change wraps the call to `GetStatusAsync()` with a `GetSafeStatusAsync()` that ensures the returned status is moved to our Default value to correct any issues with time zones. Doing it this way means that all derived `ScheduleMonitor` classes also get the safe API.

A test was added for the `BindAsync()` call where this was happening, and it failed in the Tokyo time zone before the fix.

### DIFF
--- a/src/WebJobs.Extensions/Extensions/Timers/Bindings/TimerTriggerBinding.cs
+++ b/src/WebJobs.Extensions/Extensions/Timers/Bindings/TimerTriggerBinding.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers.Bindings
                 ScheduleStatus status = null;
                 if (_attribute.UseMonitor && _scheduleMonitor != null)
                 {
-                    status = await _scheduleMonitor.GetStatusAsync(_timerName);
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions/Extensions/Timers/Bindings/TimerTriggerBinding.cs
```diff
@@ -66,7 +66,7 @@ public async Task<ITriggerData> BindAsync(object value, ValueBindingContext cont
                 ScheduleStatus status = null;
                 if (_attribute.UseMonitor && _scheduleMonitor != null)
                 {
-                    status = await _scheduleMonitor.GetStatusAsync(_timerName);
+                    status = await _scheduleMonitor.GetSafeStatusAsync(_timerName);
```
brettsam, line 69: The bug was here -- I wasn't correcting any invalid dates as I was in TimerListener.

liliankasem (approved):

PR #950, opened 2025-08-29
Refactor releases to be individual pipelines

jviau:
Refactor and split up release pipeline to be an individual pipeline per package. The previous release pipeline attempted to consume the current `official-build.yml` pipeline, but this became too complex and convoluted. Instead, the new pipelines make no use of the common CI and build & test the intended projects directly when it is queued.


mattchenderson (approved):

jviau:
@microsoft-github-policy-service rerun

PR #955, opened 2025-09-03
Rev/fix versions for packages to prepare for release

jviau:
Prepare to release the following packages:

Microsoft.Azure.WebJobs.Extensions -- `5.2.0` -> `5.2.1`. Found the current version in `Directory.Version.props` was wrong. Current released version is `5.2.0` in our _internal_ feed only. Nuget.org is old.
Microsoft.Azure.WebJobs.Extensions.Http -- `3.2.0` -> `3.2.1`
Microsoft.Azure.WebJobs.Extensions.SendGrid -- `3.1.0` -> `3.1.1`
Microsoft.Azure.WebJobs.Extensions.Timers.Storage -- no change. `1.0.0` -> `1.1.0` is already staged.
Microsoft.Azure.WebJobs.Extensions.Twilio -- `4.0.0` -> `4.0.1`. Nuget.org is old, we have 4.0.0 released in our internal feed.

### DIFF
--- a/src/WebJobs.Extensions/Directory.Version.props
+++ b/src/WebJobs.Extensions/Directory.Version.props
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>3.0.0</VersionPrefix>
+    <VersionPrefix>5.2.1</VersionPrefix>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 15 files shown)


src/WebJobs.Extensions/Directory.Version.props
```diff
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>3.0.0</VersionPrefix>
+    <VersionPrefix>5.2.1</VersionPrefix>
```
surgupta-msft, line 3: How was it working with version 3.0.0 here? During the release of this package, it should give incorrect version, right?

surgupta-msft (approved):

RohitRanjanMS (approved):

src/WebJobs.Extensions/Directory.Version.props
jviau: We never released with the incorrect version.

PR #962, opened 2025-09-23
Add Change Feed mode support to CosmosDBTriggerAttribute and related …

MoimHossain:
This pull request adds support for configuring the Change Feed mode in the Cosmos DB Trigger, allowing users to opt-in to the new "AllVersionsAndDeletes" mode (a Cosmos DB preview feature), in addition to the default "LatestVersion" mode. The implementation ensures forward and backward compatibility with different SDK versions and includes comprehensive tests for the new functionality.

**Cosmos DB Trigger: Change Feed mode support**

* Added a new `CosmosDBTriggerChangeFeedMode` enum to represent the available change feed modes: `LatestVersion` (default) and `AllVersionsAndDeletes` (preview feature in Azure Cosmos DB).
* Updated the `CosmosDBTriggerAttribute` to include a `ChangeFeedMode` property, defaulting to `LatestVersion`, with XML documentation explaining its usage and impact.

**SDK and compatibility updates**

* Upgraded the `Microsoft.Azure.Cosmos` SDK dependency from version 3.44.1 to 3.52.0 to support the new change feed modes.
* Modified the `CosmosDBTriggerListener` to use reflection to set the change feed mode on the processor builder, ensuring compatibility with different SDK versions and logging a warning if the requested mode is unavailable.

**Testing enhancements**

* Added unit tests to verify that the `ChangeFeedMode` property is correctly set and defaults as expected, and that the attribute behaves correctly when configured with `AllVersionsAndDeletes`. 
* Minor test file import adjustments to support the new enum and SDK version.

### Summary of changes implemented:

- Added `CosmosDBTriggerChangeFeedMode ` enum and `ChangeFeedMode ` property on `CosmosDBTriggerAttribute` (default `LatestVersion`).
- Implemented reflection in `CosmosDBTriggerListener` to invoke SDK `WithChangeFeedMode` if available, falling back gracefully and logging a warning if `AllVersionsAndDeletes` requested but unsupported.
- Updated tests to validate default and explicit mode selection using new enum.
- Removed direct dependency on SDK `ChangeFeedMode` objects to keep attribute arguments valid.
Bumped Cosmos SDK to 3.52.0

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttribute.cs
@@ -130,5 +130,14 @@ namespace Microsoft.Azure.WebJobs
         /// PreferredLocations = "East US,South Central US,North Europe".
         /// </example>
         public string PreferredLocations { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Change Feed mode to use when processing changes. Defaults to <see cref="CosmosDBTriggerChangeFeedMode.LatestVersion"/>.
+        /// </summary>
+        /// <remarks>
+        /// Set to <see cref="CosmosDBTriggerChangeFeedMode.AllVersionsAndDeletes"/> to receive all intermediate mutations and delete events (preview feature in Azure Cosmos DB).
+        /// When using AllVersionsAndDeletes the JSON payload for each item will include metadata describing the operation type and related system information. Use a parameter type such as <c>IReadOnlyList<JObject></c> to access these metadata fields.
+        /// </remarks>
+        public CosmosDBTriggerChangeFeedMode ChangeFeedMode { get; set; } = CosmosDBTriggerChangeFeedMode.LatestVersion;
     }
 }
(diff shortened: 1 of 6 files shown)


jviau:
@MoimHossain the feature `AllVersionsAndDeletes` is only available in Cosmos preview packages. I don't think we can expose this in a "GA" package.

What is your planned usage for this? Are you using WebJobs packages directly, or are you an Azure Functions customer? If Azure functions, what language?

MoimHossain:
@jviau a customer of us (MSFT) requires this for a mission critical project. They are using Golang on Flex Consumption and requires this feature. They are coming from AWS - and we need a parity with Cosmos + Flex Consumption for Dynamodb + Lambda. 
Would be great to have a fix for this - to unblock them.

jviau (changes requested):
Blocking this PR based on our teams discussion.

Summary:

The goal of this PR is to enable a flex consumption customer to use this preview feature. The PR in this state will not accomplish that as said customer would also need to get extensions bundle to bring in a preview version of the CosmosDB package - which they have no ability to do so.

Additionally, we cannot ship a preview dependency/feature in a GA package. We would need to ship it as a separate `-preview` version of this package. There is the dotnet `ExperimentalAttribute`, but that preview-quality signal won't be available to functions out of proc customers, which means it is insufficient for functions.

We have two options available:

1. Follow up with CosmosDB team to see when they plan to make this feature a part of the GA package. When they do so we can include the feature here.
2. Make a separate preview package, much like CosmosDB does, which will release along-side the GA package. We can leverage bundles preview feed for this. However, this will have non-negligible cold start impact as preview bundle would need to be fetched on every host startup. This may defeat the point of having a preview package for the customer.

jviau:
Closing in favor of #972

PR #966, opened 2025-10-10
Fix timers release path

jviau:


eng/ci/official-build.yml
```diff
@@ -55,5 +55,5 @@ extends:
       jobs:
       - template: /eng/ci/templates/jobs/build.yml@self
         parameters:
-          sign: true
+          sign: false
```
jviau, line 58: I changed this to `true` last week thinking it was a mistake. It was intentional to have it as `false` -- we only sign in our release builds (which independently build the specific package being released)

mathewc: In commit https://github.com/Azure/azure-webjobs-sdk-extensions/commit/c459c5509e4ecb7924b001dbddefe0a5611461af I see a test was marked as skipped, with no work item to fix the test. What's the plan there?

eng/ci/official-release.timers.yml
```diff
@@ -14,5 +14,5 @@ extends:
     public: ${{ parameters.public }}
     folder: azure-webjobs-extensions-timers
     projects: |
-      $(Build.SourcesDirectory)/src/WebJobs.Extensions.Timers/WebJobs.Extensions.Timers.csproj
-      $(Build.SourcesDirectory)/test/WebJobs.Extensions.Timers.Tests/WebJobs.Extensions.Timers.Tests.csproj
+      $(Build.SourcesDirectory)/src/WebJobs.Extensions.Timers.Storage/WebJobs.Extensions.Timers.Storage.csproj
```
mathewc, line 17: Can you explain this? If we weren't building this properly then how are we releasing the nuget packages?

mathewc (approved):

aishwaryabh (approved):

eng/ci/official-build.yml
jviau: Will follow up with a work item

eng/ci/official-release.timers.yml
jviau: We never released with this refactored CI before, this is the first one so ironing out the kinks

Issue #961, opened 2025-09-17
Cosmos DB change feed trigger should support Incremental (all versions and deletes) mode


nzthiago:
Cosmos DB change feed has now 2 modes:
* Latest version mode (the function only sees the latest changes, and does not see hard-deletes)
* [Incremental (All version and deletes) mode](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-feed-design-patterns?tabs=all-versions-and-deletes#intermediate-updates) (currently in preview) - in this mode function sees all intermediate changes including hard-deletes.

Customers are asking for the incremental mode to be supported by the Azure Functions Cosmos DB change feed trigger. This likely requires introducing a change feed mode option to the trigger configuration to reflect the [ChangeFeedMode](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.changefeedmode?view=azure-dotnet) from the SDK and any other changes required to process the change feed in the incremental mode to reflect and expose [the operation type](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-feed-modes?tabs=all-versions-and-deletes#parse-the-response-object-1).

MoimHossain:
+1

Marty-Me:
+1

xqrzd:
Now that this is in preview for Microsoft.Azure.WebJobs.Extensions.CosmosDB, could this also be added to isolated functions (Microsoft.Azure.Functions.Worker.Extensions.CosmosDB)?
https://github.com/Azure/azure-functions-dotnet-worker/issues/3182

PR #969, opened 2025-10-14
Add engineering for preview train packages

jviau:
Engineering work to support #961

This change allows for us to have a 'preview' train for our packages. It works by allowing us to if/def on the `#if PREVIEW` compiler constant.

- When a package opts to have a preview train, it will need to onboard by:
   - (1) add to `/eng/sln/WebJobs.Extensions.Preview.proj` (so it is part of PR CI)
        - We use a separate project because preview & stable build to the same location, so they cannot be ran in parallel.
   - (2) add its conditional `VersionSuffix` to `Directory.Version.props`.
   - (3) update its release ci yml to include a quality toggle (stable/preview)
   - (4) update its configuration mapping for `WebJobs.Extensions.sln`

How it works:

1. Preview is a build time property, intended to be passed in via `-p:Preview=true`
2. For VS development, `Debug-Preview` and `Release-Preview` configurations are added to enable toggling PREVIEW sections on and off
3. Preview trains are part of public and official CI via `/eng/sln/WebJobs.Extensions.Preview.proj`, which will ensure they are built and tested.
4. Preview releases are ran independently of stable releases.


WebJobs.Extensions.sln
```diff
@@ -40,65 +40,123 @@ EndProject
...
 		{4F335FB9-2BE8-4B51-9CE9-01465CCF05CB}.Release|Any CPU.Build.0 = Release|Any CPU
+		{4F335FB9-2BE8-4B51-9CE9-01465CCF05CB}.Release-Preview|Any CPU.ActiveCfg = Release|Any CPU
+		{4F335FB9-2BE8-4B51-9CE9-01465CCF05CB}.Release-Preview|Any CPU.Build.0 = Release|Any CPU
 		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Debug-Preview|Any CPU.ActiveCfg = Debug-Preview|Any CPU
+		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Debug-Preview|Any CPU.Build.0 = Debug-Preview|Any CPU
 		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Release|Any CPU.Build.0 = Release|Any CPU
+		{A0DE3BF4-AD56-44D9-8404-8E0778AE5CA0}.Release-Preview|Any CPU.ActiveCfg = Release-Preview|Any CPU
```
jviau, line 94: Solution files are always hard to understand, but the gist is:

1. I added 2 new configurations, `Debug-Preview` and `Release-Preview`.
2. When that config is active in Visual Studio, only specific projects (the ones with a preview) actually 'use' it. All others just map back to 'debug' or 'release'.

End result is, you select `debug-preview` in VS and only CosmosDB and CosmosDB.Tests will actually be `debug-preview`, everything else is still just `debug`.

fabiocav (commented):
Initial set of changes look good. Will follow up on questions offline.

fabiocav (approved):

PR #973, opened 2025-11-06
[CosmosDb] Avoid swallowing exceptions when retrieving votes

alrod:
**Proposed Improvement**
We want to emit error logs to the customer’s Application Insights whenever there are connectivity issues to the resource.

**Current Behavior**
Exceptions are swallowed in the extension code and logged only in internal logs.

**Planned Change**
- The change will allow connectivity exceptions to bubble up to the WebJobs SDK, so they can be logged in the standard WebJobs format.
- All extensions will follow the same logging format for consistency.
- In case of an exception, the trigger will be ignored for the current vote integration.

**References**
-https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs#L195
-https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs#L121

**Additional Notes**
- The ScaleManager class in WebJobs SDK is used for both regular scale and runtime-driven scale.
- We are also adding voting logs in the PR to standardize how the current vote is logged. These logs will also be provided to the customer.
- A bunch of exceptions that were previously not emitted in Scale Controller internally (client exceptions) have been removed. Exceptions are swallowed only in scaling code; the listener code already logs them. Need to follow with CosmosDb team if we need to filter out some transient CosmosDb exceptions.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
@@ -20,20 +20,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB.Trigger
         private readonly int _maxAssignWorkerOnNotFoundCount = 5;
         private int _assignWorkerOnNotFoundCount = 0;
 
-        private static readonly Dictionary<string, string> KnownDocumentClientErrors = new Dictionary<string, string>()
-        {
-            { "Resource Not Found", "Please check that the CosmosDB container and leases container exist and are listed correctly in Functions config files." },
-            { "The input authorization token can't serve the request", string.Empty },
-            { "The MAC signature found in the HTTP request is not the same", string.Empty },
-            { "Service is currently unavailable.", string.Empty },
-            { "Entity with the specified id does not exist in the system.", string.Empty },
-            { "Subscription owning the database account is disabled.", string.Empty },
-            { "Request rate is large", string.Empty },
-            { "The remote name could not be resolved:", string.Empty },
-            { "Owner resource does not exist", string.Empty },
-            { "The specified document collection is invalid", string.Empty }
-        };
-
         public CosmosDBMetricsProvider(ILogger logger, Container monitoredContainer, Container leaseContainer, string processorName)
         {
             _logger = logger;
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -93,45 +79,6 @@ public async Task<CosmosDBTriggerMetrics> GetMetricsAsync()
                 partitionCount = 1;
                 remainingWork = 1;
             }
-            catch (Exception e) when (e is CosmosException || e is InvalidOperationException)
-            {
-                if (!TryHandleCosmosException(e))
```
mathewc, line 98: From the links you shared Alexey, we have upstream handling already for any exceptions coming from this code on the scale monitor path. My primary concern with this was whether letting these through was going to start failing other scenarios that are currently passing (i.e. regression). I don't know who else may be using these APIs.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBMetricsProvider.cs
```diff
@@ -93,45 +79,6 @@ public async Task<CosmosDBTriggerMetrics> GetMetricsAsync()
...
-                    webException.Status == WebExceptionStatus.NameResolutionFailure)
-                {
-                    errormsg = string.Format("CosmosDBTrigger Exception message: {0}.", webException.Message);
-                }
-                else
-                {
-                    errormsg = e.ToString();
-                }
-
-                _logger.LogWarning(Events.OnScaling, errormsg);
```
mathewc, line 129: I don't understand - why won't these error messages written to the supplied ILogger go to the customer's App Insights or other log streams? The logger should already be configured to do this.

jviau (changes requested):
Changing this exception logic is a high risk of a breaking change. While this class is `internal`, the exception could now bubble up into public API and customer code depending on how it is used.

Unless we can verify 100% this will never impact a customer scenario I don't think we can take it. And I don't mean just specific functions scenario, I mean any customer using any public API in this package, functions or not.

alrod:
Closing in favor - https://github.com/Azure/azure-webjobs-sdk-extensions/pull/975

PR #976, opened 2025-11-21
[CosmosDB] Fix AllVersionsAndDelete type binding

jviau:
`AllVersionsAndDelete` mode requires customers wrap their bound POCO in `ChangeFeedItem<>` as this is how Microsoft.Azure.Cosmos supplies these values. However, when a binding does have that type it then influences the `<T>` of the listener, which means we end up binding to `ChangeFeedItem<ChangeFeedItem<T>>` (bad!) when creating the change feed listener. To address this we need to use reflection to:

1. Validate in `AllVersionsAndDelete` mode the type is `ChangeFeedItem<T>`
2. Unwrap that generic type and create a listener with just `T`. Which then results in CosmosDB supplying `ChangeFeedItem<T>`, which is what we want.

### DIFF
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
@@ -284,7 +284,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
             Type itemType = typeof(T);
             if (!IsChangeFeedItemType(itemType))
             {
-                throw new InvalidOperationException("When using ChangeFeedMode.AllVersionsAndDeletes, the trigger type must be of type System.Collections.Generic.IReadOnlyCollection<Microsoft.Azure.Cosmos.ChangeFeedItem<T>>.");
+                throw new InvalidOperationException("When using ChangeFeedMode.AllVersionsAndDeletes, the document type must be of type Microsoft.Azure.Cosmos.ChangeFeedItem<T>.");
             }
 
             itemType = itemType.GetGenericArguments()[0];
--- a/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
+++ b/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
@@ -269,29 +269,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB
 
         private ChangeFeedProcessorBuilder GetAllVersionsAndDeleteBuilder()
         {
-            static bool IsChangeFeedItemType(Type type)
-            {
-                if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(ChangeFeedItem<>))
-                {
-                    return true;
-                }
-
-                return false;
-            }
-
-            // We need to unwrap T from ChangeFeedItem<T> to T, and then construct a builder from that. We will also validate the
-            // binding is of type ChangeFeedItem<T>.
+            // We need to unwrap T from ChangeFeedItem<T> to T, and then construct a builder from that.
             Type itemType = typeof(T);
-            if (!IsChangeFeedItemType(itemType))
+            if (!itemType.IsGenericType || itemType.GetGenericTypeDefinition() != typeof(ChangeFeedItem<>))
             {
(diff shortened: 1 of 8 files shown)


src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -253,12 +253,46 @@ private ChangeFeedProcessorBuilder GetBuilder()
...
+
+                return false;
+            }
+
+            // We need to unwrap T from ChangeFeedItem<T> to T, and then construct a builder from that. We will also validate the
+            // binding is of type ChangeFeedItem<T>.
+            Type itemType = typeof(T);
+            if (!IsChangeFeedItemType(itemType))
+            {
+                throw new InvalidOperationException("When using ChangeFeedMode.AllVersionsAndDeletes, the document type must be of type Microsoft.Azure.Cosmos.ChangeFeedItem<T>.");
```
fabiocav: Include the actual type of `T` in the exception message.

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -253,12 +253,46 @@ private ChangeFeedProcessorBuilder GetBuilder()
...
+            }
+
+            itemType = itemType.GetGenericArguments()[0];
+
+            MethodInfo method = typeof(CosmosDBTriggerListener<T>).GetMethod(nameof(GetAllVersionsAndDeleteBuilderCore), BindingFlags.NonPublic | BindingFlags.Instance);
+            MethodInfo genericMethod = method.MakeGenericMethod(itemType);
+            return (ChangeFeedProcessorBuilder)genericMethod.Invoke(this, []);
```
fabiocav: Pass `null` here instead of `[]` to avoid allocation and follow documented behavior

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -253,12 +253,46 @@ private ChangeFeedProcessorBuilder GetBuilder()
...
+
+#if PREVIEW
+        private ChangeFeedProcessorBuilder GetAllVersionsAndDeleteBuilderCore<TInner>()
+        {
+            return _monitoredContainer.GetChangeFeedProcessorBuilderWithAllVersionsAndDeletes<TInner>(_processorName, ProcessChangesAsync);
+        }
+
+        private ChangeFeedProcessorBuilder GetAllVersionsAndDeleteBuilder()
+        {
+            static bool IsChangeFeedItemType(Type type)
```
fabiocav: Can we simplify this by inlining this logic? This is not abstracting a lot of functionality, and I think an inline check here would be clearer:

``` csharp
  Type originalType = typeof(T);
  if (!(originalType.IsGenericType && originalType.GetGenericTypeDefinition() == typeof(ChangeFeedItem<>)))
  {
      throw new InvalidOperationException($"When using ChangeFeedMode.AllVersionsAndDeletes, the trigger binding type must be Microsoft.Azure.Cosmos.ChangeFeedItem<T>. Actual type: '{originalType.FullName}'.");
  }
```

src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs
```diff
@@ -253,12 +253,46 @@ private ChangeFeedProcessorBuilder GetBuilder()
...
+            // binding is of type ChangeFeedItem<T>.
+            Type itemType = typeof(T);
+            if (!IsChangeFeedItemType(itemType))
+            {
+                throw new InvalidOperationException("When using ChangeFeedMode.AllVersionsAndDeletes, the document type must be of type Microsoft.Azure.Cosmos.ChangeFeedItem<T>.");
+            }
+
+            itemType = itemType.GetGenericArguments()[0];
+
+            MethodInfo method = typeof(CosmosDBTriggerListener<T>).GetMethod(nameof(GetAllVersionsAndDeleteBuilderCore), BindingFlags.NonPublic | BindingFlags.Instance);
```
fabiocav, line 281: Do we expect multiple functions with the same type binding in the app? Wondering we should cache this to avoid the reflection lookup.

jviau: I feel like to complexities of caching here will outweigh any performance benefits. This is only done at listener startup, so it isn't a perf impact through the life of the app, just startup.

fabiocav (approved):

PR #980, opened 2025-12-09
Updates to remove IApplicationLifetime dependency

fabiocav:
This pull request updates the `WebJobs.Extensions.Http` package to version 3.3.0 and introduces changes to the HTTP binding extension methods, primarily by deprecating overloads that depend on `IApplicationLifetime`. It also updates several package dependencies to newer versions for improved compatibility and maintenance.

### API Changes (Obsoleting Overloads)
* Marked `UseHttpBinding` and the overload of `UseHttpBindingRouting` that accepts `IApplicationLifetime` as obsolete, signaling developers to migrate to the new overloads that do not require `IApplicationLifetime`. [[1]](diffhunk://#diff-1e8cc4b1bafa0ae1708eeb0ab617f3f6865d8b46d0541e6007c2e1457c34fbedR22-R24) [[2]](diffhunk://#diff-1e8cc4b1bafa0ae1708eeb0ab617f3f6865d8b46d0541e6007c2e1457c34fbedR33-R47)
* Added a new overload for `UseHttpBindingRouting` that does not require `IApplicationLifetime`, simplifying the API surface.

### Dependency Updates
* Updated NuGet package dependencies in `WebJobs.Extensions.Http.csproj` to newer versions for ASP.NET Core and Azure WebJobs libraries, improving security and compatibility.

### Versioning and Documentation
* Bumped the package version to 3.3.0 in `Directory.Version.props` and updated the release notes to reflect the API deprecations. [[1]](diffhunk://#diff-4fd42fd00892a843c851818f537e98e753b3f4eadcc5b0adf03f03e1919eff88L3-R3) [[2]](diffhunk://#diff-4bded74f70a8017b27b84a10a41dd8e7afd9a1e55631c42e69130cef2a4e47d1L1-R3)

### DIFF
--- a/src/WebJobs.Extensions.Http/HttpBindingApplicationBuilderExtension.cs
+++ b/src/WebJobs.Extensions.Http/HttpBindingApplicationBuilderExtension.cs
@@ -19,8 +19,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Http
         /// <param name="applicationLifetime">The application lifetime instance.</param>
         /// <param name="routes">A route configuration handler.</param>
         /// <returns>The updated <see cref="IApplicationBuilder"/>.</returns>
+        [Obsolete("This method is obsolete. Use UseHttpBindingRouting instead.")]
         public static IApplicationBuilder UseHttpBinding(this IApplicationBuilder builder, IApplicationLifetime applicationLifetime, Action<WebJobsRouteBuilder> routes)
-            => UseHttpBindingRouting(builder, applicationLifetime, routes);
+            => UseHttpBindingRouting(builder, routes);
 
         /// <summary>
         /// Adds the WebJobs HTTP Binding routing feature to the <seealso cref="IApplicationBuilder"/> execution pipeline.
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.Http/HttpBindingApplicationBuilderExtension.cs
```diff
@@ -47,5 +58,6 @@ public static IApplicationBuilder UseHttpBindingRouting(this IApplicationBuilder
 
             return builder;
         }
+
```
brettsam: extra line here?

brettsam (approved):

src/WebJobs.Extensions.Http/HttpBindingApplicationBuilderExtension.cs
```diff
@@ -19,8 +19,9 @@ public static class HttpBindingApplicationBuilderExtension
         /// <param name="applicationLifetime">The application lifetime instance.</param>
         /// <param name="routes">A route configuration handler.</param>
         /// <returns>The updated <see cref="IApplicationBuilder"/>.</returns>
+        [Obsolete("This method is obsolete. Use UseHttpBindingRouting instead.")]
```
mathewc, line 22: Makes sense to remove one of UseHttpBinding or UseHttpBindingRouting since they were duplicates from a signature point of view (confusing). Unsure what refactorings over time lead to this.

I see Host uses UseHttpBindingRouting and also the standard AddHttp extension registration method. I don't remember now why AddHttp doesn't also register requisite binding routing, do you remember? Confusing that we have both AddHttp and UseHttpBinding/UseHttpBindingRouting.

fabiocav: > Makes sense to remove one of UseHttpBinding or UseHttpBindingRouting since they were duplicates from a signature point of view (confusing). Unsure what refactorings over time lead to this.

Yeah, it's odd we have those two methods. I'm hesitant to remove that now and introduce a binary breaking change.

src/WebJobs.Extensions.Http/HttpBindingApplicationBuilderExtension.cs
fabiocav: Yeah, caused a build break :)

mathewc (approved):

PR #991, opened 2026-04-15
Add CancellationToken overloads on ScheduleMonitor and flow them from TimerListener

liliankasem:
## Bug

In production (ICM IN_50069), an instance acquired the timer's singleton listener lock at 01:06:40 but never started the timer listener. It held the lock for ~1.5 hours, blocking every other instance from running the timer, until the instance was recycled at 02:36:30.

Root cause: `TimerListener.StartAsync` calls `ScheduleMonitor.GetSafeStatusAsync`, which (in `StorageScheduleMonitor`) calls `BlobClient.DownloadAsync` without a `CancellationToken`. When AAD token acquisition (`BearerTokenAuthenticationPolicy.GetHeaderValueAsync`) hung, there was nothing to cancel it. `SingletonListener` (in WebJobs SDK) holds its lock around the inner listener's `StartAsync`, so the hang held the singleton lock until process exit.

## Scope of this PR

This is the first of two PRs.

**This PR (Microsoft.Azure.WebJobs.Extensions):** add the API surface and flow the token from `TimerListener`.

- `ScheduleMonitor`: add `CancellationToken`-aware virtual overloads of `GetStatusAsync`, `GetSafeStatusAsync`, and `UpdateStatusAsync`. Originals retained for binary compatibility; new virtuals delegate to the no-token versions by default so existing subclasses keep working unchanged.
- `TimerListener.StartInternalAsync`: pass `cancellationToken` to the pre-invocation `GetSafeStatusAsync` call (the observed hang site).
- `TimerListener` post-invocation `UpdateStatusAsync` is intentionally **not** cancelled. Once a function has been invoked, status persistence should complete, otherwise a host shutdown mid-write could cause duplicate invocations on the next start.
- `CheckPastDueAsync` is intentionally left untouched. It is a documented `virtual` extension point with the past-due logic in its body, so any third-party subclass override would have been silently bypassed if `TimerListener` switched to a CT overload. The observed hang was not in this method, so we keep behavior here unchanged.

The change is purely additive: no existing public method signatures change, and no behavioral change for any existing subclass (in-repo or third-party).

**Follow-up PR (Microsoft.Azure.WebJobs.Extensions.Timers.Storage):** override the new `ScheduleMonitor` overloads to pass the token down to the `BlobClient` calls. That PR will be opened once `Microsoft.Azure.WebJobs.Extensions` 5.3.0 is published to NuGet (the Storage package consumes Extensions as a NuGet `PackageReference`).

## Versioning

`Microsoft.Azure.WebJobs.Extensions`: 5.2.1 → 5.3.0 (new public API surface on `ScheduleMonitor`).

### DIFF
--- a/src/WebJobs.Extensions.Timers.Storage/StorageScheduleMonitor.cs
+++ b/src/WebJobs.Extensions.Timers.Storage/StorageScheduleMonitor.cs
@@ -11,7 +11,6 @@ using Azure.Storage.Blobs;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Logging;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 
@@ -27,7 +26,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
         private readonly JsonSerializer _serializer;
         private readonly ILogger _logger;
         private readonly IHostIdProvider _hostIdProvider;
-        private readonly CancellationToken _shutdownToken;
         private string _timerStatusPath;
         private BlobContainerClient _containerClient;
 
@@ -37,13 +35,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
         /// <param name="hostIdProvider"><see cref="IHostIdProvider"/> to retrieve the current Host ID.</param>
         /// <param name="loggerFactory"><see cref="ILoggerFactory"/> for logging purposes.</param>
         /// <param name="azureStorageProvider"><see cref="IAzureBlobStorageProvider"/> to instantiate Blob-related clients for Timer operations.</param>
-        /// <param name="applicationLifetime"><see cref="IApplicationLifetime"/> to observe host shutdown for cancellation of storage operations.</param>
-        public StorageScheduleMonitor(IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory, IAzureBlobStorageProvider azureStorageProvider, IApplicationLifetime applicationLifetime)
+        public StorageScheduleMonitor(IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory, IAzureBlobStorageProvider azureStorageProvider)
         {
             _hostIdProvider = hostIdProvider ?? throw new ArgumentNullException(nameof(hostIdProvider));
             _logger = loggerFactory.CreateLogger(LogCategories.CreateTriggerCategory("Timer"));
             _azureStorageProvider = azureStorageProvider ?? throw new ArgumentNullException(nameof(azureStorageProvider));
-            _shutdownToken = applicationLifetime?.ApplicationStopping ?? CancellationToken.None;
 
             JsonSerializerSettings settings = new JsonSerializerSettings
             {
@@ -63,7 +59,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Timers
                 // be available AFTER the host as been started
                 if (string.IsNullOrEmpty(_timerStatusPath))
                 {
-                    string hostId = _hostIdProvider.GetHostIdAsync(_shutdownToken).GetAwaiter().GetResult();
+                    string hostId = _hostIdProvider.GetHostIdAsync(CancellationToken.None).GetAwaiter().GetResult();
                     if (string.IsNullOrEmpty(hostId))
                     {
                         throw new InvalidOperationException("Unable to determine host ID.");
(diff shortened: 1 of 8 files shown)


src/WebJobs.Extensions.Timers.Storage/StorageScheduleMonitor.cs
```diff
@@ -35,11 +37,13 @@ public class StorageScheduleMonitor : ScheduleMonitor
         /// <param name="hostIdProvider"><see cref="IHostIdProvider"/> to retrieve the current Host ID.</param>
         /// <param name="loggerFactory"><see cref="ILoggerFactory"/> for logging purposes.</param>
         /// <param name="azureStorageProvider"><see cref="IAzureBlobStorageProvider"/> to instantiate Blob-related clients for Timer operations.</param>
-        public StorageScheduleMonitor(IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory, IAzureBlobStorageProvider azureStorageProvider)
+        /// <param name="applicationLifetime"><see cref="IApplicationLifetime"/> to observe host shutdown for cancellation of storage operations.</param>
+        public StorageScheduleMonitor(IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory, IAzureBlobStorageProvider azureStorageProvider, IApplicationLifetime applicationLifetime)
```
jviau: This is a public type, adding to the ctor is a breaking change. We will need to create a separate ctor and ensure it is selected in DI.

liliankasem: Good catch. Restored the original 3-arg public constructor to preserve binary compatibility, and added [ActivatorUtilitiesConstructor] to the new 4-arg constructor so DI selects it. The 3-arg ctor now delegates to the 4-arg ctor with a null IApplicationLifetime (falling back to CancellationToken.None).

mathewc:
This monitor isn't performing any singleton lock operations. When you say "the instance would hold the singleton lock" what is the actual call path that is producing a hang? I'm assuming you're saying its upstream in TimerListener. E.g. listener stop being held because there's a hung outstanding invocation.

If so, I'd expect you to also be passing a cancellation token from TimerListener usages of the monitor, e.g. here https://github.com/Azure/azure-webjobs-sdk-extensions/blob/ebb371c86d22acf34a1fe8056804fb8b8a42f2e2/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs#L404.

However, if we've already invoked a function and are attempting to update the tracking blob, it'd be a mistake to cancel that operation - we really want to let that succeed if possible, otherwise the customer can get repeat invocations.

Can you share details on the bug you're fixing with this? I'm thinking what you want to do is flow tokens from TimerListener not use application lifetime. And perhaps you only flow tokens on the calls that are before the invocation took place, and on the update status blob call you use None, or perhaps a 5 second timeout token to try to allow that critical operation to complete.

liliankasem:
> This monitor isn't performing any singleton lock operations. When you say "the instance would hold the singleton lock" what is the actual call path that is producing a hang? I'm assuming you're saying its upstream in TimerListener. E.g. listener stop being held because there's a hung outstanding invocation.
> 
> If so, I'd expect you to also be passing a cancellation token from TimerListener usages of the monitor, e.g. here
> 
> https://github.com/Azure/azure-webjobs-sdk-extensions/blob/ebb371c86d22acf34a1fe8056804fb8b8a42f2e2/src/WebJobs.Extensions/Extensions/Timers/Listener/TimerListener.cs#L404

You're right, it's upstream in TimerListener. The lock in question is SingletonListener's lock, held around the inner listener's StartAsync. The hang is:

`SingletonListener.StartAsync (acquires lock) → TimerListener.StartAsync → ScheduleMonitor.GetSafeStatusAsync → GetStatusAsync → BlobClient.DownloadAsync → BearerTokenAuthenticationPolicy.GetHeaderValueAsync hangs with no cancellation.`

> Can you share details on the bug you're fixing with this?

This was from a CRI where an instance acquired the singleton lock and never started the timer listener. The lock wasn't released until the instance recycled ~1.5 hours later. During that window no other instance could run the timer. Other instances on the same app failed cleanly with `TaskCanceledException` from `BlobLeaseDistributedLockManager.ReadLeaseBlobMetadata` (their lock acquisition flowed a CT), but our inner StartAsync path doesn't, so it just hangs.

> I'm thinking what you want to do is flow tokens from TimerListener not use application lifetime. And perhaps you only flow tokens on the calls that are before the invocation took place, and on the update status blob call you use None, or perhaps a 5 second timeout token to try to allow that critical operation to complete.

This is a great suggestion - I'll redo the PR!

liliankasem:
@mathewc @jviau thoughts on latest iteration of this? Follow up PR for timer storage here:

https://github.com/Azure/azure-webjobs-sdk-extensions/pull/998

jviau (approved):

